PR #23, opened 2017-09-24
Closes #18: Recover account using mnemonic phrase.

### LINKED ISSUE #18: Recover account using mnemonic phrase

fmrsabino:


app-core/src/main/java/pm/gnosis/heimdall/util/ContextUtils.kt
```diff
@@ -35,3 +37,13 @@ fun Context.shareExternalText(text: String, dialogTitle: String = "") {
 
 fun Activity.scanQrCode() = ZxingIntentIntegrator(this).initiateScan(ZxingIntentIntegrator.QR_CODE_TYPES)
 fun Fragment.scanQrCode() = ZxingIntentIntegrator(this).initiateScan(ZxingIntentIntegrator.QR_CODE_TYPES)
+
+fun <T : AppCompatActivity> Activity.startActivity(activity: KClass<T>, noHistory: Boolean = false) {
+    val i = Intent(this, activity.java)
```
rmeissner: I would propose to pass in an Intent, else it is not possible to add params with this method. Also I like the pattern that the activity that is started has an method to create the intent. This makes sure that you don't forget expected parameters

app/src/main/java/pm/gnosis/heimdall/ui/base/BaseFragment.kt
```diff
@@ -3,15 +3,15 @@ package pm.gnosis.heimdall.ui.base
...
+import pm.gnosis.heimdall.HeimdallApplication
 import pm.gnosis.heimdall.di.component.ApplicationComponent
 
 abstract class BaseFragment : Fragment() {
     protected val disposables = CompositeDisposable()
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        inject(GnosisAuthenticatorApplication[this.context].component)
+        inject(HeimdallApplication[this.context].component)
```
rmeissner: Can we write an extension to get the app component from the context. I hate this linr xD

PR #24, opened 2017-09-26
Closes #22. Extract string resources.

### LINKED ISSUE #22: Extract string resources

fmrsabino:


app/src/main/res/values/static_strings.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="eth" translatable="false">ETH</string>
```
rmeissner: maybe something like `currency_eth` would be a little more descriptive

PR #49, opened 2017-10-06
Closes #25: Restore account is now a ViewModel. Add tests.

### LINKED ISSUE #25: Adjust Onboarding Screens
- [x] Change design of generate mnemonic 
- [x] Change design of restore mnemonic
- [x] Add longpress to copy mnemonic
- [x] Remove initial screen to select generate or restore and add a `restore` option to the generate screen
- [x] Adjust architecture for both screens
- [x] Add tests for both screens

fmrsabino:


accounts/geth/src/main/java/pm/gnosis/heimdall/accounts/repositories/impl/GethAccountsRepository.kt
```diff
@@ -73,4 +74,8 @@ class GethAccountsRepository @Inject constructor(
             Completable.fromCallable {
                 gethKeyStore.importECDSAKey(privateKey, generateRandomString())
             }
+
+    override fun generateMnemonic(): Single<String> = Single.just(bip39.generateMnemonic())
```
rmeissner: This is happening on the UI thread, correct? Is that on purpose?

accounts/geth/src/main/java/pm/gnosis/heimdall/accounts/repositories/impl/GethAccountsRepository.kt
```diff
@@ -73,4 +74,8 @@ class GethAccountsRepository @Inject constructor(
...
+
+    override fun validateMnemonic(mnemonic: String): Single<Boolean> = Single.just(bip39.validateMnemonic(mnemonic))
```
rmeissner: same here

accounts/kethereum/src/main/java/pm/gnosis/heimdall/accounts/repositories/impl/KethereumAccountsRepository.kt
```diff
@@ -63,4 +64,8 @@ class KethereumAccountsRepository @Inject internal constructor(private val accou
             putString(PreferencesManager.MNEMONIC_KEY, mnemonic)
         }
     }
+
+    override fun generateMnemonic(): Single<String> = Single.just(bip39.generateMnemonic())
```
rmeissner: see above

app/src/main/java/pm/gnosis/heimdall/common/di/component/ApplicationComponent.kt
```diff
@@ -50,6 +53,7 @@ interface ApplicationComponent {
     fun gnosisAuthenticatorDb(): GnosisAuthenticatorDb
 
     fun gnosisMultiSig(): GnosisMultisigWrapper
+    fun bip39(): Bip39
```
rmeissner: Do we need to expose this, if we only use it within the repositories?

app/src/main/java/pm/gnosis/heimdall/ui/onboarding/RestoreAccountViewModel.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import io.reactivex.Completable
+import io.reactivex.Single
+import io.reactivex.schedulers.Schedulers
+import pm.gnosis.heimdall.accounts.base.repositories.AccountsRepository
+import javax.inject.Inject
+
+class RestoreAccountViewModel @Inject constructor(private val accountsRepository: AccountsRepository) : RestoreAccountContract() {
+    override fun isValidMnemonic(mnemonic: String): Single<Boolean> {
+        val words = mnemonic.split(Regex("\\s+"))
```
rmeissner: Shouldn't that be part of the validateMnemonic function? (I got a bad comment on one of my challenges because this is business logic in the viewmodel class :D )

app/src/main/java/pm/gnosis/heimdall/ui/onboarding/RestoreAccountViewModel.kt
```diff
@@ -0,0 +1,21 @@
...
+        if (words.size < 12) return Single.just(false)
```
rmeissner: Also we could return throw an exception if invalid to pass along some information why it is invalid

mnemonic/src/main/java/pm/gnosis/mnemonic/Bip39Generator.kt
```diff
@@ -0,0 +1,101 @@
...
+import java.security.spec.KeySpec
+import java.text.Normalizer
+import javax.crypto.SecretKey
+import javax.crypto.spec.PBEKeySpec
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class Bip39Generator @Inject constructor() : Bip39 {
+    class Hasher : PBEPBKDF2.PBKDF2withSHA512() {
```
rmeissner: this class can be private (just a general reminder for us :D)

mnemonic/src/main/java/pm/gnosis/mnemonic/Bip39Generator.kt
```diff
@@ -0,0 +1,101 @@
...
+                throw IllegalArgumentException("Invalid mnemonic")
+
+        val binaryIndexes = wordList.words.getIndexesAllMatching(words).joinToString("") { Integer.toBinaryString(it).padStart(11, '0') }
+
+        val checksum = binaryIndexes.subSequence(entropyNBits, binaryIndexes.length)
+        val originalEntropy = binaryIndexes.subSequence(0, binaryIndexes.length - checksumNBits)
+        val originalBytes = (0 until originalEntropy.length step 8).map { (Integer.valueOf((originalEntropy.subSequence(it, it + 8).toString()), 2) and 0xFF).toByte() }.toByteArray()
+
+
+        val digest = MessageDigest.getInstance("SHA-256")
```
rmeissner: val digest = SHA256.Digest()

mnemonic/src/main/java/pm/gnosis/mnemonic/di/Bip39BindingModule.kt
```diff
@@ -0,0 +1,12 @@
+package pm.gnosis.mnemonic.di
+
+import dagger.Binds
+import dagger.Module
+import pm.gnosis.mnemonic.Bip39
+import pm.gnosis.mnemonic.Bip39Generator
+
+@Module
+abstract class Bip39BindingModule {
+    @Binds
```
rmeissner: `@Singleton`?

accounts/geth/src/main/java/pm/gnosis/heimdall/accounts/repositories/impl/GethAccountsRepository.kt
fmrsabino: The `ViewModel` specifies on which thread it will run. I think it makes more sense in this way because we can have situations where we already are on a different thread and we want this computation to be executed there

app/src/main/java/pm/gnosis/heimdall/ui/onboarding/RestoreAccountViewModel.kt
fmrsabino: Since a validation function would return true if it was valid and an exception if it was not valid (instead of false) I decided to go with a `sealed class` implementation.

app/src/main/java/pm/gnosis/heimdall/ui/onboarding/RestoreAccountActivity.kt
```diff
@@ -50,12 +52,28 @@ class RestoreAccountActivity : BaseActivity() {
...
+                disposables += saveAccountWithMnemonicDisposable(layout_restore_account_mnemonic.text.toString())
+            }
+            is InvalidEntropy -> {
+                layout_restore_account_mnemonic_input_layout.error = getString(R.string.invalid_mnemonic)
+            }
+            is InvalidChecksum -> {
+                layout_restore_account_mnemonic_input_layout.error = getString(R.string.invalid_mnemonic)
+            }
+            is MnemonicNotInWordlist -> {
+                val wordLists = BIP39_WORDLISTS.keys.joinToString()
```
rmeissner: Having this in the view feels kind of weird, as it is a pretty specifix implementation detail

PR #82, opened 2017-11-30
Update Gradle and Kotlin

fmrsabino:
@gnosis/mobile-devs


fmrsabino:
Will do that and try to adjust some stuff in the code for the new kotlin version.

PR #90, opened 2017-12-13
Fixed authenticate screen

rmeissner:
Fixed authenticate screen

- Use correct layout for error snackbar
- Enable scanning of any transaction
- Fixed handling of returned results from scanner

@gnosis/mobile-devs


rmeissner:
I was thinking about that, but I wanted to change as little as possible (didn't want to adjust the tests). We will need to refactor that screen soonish anyways

PR #102, opened 2017-12-29
WA-89: Show transaction status in list

rmeissner:
Closes WA-89.

- Show transaction status in safe transaction list

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/GnosisSafeTransactionRepository.kt
+++ b/app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/GnosisSafeTransactionRepository.kt
@@ -9,10 +9,12 @@ import pm.gnosis.heimdall.GnosisSafe
 import pm.gnosis.heimdall.accounts.base.repositories.AccountsRepository
 import pm.gnosis.heimdall.data.db.GnosisAuthenticatorDb
 import pm.gnosis.heimdall.data.db.models.TransactionDescriptionDb
+import pm.gnosis.heimdall.data.db.models.TransactionPublishStatusDb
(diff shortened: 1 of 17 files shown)


app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/GnosisSafeTransactionRepository.kt
```diff
@@ -173,17 +176,43 @@ class GnosisSafeTransactionRepository @Inject constructor(
...
+                    descriptionsDao.insert(
                             TransactionDescriptionDb(transactionUuid, safeAddress, transaction.address,
                                     transaction.value?.value ?: BigInteger.ZERO,
                                     transaction.data ?: "", DEFAULT_OPERATION,
                                     transaction.nonce ?: DEFAULT_NONCE, System.currentTimeMillis(), null, it.toHexString())
-
+                    )
+                    descriptionsDao.insert(
+                            TransactionPublishStatusDb(transactionUuid, txChainHash, null)
                     )
```
fmrsabino: Maybe we can do both `inserts` in the context of 1 Transaction by using [`@Transaction`](https://developer.android.com/reference/android/arch/persistence/room/Transaction.html)

rmeissner: we would have to define that on the Dao, right? We could do that, but it feels a little weird. Let's check for stuff like that later :)

fmrsabino: Yeah it needs to be in the Dao 😉

PR #118, opened 2018-01-19
WA-107: Added tests for GnoSafeUrlParser

rmeissner:
Closes WA-107.

- Added tests for GnoSafeUrlParser
- Fixed bug in GnoSafeUrlParser with null values

@gnosis/mobile-devs

### DIFF
--- a/models/src/main/java/pm/gnosis/models/Wei.kt
+++ b/models/src/main/java/pm/gnosis/models/Wei.kt
@@ -4,11 +4,13 @@ import java.math.BigDecimal
 import java.math.BigInteger
 
 data class Wei(val value: BigInteger) {
-    fun toEther(scale: Int = 18): BigDecimal = BigDecimal(value).setScale(scale).div(BigDecimal(10).pow(18))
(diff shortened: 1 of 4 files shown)


models/src/main/java/pm/gnosis/models/Wei.kt
```diff
@@ -4,11 +4,13 @@ import java.math.BigDecimal
...
 data class Wei(val value: BigInteger) {
-    fun toEther(scale: Int = 18): BigDecimal = BigDecimal(value).setScale(scale).div(BigDecimal(10).pow(18))
+    fun toEther(scale: Int = 18): BigDecimal = BigDecimal(value).setScale(scale).div(WEI_TO_ETHER_MULTIPLIER)
 
     fun toLong() = value.toLong()
 
     companion object {
+        private val WEI_TO_ETHER_MULTIPLIER = BigDecimal(10).pow(18)
         val ZERO = Wei(BigInteger.ZERO)
+        fun ether(value: Double) = Wei((BigDecimal.valueOf(value) * WEI_TO_ETHER_MULTIPLIER).toBigInteger())
```
fmrsabino: Not so sure about having `Double` here (ie.: in order to avoid loss of precision). Maybe receiving a `String` which represents the value or the `BigDecimal` itself 😉

PR #119, opened 2018-01-20
WA-109: Signature Push

rmeissner:
Closes WA-109.

- Adds possibility to request/share signatures via push

@gnosis/mobile-devs


app/src/main/java/pm/gnosis/heimdall/services/HeimdallFirebaseService.kt
```diff
@@ -0,0 +1,101 @@
...
+import pm.gnosis.heimdall.HeimdallApplication
+import pm.gnosis.heimdall.R
+import pm.gnosis.heimdall.accounts.base.repositories.AccountsRepository
+import pm.gnosis.heimdall.data.repositories.SignaturePushRepository
+import pm.gnosis.heimdall.ui.transactions.SignTransactionActivity
+import pm.gnosis.heimdall.utils.GnoSafeUrlParser
+import pm.gnosis.utils.asEthereumAddressString
+import pm.gnosis.utils.removeHexPrefix
+import timber.log.Timber
+import javax.inject.Inject
```
fmrsabino: Unecessary imports: `android.util.Log` and `com.squareup.moshi.Moshi`

app/src/main/java/pm/gnosis/heimdall/services/HeimdallFirebaseService.kt
```diff
@@ -0,0 +1,101 @@
...
+
+        disposables += accountsRepo.loadActiveAccount()
+                .subscribe({
+                    if (targets.contains(it.address.asEthereumAddressString().removeHexPrefix().toLowerCase())) {
+                        showNotification(parsed)
+                    }
+                }, Timber::e)
+    }
+
+    private fun showNotification(signRequest: GnoSafeUrlParser.Parsed.SignRequest) {
```
fmrsabino: Not something that I'd do now but maybe in the future we can have a module just for notifications (like `HeimdallNotificationManager`) if we want to show other types of notification. (basically just an abstraction with this builder).

app/src/main/java/pm/gnosis/heimdall/ui/dialogs/share/RequestSignatureDialog.kt
```diff
@@ -76,6 +86,19 @@ class RequestSignatureDialog : BaseShareQrCodeDialog() {
...
+                    signaturePushRepository.request(safe, data())
+                            .mapToResult()
+                }
+                .observeOn(AndroidSchedulers.mainThread())
+                .subscribeForResult({
+                    context!!.toast(R.string.signature_request_sent)
+                    dismiss()
+                }, {
+                    Timber.e(it)
+                    errorSnackbar(dialog_request_signature_push, it)
```
fmrsabino: isn't the snackbar dimmed in the dialog? If it is maybe we can use `Context.errorToast`

app/src/main/java/pm/gnosis/heimdall/ui/settings/SettingsActivity.kt
```diff
@@ -39,6 +39,7 @@ class SettingsActivity : BaseActivity() {
             startActivity(TokenManagementActivity.createIntent(this))
         }
 
+        // TODO: this check should be done in the SecuritySettingsActivity
```
fmrsabino: Actually this is not correct in the current state 😬 

We added this when we only had the fingerprint setting. But we added the "Reveal Security Phrase" and "Change Password" later on... So these options are only accessible on devices with fingerprint.

build.gradle
```diff
@@ -15,6 +15,7 @@ buildscript {
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$versions.kotlin"
         classpath "com.github.gnosis.bivrost-kotlin:bivrost-gradle-plugin:$versions.bivrost"
         classpath "io.fabric.tools:gradle:$versions.fabric"
+        classpath 'com.google.gms:google-services:3.1.0'
```
fmrsabino: We have this version in a constant: `$versions.google_services`

app/src/main/java/pm/gnosis/heimdall/ui/settings/SettingsActivity.kt
rmeissner: I opened a ticket https://gnosisltd.atlassian.net/browse/WA-111

app/src/main/java/pm/gnosis/heimdall/ui/dialogs/share/RequestSignatureDialog.kt
rmeissner: Let's wait for designs with that

app/src/main/java/pm/gnosis/heimdall/services/HeimdallFirebaseService.kt
rmeissner: agreed, we also should support notification channels

PR #134, opened 2018-02-07
WA-143: Adjust design of Splash and Unlock Activities

fmrsabino:
- Adjust design of Splash and Unlock Activities

@gnosis/mobile-devs


app/src/main/res/layout/layout_unlock.xml
```diff
@@ -18,21 +18,21 @@
...
-                android:layout_marginTop="56dp"
+                android:layout_marginTop="32dp"
                 android:gravity="center_horizontal"
                 android:orientation="vertical">
 
                 <ImageView
                     android:id="@+id/setup_fingerprint"
-                    android:layout_width="wrap_content"
-                    android:layout_height="wrap_content"
+                    android:layout_width="75dp"
```
rmeissner: Why not wrap?

rmeissner: We could use different assets. Would be a minor performance improvement. I am fine either way

PR #151, opened 2018-03-13
WA-169, WA-173: Preparation for MVP 3

rmeissner:
Closes WA-169 and WA-173.

- Integrate transaction execution service to execute transactions (and safe creation) on Rinkeby
- Add 3rd party wallet integration as an alternative to the transaction execution service

Transactions will no longer be executed by the account of the Gnosis Safe app. Therefore it is not necessary that the app account holds any ether. The user can either use the transaction execution service by paying via in-app purchases or trigger the execution via an external wallet.

@gnosis/mobile-devs


app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/PlayBillingRepository.kt
```diff
@@ -0,0 +1,164 @@
...
+
+    private val billingClient = BillingClient.newBuilder(context).setListener(this).build()
+
+    private val pendingActionsLock = Any()
+    private val pendingActions = ArrayList<BillingAction>()
+
+    private var isConnected = false
+
+    override fun init() {
+        updatePurchases()
```
fmrsabino: I would keep track of this disposable in `HeimdallApplication` since it is tied to that lifecycle.

app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/PlayBillingRepository.kt
```diff
@@ -0,0 +1,164 @@
...
+                true
+            } ?: run {
+                // Default to error if no activity found
+                false
+            }
+        }
+    }
+
+    override fun onPurchasesUpdated(billingResponse: Int, purchases: MutableList<Purchase>?) {
+        updatePurchases()
```
fmrsabino: Maybe emitting an event to the stream might be better than creating a new disposable.

PR #152, opened 2018-03-13
WA-98: Add in-app barcode scanner

fmrsabino:
* Replaced external barcode dependency (using Zxing) with an in-app camera view so the user doesn't need to have external barcode scanner apps installed.

@gnosis/mobile-devs


app/src/main/java/pm/gnosis/heimdall/ui/qrscan/Videographer.kt
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    /**
+     * Sets the camera display orientation to be the same as the display's.
+     *
+     * @see android.hardware.Camera.setDisplayOrientation(int)
+     */
+    private fun setCameraDisplayOrientation() {
+        val info = Camera.CameraInfo()
+        Camera.getCameraInfo(defaultCameraId, info)
```
rmeissner: that looks wrong

app/src/main/java/pm/gnosis/heimdall/ui/qrscan/Videographer.kt
```diff
@@ -0,0 +1,182 @@
...
+import android.os.HandlerThread
+import android.view.Surface
+import android.view.TextureView
+import timber.log.Timber
+
+/*
+ * Check https://github.com/walleth/walleth/tree/master/app/src/main/java/org/walleth/activities/qrscan
+ */
+@Suppress("DEPRECATION")
+class Videographer(val activity: Activity) {
```
rmeissner: We could use the context here and acquire the window manager as a system service:
https://developer.android.com/reference/android/view/WindowManager.html

This would allow us to inject this class

app/src/main/java/pm/gnosis/heimdall/ui/qrscan/Videographer.kt
```diff
@@ -0,0 +1,182 @@
...
+    private val scanner = Scanner(this)
```
rmeissner: What about injecting this (and maybe even the camera) that would allow us to do some easier testing

rmeissner: Also using our own interface for the camera would allow us to add support for Camera2 api later on. For this the necessary methods should expose observables (not sure if you want to do this in this PR)

fmrsabino: The problem with this injection is that there's a circular dependency also on the scanner side. `Videographer` depends on the `Scanner` and the `Scanner` depends on the `Videographer`. I'll look into it in more detail.

fmrsabino:
Restructured the dependencies. At some point we can introduce the interface + `cameraV2`.

app/src/main/java/pm/gnosis/heimdall/ui/qrscan/Videographer.kt
```diff
@@ -43,7 +48,7 @@ class Videographer(val activity: Activity) {
         }
     }
 
-    lateinit var onSuccessfulScan: (result: String) -> Unit
+    var onSuccessfulScan: (result: String) -> Unit = {}
```
rmeissner: I am always more a fan of nullable callbacks, to avoid unnecessary creation of empty function classes

app/src/main/java/pm/gnosis/heimdall/ui/qrscan/QRCodeScanActivity.kt
```diff
@@ -50,12 +57,20 @@ class QRCodeScanActivity : BaseActivity() {
         handlePermissionsResultRequest(requestCode, CAMERA_REQUEST_CODE, grantResults, onPermissionDenied = { finish() })
     }
 
+    // We can use Rx to listen for successful events and finish instead of passing a callback
```
rmeissner: Is this a TODO?

app/src/main/java/pm/gnosis/heimdall/ui/qrscan/Videographer.kt
```diff
@@ -9,17 +9,22 @@ import android.os.Handler
...
 import timber.log.Timber
+import javax.inject.Inject
 
 /*
  * Check https://github.com/walleth/walleth/tree/master/app/src/main/java/org/walleth/activities/qrscan
  */
 @Suppress("DEPRECATION")
-class Videographer(val activity: Activity) {
-    private val scanner = Scanner(this)
+@ViewContext
```
rmeissner: `@ForView`?

PR #159, opened 2018-04-03
WA-254: Focus on single safe

rmeissner:
Closes WA-254 .

- When the app is started one safe is automatically selected. The selected safe can be changed in the drawer menu.

@gnosis/mobile-devs


app/src/main/java/pm/gnosis/heimdall/ui/safe/main/SafeMainViewModel.kt
```diff
@@ -0,0 +1,63 @@
...
+        }.onErrorResumeNext {
+            safeRepository.observeSafes()
+                .map {
+                    it.first()
+                }
+                .firstOrError()
+        }
+
+    private fun loadSafe(addressOrHash: BigInteger): Single<AbstractSafe> =
+    // Fuck you kotlin why do I have to add this map for type casting?
```
fmrsabino: This is because of invariance https://kotlinlang.org/docs/reference/generics.html#variance

Changing `GnosisSafeRepository.loadSafe(address: BigInteger): Single<Safe>` to `GnosisSafeRepository.loadSafe(address: BigInteger): Single<AbstractSafe>` might work

PR #165, opened 2018-04-12
WA-217: Use Solidity types from Bivrost

fmrsabino:
Closes WA-217.

- Use Solidity types from Bivrost

@gnosis/mobile-devs


fmrsabino:
To merge only after Svalinn release

PR #172, opened 2018-05-29
WA-353: Improve unlock and splash screens

fmrsabino:
- Improve Splash Screen
- Improve Unlock Screen

@gnosis/mobile-devs


tschubotz:
Works for me 👍🏼
1 fun "bug" though: 
- Setup app
- Kill app
- Open app again
- Enter wrong password multiple times (>10 times)
- Enter correct password
- Kill app
-> The toast does not disappear for quite some time
- Reopen app
-> Already unlocked as long as the toast is visible 🙈

Richard told me, it's an Android bug. So I would leave that as is for now and change it at a later point.

rmeissner:
@tschubotz found an interesting behaviour: If you span the "done" button on the keyboard with an invalid password we spawn a lot of toast that are put into queue and display for a very long time. While a toast is displayed the app cannot be completely killed (at least on the Moto G), therefore you cannot lock the app by removing it from the recents. 

For now I would ignore this (we should just check that the time based locking still works). In the future we maybe should try not to use the toast.

PR #176, opened 2018-06-07
Upgrade dependencies

rmeissner:
- Upgrade firebase dependencies
- Upgrade moshi dependencies
- Enable moshi adapter generation
- Update proguard config

@gnosis/mobile-devs


app/src/main/java/pm/gnosis/heimdall/data/remote/models/push/PushMessages.kt
```diff
@@ -3,23 +3,23 @@ package pm.gnosis.heimdall.data.remote.models.push
 import com.squareup.moshi.Json
 
 sealed class PushMessage(
-    @Json(name = "type") val type: String
```
fmrsabino: why is the `@Json` annotation removed?

rmeissner: `PushMessage` classes are only parsed from maps, therefore I removed all stuff related to moshi

PR #191, opened 2018-06-18
WA-378: Change safe name; WA-381: remove safe

rmeissner:
Closes WA-378 and WA-381.

- Change safe name
- Remove safe

@gnosis/mobile-devs


app/src/main/java/pm/gnosis/heimdall/ui/safe/main/SafeMainViewModel.kt
```diff
@@ -69,8 +76,25 @@ class SafeMainViewModel @Inject constructor(
             loadSelectedSafe()
         }
 
+    override fun observeSafe(safe: AbstractSafe): Flowable<Pair<String, String>> =
+        when(safe) {
+            is Safe -> safeRepository.observeSafe(safe.address)
+                .map { it.displayName(context) to it.address.shortChecksumString() }
+            is PendingSafe -> safeRepository.observePendingSafe(safe.hash)
+                .map { it.displayName(context) to it.address.shortChecksumString() }
+        }
```
fmrsabino: What do you think about moving the `address` to the `AbstractSafe` since it is common to both `Safe` and `PendingSafe`. 

Then we could also have:
```kotlin
when(safe) {
            is Safe -> safeRepository.observeSafe(safe.address)
            is PendingSafe -> safeRepository.observePendingSafe(safe.hash)
        }.map { it.displayName(context) to it.address.shortChecksumString() }
```

but really not important 😉 was more a question regarding the model itself

rmeissner: I wanted to do that with address and name, but it is ugly with data classes.q

PR #200, opened 2018-07-05
WA-430: Fix safe name reset after deploy

rmeissner:
Closes WA-430.

- Fix safe name reset after deploy
- Fix positioning of shadow on safe creation intro

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/DefaultShortcutRepository.kt
+++ b/app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/DefaultShortcutRepository.kt
@@ -1,0 +1,86 @@
+package pm.gnosis.heimdall.data.repositories.impls
+
+import android.annotation.TargetApi
+import android.content.Context
+import android.content.Intent
+import android.content.pm.ShortcutInfo
+import android.content.pm.ShortcutManager
+import android.graphics.Bitmap
+import android.graphics.Canvas
+import android.graphics.drawable.Icon
+import android.os.Build
+import io.reactivex.rxkotlin.subscribeBy
+import io.reactivex.schedulers.Schedulers
+import pm.gnosis.blockies.Blockies
+import pm.gnosis.blockies.BlockiesPainter
+import pm.gnosis.heimdall.R
+import pm.gnosis.heimdall.data.repositories.GnosisSafeRepository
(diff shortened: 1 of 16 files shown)


rmeissner:
@fmrsabino Please also review https://github.com/gnosis/svalinn-kotlin/pull/25, due to some jitpack issues I had to merge it before, but I would like to get this reviewed since we should not have unreviewed commits ;)

app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/DefaultShortcutRepository.kt
```diff
@@ -0,0 +1,86 @@
...
+
+    private fun blockiesBitmap(address: Solidity.Address) =
+        Bitmap.createBitmap(iconDimension.toInt(), iconDimension.toInt(), Bitmap.Config.ARGB_8888).apply {
+            blockiesPainter.draw(Canvas(this), Blockies.fromAddress(address)!!)
+        }
+
+    private fun Intent.prepare(): Intent = apply {
+        action = Intent.ACTION_VIEW
+        flags = Intent.FLAG_ACTIVITY_CLEAR_TASK or Intent.FLAG_ACTIVITY_NEW_TASK
+    }
```
fmrsabino: Not super important but I was checking the behavior in some of the apps and the back goes to the "parent" activity (when launched from a shortcut). But this requires changing the behavior of the "up" navigation that we have across the screens.

app/src/main/java/pm/gnosis/heimdall/ui/safe/create/ConfirmSafeRecoveryPhraseViewModel.kt
```diff
@@ -45,12 +45,12 @@ class ConfirmSafeRecoveryPhraseViewModel @Inject constructor(
...
         }
-            .andThen(loadShuffledWords())
+            .andThen(loadWordsToDisplay())
             .subscribeOn(Schedulers.io())
 
-    private fun loadShuffledWords(): Single<List<String>> =
+    private fun loadWordsToDisplay(): Single<List<String>> =
         Single.fromCallable {
-            mnemonic.words().shuffled()
+            mnemonic.words().sorted()
```
fmrsabino: 😆

PR #206, opened 2018-07-17
WA-447: Update screen view tracking

fmrsabino:
Closes WA-447.

- Update screen view tracking
- Disable Advertising ID collection

@gnosis/mobile-devs


app/src/main/java/pm/gnosis/heimdall/reporting/EventTracker.kt
```diff
@@ -26,32 +26,51 @@ enum class ScreenId {
     ADDRESS_BOOK_EDIT_ENTRY,
     ADDRESS_BOOK_ENTRY,
     ADDRESS_BOOK_ENTRY_DETAILS,
-    CHECK_SAFE,
-    CONFIRM_SAFE_RECOVERY_PHRASE,
-    CONFIRM_TRANSACTION,
-    CREATE_ASSET_TRANSFER,
-    CREATE_SAFE_INTRO,
+    INPUT_SAFE_ADDRESS,
```
rmeissner: out of order ;)

app/src/main/java/pm/gnosis/heimdall/ui/safe/details/transactions/SafeTransactionsFragment.kt
```diff
@@ -49,6 +57,8 @@ class SafeTransactionsFragment : BaseFragment(), ScrollableContainer {
 
     override fun onStart() {
         super.onStart()
+        eventTracker.submit(Event.ScreenView(ScreenId.SAFE_TRANSACTION_LIST))
```
rmeissner: This is not correct, as it will be triggered even if the tab is not selected (fragment gets created and initialized in the background).

Either we leave it out for now or you hook it to `getUserVisibilityHint` https://stackoverflow.com/questions/9323279/how-to-test-if-a-fragment-view-is-visible-to-the-user

app/src/main/java/pm/gnosis/heimdall/ui/tokens/balances/TokenBalancesFragment.kt
```diff
@@ -61,6 +68,7 @@ class TokenBalancesFragment : BaseFragment() {
 
     override fun onStart() {
         super.onStart()
+        eventTracker.submit(Event.ScreenView(ScreenId.SAFE_ASSETS_VIEW))
```
rmeissner: Same as above

PR #234, opened 2018-09-12
Closes #228: Prevent flashing on fresh start

### LINKED ISSUE #228: Target screen is briefly shown before the Unlock screen pops up
### Expected behavior

When unlocking the app the unlock screen should show up without any "flashing" (ie.: target screen being briefly shown)

### Actual behavior

When I unlock the app there is a short moment where you see a white background. Preferably this would not happen

### Steps to reproduce the behaviour

Open the application whenever it is on a locked state.

### Additional information

- Device: Pixel 2
- Android OS version: 9
- Safe app version: 0.6.3

rmeissner:
Closes #228.

- If app is locked on splash start unlock screen to prevent "white flash"

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/pm/gnosis/heimdall/ui/splash/SplashActivity.kt
+++ b/app/src/main/java/pm/gnosis/heimdall/ui/splash/SplashActivity.kt
@@ -14,6 +14,7 @@ import pm.gnosis.heimdall.reporting.ScreenId
 import pm.gnosis.heimdall.ui.base.BaseActivity
 import pm.gnosis.heimdall.ui.onboarding.OnboardingIntroActivity
 import pm.gnosis.heimdall.ui.safe.main.SafeMainActivity
+import pm.gnosis.heimdall.ui.security.unlock.UnlockActivity
 import pm.gnosis.svalinn.common.utils.startActivity
 import timber.log.Timber
 import javax.inject.Inject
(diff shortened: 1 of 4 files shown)


app/src/main/java/pm/gnosis/heimdall/ui/splash/SplashActivity.kt
```diff
@@ -43,6 +44,7 @@ class SplashActivity : BaseActivity() {
     private fun handleAction(action: ViewAction) {
         when (action) {
             is StartMain -> startMain()
+            is StartUnlock -> startUnlock()
```
fmrsabino: If the action is `StartUnlock` then after unlocking it goes back to this activity (Splash)

rmeissner: Yes, it will then check again and start main if it was unlocked

app/src/main/java/pm/gnosis/heimdall/ui/splash/SplashViewModel.kt
```diff
@@ -21,21 +20,28 @@ class SplashViewModel @Inject constructor(
...
-            .onErrorReturn {
+            .flatMap { checkUnlocked() }
+            .onErrorResumeNext {
                 when (it) {
                     is EmptyResultSetException, is NoSuchElementException ->
-                        StartPasswordSetup()
+                        Single.just(StartPasswordSetup)
                     else ->
-                        StartMain()
+                        checkUnlocked()
```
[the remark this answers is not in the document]
rmeissner: CheckUnlocked cannot fail

app/src/main/java/pm/gnosis/heimdall/ui/splash/SplashActivity.kt
fmrsabino: 👌

fmrsabino:
There are other scenarios where the flashing can occur thus not completely closing #228 but maybe we can create a follow up issue.

For example when we open the app via a push:
https://github.com/gnosis/safe-android/blob/a2f2bb9c1021788d4b9dbf5ed9bb0ff17e3c4a02/app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/DefaultPushServiceRepository.kt#L217-L226

rmeissner:
Yeah it only fixes this for a fresh start (which I think is already an improvement :P )

fmrsabino:
We can close it. I think it'd only happen in this scenario and is nothing critical 😉

PR #240, opened 2018-09-25
Closes #238: Add in-memory cache for nonce

### LINKED ISSUE #238: Add nonce cache
**Is your feature request related to a problem? Please describe.**
When triggering multiple transactions within a short time period the nonce will be incorrect.

**Describe the solution you'd like**
The nonce for each contract should be tracked in-memory. This will allow to detect if the nonce has not been propagated yet and make the required adjustments.

**Describe alternatives you've considered**
Alternatively it would be possible to add a delay between each transaction (or constantly update the nonce in the signing screen). As this would potentially invalidate signatures the complexity is higher.

**Additional context**
I would not use persisted storage, as due to reorgs the nonce might decrease. When caching in-memory, worst case would be that you have to restart the app. If persisted you would need to reset the app or wait for the nonce to be correct again.

rmeissner:
Closes #238.

- Add in-memory cache for last used nonce
  - The cache will be used if the nonce returned by the remote node is lower than or equal to the cached nonce. In this case the next nonce based on the cached nonce will be used.

@gnosis/mobile-devs


gradle.properties
```diff
@@ -11,4 +11,4 @@ org.gradle.jvmargs=-Xmx8g
 # This option should only be used with decoupled projects. More details, visit
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
 # org.gradle.parallel=true
-android.enableD8MainDexList=false
+android.enableD8MainDexList=true
```
fmrsabino: Maybe we can remove this line. [Default option](https://android.googlesource.com/platform/tools/base/+/studio-master-dev/build-system/gradle-core/src/main/java/com/android/build/gradle/options/BooleanOption.java?autodive=0%2F%2F#48) seems to be set to `true`

PR #328, opened 2019-02-08
Handle creation push

rmeissner:
- This will show a notification when a safe was deployed

@gnosis/mobile-devs


fmrsabino:
Shall we hold this PR until the chrome extension also implements it and then we'd remove the sending push automatically once we go from "Pending Safe" to "Safe"?

rmeissner:
The extension has this already (before the push was send by the client). We only need to check with iOS

PR #382, opened 2019-06-04
Adjust transaction list headers (#370)

### LINKED ISSUE #370: Adjust transaction list headers
The entries of the transaction list will only contain the time, therefore the header between entries should correspond to the data (no grouping of multiple dates).

See https://github.com/gnosis/safe/blob/master/specs/transaction_list/02_user_interface.rst#1transaction-list

elgatovital:
Closes #370: Adjust transaction list headers

- the header between entries correspond to the data (no grouping of multiple dates)

@gnosis/mobile-devs

### DIFF
--- a/app/src/test/java/pm/gnosis/heimdall/ui/safe/details/transactions/AdapterEntryHeaders.kt
+++ b/app/src/test/java/pm/gnosis/heimdall/ui/safe/details/transactions/AdapterEntryHeaders.kt
@@ -1,0 +1,7 @@
+package pm.gnosis.heimdall.ui.safe.details.transactions
+
+const val PENDING = "pending"
+const val SUBMITTED = "submitted"
+const val TODAY = "today"
+const val YESTERDAY = "yesterday"
+const val DATE = " date"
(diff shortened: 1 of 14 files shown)


app/src/test/java/pm/gnosis/heimdall/ui/safe/details/transactions/AdapterEntryHeaders.kt
```diff
@@ -0,0 +1,7 @@
+package pm.gnosis.heimdall.ui.safe.details.transactions
+
+const val PENDING = "pending"
```
rmeissner: What are these used for? Are they only for testing?

elgatovital: Yes

rmeissner: I would move them to the test folder in this case, else it looks good

elgatovital: Ok

PR #386, opened 2019-06-06
Feature/issue 369

### LINKED ISSUE #369: Use separate owner for each safe
Currently for each Safe that is managed by the app (=> app is an owner) the same private key is used. We should use a different key for each Safe.

Possible solutions:
- We create a new table (safe_information) that contains a mapping from Safe address to key index
- We create a new device mnemonic that we store encrypted (additional to current global key)
- If we need to sign something for a Safe we load the key index 
- If no index exists (Safes created before the change) we use the existing global private key
- Else we use a the mnemonic + the index to generate the private key

Notes:
- How is the performance for a high index (since this is an iterative process)?
  - Additional to the index we could store the private key encrypted in the safe_information table

elgatovital:
Closes #369: Use separate owners for safes

- Create separate owner for each safe instead of using device account
- Sign transaction using safes owner
- If no owner present (old safe) sign using device account

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/pm/gnosis/heimdall/data/db/SolidityAddressConverter.kt
+++ b/app/src/main/java/pm/gnosis/heimdall/data/db/SolidityAddressConverter.kt
@@ -7,8 +7,8 @@ import pm.gnosis.utils.asEthereumAddressString
 
 class SolidityAddressConverter {
     @TypeConverter
-    fun fromHexString(address: String) = address.asEthereumAddress()!!
+    fun fromHexString(address: String?) = address?.asEthereumAddress()!!
 
     @TypeConverter
-    fun toHexString(address: Solidity.Address): String = address.asEthereumAddressString()
+    fun toHexString(address: Solidity.Address?): String? = address?.asEthereumAddressString()
 }
--- a/app/src/main/java/pm/gnosis/heimdall/data/db/SolidityAddressConverter.kt
+++ b/app/src/main/java/pm/gnosis/heimdall/data/db/SolidityAddressConverter.kt
@@ -7,8 +7,8 @@ import pm.gnosis.utils.asEthereumAddressString
 
 class SolidityAddressConverter {
     @TypeConverter
-    fun fromHexString(address: String?) = address?.asEthereumAddress()!!
+    fun fromHexString(address: String) = address.asEthereumAddress()!!
 
     @TypeConverter
-    fun toHexString(address: Solidity.Address?): String? = address?.asEthereumAddressString()
+    fun toHexString(address: Solidity.Address): String = address.asEthereumAddressString()
 }
--- a/app/src/main/java/pm/gnosis/heimdall/data/db/models/GnosisSafeDb.kt
+++ b/app/src/main/java/pm/gnosis/heimdall/data/db/models/GnosisSafeDb.kt
@@ -27,37 +27,25 @@ fun Safe.toDb() = GnosisSafeDb(address)
 fun GnosisSafeDb.fromDb() = Safe(address)
 
 @Entity(
-    tableName = GnosisSafeOwnerDb.TABLE_NAME,
-    foreignKeys = [
-        ForeignKey(
-            entity = GnosisSafeDb::class,
-            parentColumns = [GnosisSafeDb.COL_ADDRESS],
-            childColumns = [GnosisSafeOwnerDb.COL_SAFE_ADDRESS],
-            onUpdate = ForeignKey.NO_ACTION,
-            onDelete = ForeignKey.CASCADE,
-            deferred = true
-        )
-    ],
-    indices = [
-        Index(value = [GnosisSafeOwnerDb.COL_SAFE_ADDRESS])
-    ]
+    tableName = GnosisSafeInfoDb.TABLE_NAME
 )
-data class GnosisSafeOwnerDb(
+data class GnosisSafeInfoDb(
     @PrimaryKey
-    @ColumnInfo(name = COL_PRIVATE_KEY)
-    var privateKey: EncryptedByteArray,
-
-    @ColumnInfo(name = COL_ADDRESS)
-    val address: Solidity.Address,
-
     @ColumnInfo(name = COL_SAFE_ADDRESS)
-    val safeAddress: Solidity.Address? = null
+    val safeAddress: Solidity.Address,
+
+    @ColumnInfo(name = COL_OWNER_ADDRESS)
+    val ownerAddress: Solidity.Address,
+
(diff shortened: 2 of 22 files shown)


app/src/main/java/pm/gnosis/heimdall/data/db/SolidityAddressConverter.kt
```diff
@@ -7,8 +7,8 @@ import pm.gnosis.utils.asEthereumAddressString
 
 class SolidityAddressConverter {
     @TypeConverter
-    fun fromHexString(address: String) = address.asEthereumAddress()!!
+    fun fromHexString(address: String?) = address?.asEthereumAddress()!!
```
rmeissner: This doesn'T really make sense, since it will always throw if you pass `null` as an address

app/src/main/java/pm/gnosis/heimdall/data/db/models/GnosisSafeDb.kt
```diff
@@ -27,6 +26,41 @@ data class GnosisSafeDb(
...
+            onDelete = ForeignKey.CASCADE,
+            deferred = true
+        )
+    ],
+    indices = [
+        Index(value = [GnosisSafeOwnerDb.COL_SAFE_ADDRESS])
+    ]
+)
+data class GnosisSafeOwnerDb(
+    @PrimaryKey
```
rmeissner: I would use the address as primary key (or if possible the safe since we normally query by the safe address)

app/src/main/java/pm/gnosis/heimdall/ui/safe/create/CreateSafeConfirmRecoveryPhraseViewModel.kt
```diff
@@ -50,6 +68,7 @@ class CreateSafeConfirmRecoveryPhraseViewModel @Inject constructor(
             .flatMap { request -> relayServiceApi.safeCreation(request).map { request to it } }
             .flatMap { (request, response) ->
                 assertResponse(request, response)
+                gnosisSafeRepository.assignOwnerToSafe(request.owners[0], response.safe)
```
rmeissner: I would probably not create the owner in the db until this point.

This would also prevent that we create owner entries without safes (which might occur if we have a network error or the user closes the app).

Also this would allow us to make the safe a primary key in the table

app/src/main/java/pm/gnosis/heimdall/data/db/models/GnosisSafeDb.kt
```diff
@@ -27,6 +26,41 @@ data class GnosisSafeDb(
...
+    var privateKey: EncryptedByteArray,
+
+    @ColumnInfo(name = COL_ADDRESS)
+    val address: Solidity.Address,
+
+    @ColumnInfo(name = COL_SAFE_ADDRESS)
+    val safeAddress: Solidity.Address? = null
+) {
+    companion object {
+        const val TABLE_NAME = "gnosis_safe_owners"
```
rmeissner: I would prefer a table that contains safe information (GnosisSafeInfoDb). Currently that might only be the local key, but in the future we might have other information, so we could easily extend it.

Also I would use a table where we have the safe address as a primary key and the signing key as an encrypted byte array (see other comments on PR).

What do you think?

elgatovital: agree. I will adjust the tables

app/src/main/java/pm/gnosis/heimdall/ui/safe/create/CreateSafeConfirmRecoveryPhraseViewModel.kt
elgatovital: yes, good point

app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/DefaultGnosisSafeRepository.kt
```diff
@@ -304,6 +315,56 @@ class DefaultGnosisSafeRepository @Inject constructor(
             .subscribeOn(Schedulers.io())
             .map { it.map { TransactionStatus(it.id, it.timestamp, false) } }
 
+
+    override fun createOwner(): Single<Pair<Solidity.Address, ByteArray>> {
+        return Single.just(
```
rmeissner: Wouldn't Single.fromCallable be better here?

app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/DefaultGnosisSafeRepository.kt
```diff
@@ -304,6 +315,56 @@ class DefaultGnosisSafeRepository @Inject constructor(
...
+        }.subscribeOn(Schedulers.io())
+
+
+    override fun loadOwnerAddress(safeAddress: Solidity.Address): Single<Solidity.Address> {
+        return safeDao.loadSafeInfo(safeAddress)
+            .map {
+                it.ownerAddress
+            }
+            .onErrorResumeNext {
+                accountsRepository.loadActiveAccount().map { it.address }
```
rmeissner: Should we check that it is an error that the safe doesn'T exist?

app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/DefaultGnosisSafeRepository.kt
```diff
@@ -304,6 +315,56 @@ class DefaultGnosisSafeRepository @Inject constructor(
...
+    }
+
+    override fun sign(safeAddress: Solidity.Address, data: ByteArray): Single<Signature> {
+
+        return safeDao.loadSafeInfo(safeAddress)
+            .map { it.ownerPrivateKey.value(encryptionManager).asBigInteger() }
+            .map { KeyPair.fromPrivate(it) }
+            .map { it.sign(data).let { Signature(it.r, it.s, it.v) } }
+            .onErrorResumeNext {
+                accountsRepository.sign(data)
```
rmeissner: SAme as above

app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/DefaultGnosisSafeRepository.kt
elgatovital: it should happen only for safes created with a previous app version that don't have separate owner (no entry in safe info table). In this case safe exists but has device address as its owner

PR #398, opened 2019-06-19
Closes #384: WalletConnect Design

### LINKED ISSUE #384: Adjust WalletConnect support
See epic https://github.com/gnosis/safe/issues/151

- [x] Add header for active session
- [x] Change list background

rmeissner:
Closes #384.

- WallectConnect session is for a single Safe
- Adjust design for WalletConnect session list 
- Add wallet connect intro

@gnosis/mobile-devs


app/src/main/AndroidManifest.xml
```diff
@@ -247,9 +247,10 @@
             android:theme="@style/WhiteTheme" />
 
         <activity
-            android:name=".ui.walletconnect.WalletConnectSessionsActivity"
+            android:name=".ui.walletconnect.sessions.WalletConnectSessionsActivity"
             android:screenOrientation="portrait"
             android:theme="@style/WhiteTheme" />
+        <activity android:name=".ui.walletconnect.intro.WalletConnectIntroActivity" />
```
elgatovital: Other activities have fixed screen orientation set to portrait

app/src/main/java/pm/gnosis/heimdall/data/repositories/BridgeRepository.kt
```diff
@@ -5,18 +5,21 @@ import io.reactivex.Observable
...
+
 interface BridgeRepository {
     data class SessionMeta(
         val id: String,
         val dappName: String?,
         val dappDescription: String?,
+        val dappUrl: String?,
         val dappIcons: List<String>?,
         val active: Boolean,
         val approvedSafes: List<Solidity.Address>?
```
elgatovital: maybe better to remove nullability for approvedSafes and add helper function/property hasApprovedSafes. Like val hasApprovedSafes = approvedSafes.isNotEmpty()

app/src/main/java/pm/gnosis/heimdall/data/repositories/BridgeRepository.kt
```diff
@@ -31,17 +34,17 @@ interface BridgeRepository {
         ) : SessionEvent()
     }
 
-    fun sessions(): Single<List<SessionMeta>>
+    fun sessions(safe: Solidity.Address?): Single<List<SessionMeta>>
```
elgatovital: why is safe nullable? Is it possible to have a session that is not attached to any safe?

app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/WalletConnectBridgeRepository.kt
```diff
@@ -156,11 +179,12 @@ class WalletConnectBridgeRepository @Inject constructor(
                 override fun sessionApproved() {}
 
                 override fun sessionClosed() {
-                    sessions.remove(config.handshakeTopic)
+                    sessionsPreferences.edit { remove(KEY_PREFIX_SESSION_SAFE + sessionId) }
```
elgatovital: I think it would be better to encapsulate SharedPreferences like this to classes and use getters & setters for editing. Also all respective keys can be kept inside companion object of those classes

app/src/main/java/pm/gnosis/heimdall/ui/walletconnect/intro/WalletConnectIntroActivity.kt
```diff
@@ -0,0 +1,109 @@
...
+import pm.gnosis.heimdall.ui.base.ViewModelActivity
+import pm.gnosis.heimdall.ui.walletconnect.sessions.WalletConnectSessionsActivity
+import pm.gnosis.model.Solidity
+import pm.gnosis.utils.asEthereumAddress
+import pm.gnosis.utils.asEthereumAddressString
+
+
+class WalletConnectIntroActivity : ViewModelActivity<WalletConnectIntroContract>() {
+
+    private val layouts =
```
elgatovital: better to use enum for intro page, as enum values can potentially hold additional useful data if needed. Also it will be more readable

app/src/main/java/pm/gnosis/heimdall/data/repositories/BridgeRepository.kt
rmeissner: null would mean no filter (so display all available sessions)

app/src/main/java/pm/gnosis/heimdall/ui/walletconnect/intro/WalletConnectIntroActivity.kt
rmeissner: wouldn't that be unnecessary overhead here (since we don't have any other information right now?)

app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/WalletConnectBridgeRepository.kt
rmeissner: I think this would be too much for this PR to change. In general a better abstraction from shared preferences would make sense, but I would recommend having this as a separate issue.

app/src/main/java/pm/gnosis/heimdall/data/repositories/BridgeRepository.kt
rmeissner: on low level there is a slight difference between `null` and emptyList here.

Currently we are not using the parameter, so I would rather not change it for now.

app/src/main/java/pm/gnosis/heimdall/data/repositories/BridgeRepository.kt
elgatovital: would it make sence to have 2 methods? One that returns all sessions (Single<List<SessionMeta>>) and one that returns sessionForSafe (Single<SessionMeta>)?

app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/WalletConnectBridgeRepository.kt
elgatovital: agree. https://github.com/gnosis/safe-android/issues/399

PR #401, opened 2019-06-21
Closes #379: WalletConnect deeplinks

### LINKED ISSUE #379: Add wallect connect deeplink handling
It should be possible to connect to the Safe app via WalletConnect by using a deeplink.

# TODO
- [x] Add Safe selection for WalletConnect
- [x] Add deeplink handler to point to Safe selection

Code can be found at: https://github.com/gnosis/safe-android/commit/467706d23360a265c26c7c2465d4e2c30ab88c1e

rmeissner:
Closes #379.

- Handle deeplinks for wallet connect
- Add network policy
- Add missing network policy xml

@gnosis/mobile-devs


app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/WalletConnectBridgeRepository.kt
```diff
@@ -224,21 +226,29 @@ class WalletConnectBridgeRepository @Inject constructor(
...
-            .setSubText(safe.shortChecksumString())
-            .setLargeIcon(icon)
-            .build()
-        localNotificationManager.show(
-            referenceId.hashCode(),
-            notification
-        )
+        // Pre Android Q we will directly show the review activity if the phone is unlocked, else we show a notification
+        // TODO: Adjust check when Q is released
+        if (BuildCompat.isAtLeastQ() || Build.VERSION.SDK_INT > Build.VERSION_CODES.P || keyguard.isKeyguardLocked) {
```
elgatovital: isn't ```BuildCompat.isAtLeastQ() || Build.VERSION.SDK_INT > Build.VERSION_CODES.P ``` is the same as just ```BuildCompat.isAtLeastQ()```?

app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/WalletConnectBridgeRepository.kt
```diff
@@ -224,21 +226,29 @@ class WalletConnectBridgeRepository @Inject constructor(
...
+            )
+                .setSubText(safe.shortChecksumString())
+                .setLargeIcon(icon)
+                .build()
+            localNotificationManager.show(
+                referenceId.hashCode(),
+                notification
+            )
+        } else {
+            intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
```
elgatovital: should we add this flag to createIntent of ReviewTransactionActivity?

app/src/main/res/layout/screen_wallet_connect_safe_selection.xml
```diff
@@ -0,0 +1,66 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.coordinatorlayout.widget.CoordinatorLayout
```
elgatovital: Why do we have this additional CoordinatorLayout? Is it not enough just to have ConstraintLayout as root here?

app/src/main/res/xml/network_security_config.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<network-security-config>
+    <domain-config cleartextTrafficPermitted="true">
```
elgatovital: why do we need cleartext traffic? This might be a security issue

app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/WalletConnectBridgeRepository.kt
rmeissner: no sadly not .. `BuildCompat.isAtLeastQ()` uses some magic config vars from the device and will not work once Q is released ... for this check this would be a problem, therefore this double check.

app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/WalletConnectBridgeRepository.kt
rmeissner: I only add this flag, because else it will not work (context is a service so the flag is required). Normally I would prefer not to have this flag here at all.

app/src/main/res/layout/screen_wallet_connect_safe_selection.xml
rmeissner: this coordinator layout makes snackbars dismisable (via swipe)

app/src/main/res/xml/network_security_config.xml
rmeissner: this is only trafic toward localhost, therefore the security risk is extremely low, since you would have a network sniffer on your defice to see this traffic ... for this you require root ... if somebody got this far everything is lost anyways ... right?

rmeissner: Also this was not enforced on all android version, had to add it for newer versions. 

Why we want this: This allows us to directly communicate via http to other apps on this device (e.g. a WalletConnect server embedded into an app). This is safes data compared to routing via an external service. But normally these apps don't ship their own ssl certificate (would also not make much sense, since you would kind of expose the secret part of it). Therefore we allow http for localhost.

PR #405, opened 2019-06-26
Closes #403: Update restricted transactions

### LINKED ISSUE #403: Update restricted transactions
Currently we prevent that transactions that perform a delegate call, or call one of the management functions are triggered from outside the app.

These checks are not 100% accurate as they prevent to call these methods on any contract (not just on the current Safe) therefore preventing management of other contracts that use the same methods (e.g. other Safes)

### TODO
- [ ] Only prevent transaction to the Safe itself (vs management tx to any contract)
- [ ] Check this for wallet connect transaction requests

rmeissner:
Closes #403.

- Update check for restricted transactions
- Check restricted tx for WC
- 

@gnosis/mobile-devs


app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/WalletConnectBridgeRepository.kt
```diff
@@ -179,9 +179,17 @@ class WalletConnectBridgeRepository @Inject constructor(
...
-                                    .flatMap { (safe, tx) -> infoRepository.parseTransactionData(tx).map { txData -> safe to txData } }
+                                    .flatMap { (safe, tx) ->
+                                        infoRepository.checkRestrictedTransaction(safe, tx)
+                                            .flatMap(infoRepository::parseTransactionData)
+                                            .map { txData -> safe to txData }
+                                    }
                                     .subscribeBy(onError = { t ->
-                                        rejectRequest(id, 42, t.message ?: "Could not handle transaction").subscribe()
+                                        val message = when (t) {
+                                            is RestrictedTransactionException -> "This transaction is not allowed"
```
elgatovital: should we add these messages to strings.xml?

rmeissner: I left it out for now, since these are errors normally shown only on the developer console (in the dapp) but might make sense ... not sure :P

PR #439, opened 2019-08-08
Change launcher icons

### LINKED ISSUE #433: Rebranding: Change launcher icons

elgatovital:
Closes #433.

- Change launcher icons

@gnosis/mobile-devs


rmeissner:
Nice icons.
Some questions:
- Is it on purpose that for the rinkeby icon the banderole is not green
- Shouldn't we have a background for the icon instead of having it transparent? (How does this look like on older devices that don't automatically add the circle background?)

elgatovital:
> Nice icons.
> Some questions:
> 
> * Is it on purpose that for the rinkeby icon the banderole is not green
> * Shouldn't we have a background for the icon instead of having it transparent? (How does this look like on older devices that don't automatically add the circle background?)

updated rinkeby internal & dev launcher icons: green banderole
old launcher icons also had transparent background

rmeissner:
> old launcher icons also had transparent background.

We had a larger icon (the outline circle) and the white parts where not transparent, but white afaik, but yeah we should just check ;)

PR #450, opened 2019-08-26
Fix #443: Change modal title

### LINKED ISSUE #443: Incorrect action sheet title.
safe1

This action sheet has the title 'Choose Recipient' when you tap on 'recover existing Safe'. Title doesn't seem accurate in this context. 

Using the latest version from the Play store (Android).

A quick fix could be to rename the title to "Choose address" or something. As that would apply to all situations. Although having the title be specific to its context is desired in my opinion.

rmeissner:
Closes #443.

- Change title of address input selection dialog

@gnosis/mobile-devs


germartinez:
Should this be removed?
https://github.com/gnosis/safe-android/blob/de0f350c5f2f2b7c92420045880a485aa0933779/app/src/main/res/values/strings.xml#L414

PR #469, opened 2019-09-16
Closes #462:  Create custom progress step indicator

### LINKED ISSUE #462: KeyCard: Custom onboarding progress indicator
Create custom view indicating current step of the keycard setup

elgatovital:
Closes #462: KeyCard: create a custom progress step indicator.

- Create custom progress step indicator.

@gnosis/mobile-devs


app/src/main/java/pm/gnosis/heimdall/views/StepIndicator.kt
```diff
@@ -0,0 +1,331 @@
...
+) : LinearLayout(context, attrs, defStyleAttr) {
+
+    private val backgroundPaint: Paint
+
+    init {
+        orientation = HORIZONTAL
+        gravity = Gravity.CENTER
+        backgroundPaint = Paint(Paint.ANTI_ALIAS_FLAG)
+        backgroundPaint.style = Paint.Style.FILL_AND_STROKE
+        backgroundPaint.strokeWidth = TypedValue.applyDimension(TypedValue.COMPLEX_UNIT_DIP, 2f, resources.displayMetrics)
```
rmeissner: should we put this in our dimen.xml?

app/src/main/java/pm/gnosis/heimdall/views/StepIndicator.kt
```diff
@@ -0,0 +1,331 @@
...
+        for (i in 0 until childCount - 1) {
+
+            val child1 = getChildAt(i) as Step
+            val child2 = getChildAt(i + 1) as Step
+
+            val colorPair =
+                if (child1.state == StepState.UNCOMPLETED_INACTIVE || child1.state == StepState.SKIPPED) {
+
+                    if (child2.state == StepState.UNCOMPLETED_INACTIVE || child2.state == StepState.SKIPPED)
+                        Pair(
```
rmeissner: you can create pairs with `x to b` in kotlin

app/src/main/java/pm/gnosis/heimdall/views/StepIndicator.kt
```diff
@@ -0,0 +1,331 @@
...
+                val childWidth = child.measuredWidth
+                val childHeight = child.measuredHeight
+
+                child.layout(left + childsMargin / 2, paddingTop, left + childWidth + childsMargin / 2, childHeight + paddingTop)
+                left += childWidth + childsMargin / 2
+            }
+        }
+    }
+
+    override fun dispatchDraw(canvas: Canvas) {
```
rmeissner: what is the difference between `dispatchDraw` and `onDraw`?

elgatovital: its called in viewGroup before childs are drawn

app/src/main/java/pm/gnosis/heimdall/views/StepIndicator.kt
elgatovital: yes

app/src/main/java/pm/gnosis/heimdall/views/StepIndicator.kt
rmeissner: also wouldn't it be nicer to have something like:
```
private fun StepState.toColor() = ContextCompat.getColor(
  if (this == StepState.UNCOMPLETED_INACTIVE || this == StepState.SKIPPED) context, R.color.disabled_button else R.color.safe_green
)
```
and call 
```
child1.state.toColor() to child2.state.toColor()
```

app/src/main/java/pm/gnosis/heimdall/views/StepIndicator.kt
```diff
@@ -0,0 +1,331 @@
...
+                    else
+                        Pair(
+                            ContextCompat.getColor(context, R.color.safe_green),
+                            ContextCompat.getColor(context, R.color.safe_green)
+                        )
+                }
+
+
+            val shader = LinearGradient(
+                child1.getCircleCenter().x + child1.radius(),
```
rmeissner: Would probably be nice to somehow cache this, since you are creating an object each time you call this, which is a bad practice in `onDraw`

rmeissner: https://developer.android.com/training/custom-views/custom-drawing#createobject

rmeissner: Same actually for the color pair

app/src/main/java/pm/gnosis/heimdall/views/StepIndicator.kt
```diff
@@ -0,0 +1,346 @@
...
+
+        var colorPair: Pair<Int, Int>
+
+        for (i in 0 until childCount - 1) {
+
+            val child1 = getChildAt(i) as Step
+            val child2 = getChildAt(i + 1) as Step
+
+            colorPair =
+                if (child1.state == StepState.UNCOMPLETED_INACTIVE || child1.state == StepState.SKIPPED) {
```
rmeissner: This can be optimized, but it is fine for now.

E.g. the color of a child doesn't depend on any other child, so you only need 2 colors (active and inactive) and use these in the edge shader

elgatovital: good Point

PR #486, opened 2019-11-13
Batched transactions

rmeissner:
- Allow to execute batched transctions via wallet connect (`gs_multi_send`)
- Update wallet connect lib

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/WalletConnectBridgeRepository.kt
+++ b/app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/WalletConnectBridgeRepository.kt
@@ -28,6 +28,7 @@ import pm.gnosis.heimdall.R
 import pm.gnosis.heimdall.data.preferences.PreferencesWalletConnect
 import pm.gnosis.heimdall.data.repositories.*
 import pm.gnosis.heimdall.data.repositories.BridgeRepository.Companion.MULTI_SEND_RPC
+import pm.gnosis.heimdall.data.repositories.BridgeRepository.RejectionReason
 import pm.gnosis.heimdall.data.repositories.models.SafeTransaction
 import pm.gnosis.heimdall.di.ApplicationContext
 import pm.gnosis.heimdall.helpers.LocalNotificationManager
@@ -68,7 +69,7 @@ class WalletConnectBridgeRepository @Inject constructor(
     }
 
     override fun onTransactionRejected(referenceId: Long) {
-        sessionForRequest(referenceId)?.rejectRequest(referenceId, 4567, "Transaction rejected")
+        rejectRequest(referenceId, RejectionReason.Rejected).subscribe()
     }
 
     override fun onTransactionSubmitted(safeAddress: Solidity.Address, transaction: SafeTransaction, chainHash: String, referenceId: Long?) {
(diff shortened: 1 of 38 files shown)


app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/WalletConnectBridgeRepository.kt
```diff
@@ -243,6 +225,61 @@ class WalletConnectBridgeRepository @Inject constructor(
             config.handshakeTopic
         }
 
+    private fun handleCustomCall(sessionId: String, session: Session, call: Session.MethodCall.Custom) {
+        sessionRequests[call.id] = sessionId
+        try {
+            when (call.method) {
+                "gs_multi_send" -> {
```
elgatovital: should we extract it to enum or some const?

app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/WalletConnectBridgeRepository.kt
```diff
@@ -243,6 +225,61 @@ class WalletConnectBridgeRepository @Inject constructor(
...
+                    handleMultiSend(sessionId, session, call)
+                }
+                else ->
+                    call.apply {
+                        rpcProxyApi.proxy(RpcProxyApi.ProxiedRequest(method, (params as? List<Any>) ?: emptyList(), id))
+                            .subscribeBy(onError = { t ->
+                                rejectRequest(id, 42, t.message ?: "Could not handle custom call")
```
elgatovital: would it make sense to extract error codes? Maybe enum of some sort with comments. Those numbers alone are not very informative

app/src/main/java/pm/gnosis/heimdall/helpers/AddressHelper.kt
```diff
@@ -55,4 +61,8 @@ class AddressHelper @Inject constructor(
         imageView?.setAddress(address)
         return listOf(buildAddressInfoSingle(addressView, nameView, address).subscribeBy(onError = Timber::e))
     }
+
+    companion object {
+        private val MULTI_SEND_LIB = BuildConfig.MULTI_SEND_ADDRESS.asEthereumAddress()!!
```
elgatovital: is multisend a lib or smart contract?

rmeissner: https://github.com/gnosis/safe-contracts/blob/development/contracts/libraries/MultiSend.sol

It is a smart contract that is a library for the Safe (see https://gnosis-safe.readthedocs.io/en/latest/contracts/architecture.html#libraries)

app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/WalletConnectBridgeRepository.kt
rmeissner: We could put it into a const, enum doesn't really make sense imo

app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/WalletConnectBridgeRepository.kt
rmeissner: I think it is a random number (maybe 0 would be better), this can be improved, but is more walletconnect related than batched transactions

app/src/test/java/pm/gnosis/heimdall/data/repositories/impls/WalletConnectBridgeRepositoryIntegrationTest.kt
```diff
@@ -108,7 +108,7 @@ class WalletConnectBridgeRepositoryIntegrationTest {
 
     private fun rejectRequest(repo: WalletConnectBridgeRepository, requestId: Long, errorMessage: String) {
         System.out.println("Reject request")
-        repo.rejectRequest(requestId, 42, errorMessage).subscribeBy(
+        repo.rejectRequest(requestId, BridgeRepository.RejectionReason.RPCError(42, errorMessage)).subscribeBy(
```
elgatovital: 42 error code should be also encapsulated in the error class

PR #498, opened 2019-11-19
Normalize ENS name

### LINKED ISSUE #497: Normalize ENS names
See https://docs.ens.domains/contract-api-reference/name-processing#normalising-names

For reference web3j implementation: https://github.com/web3j/web3j/blob/116539fff875a083c896b2d569d17416dfeb8a6f/core/src/main/java/org/web3j/ens/NameHash.java

rmeissner:
Closes #497.

- Normalize ENS names

### TODOs:
- [x] Fix tests DefaultEnsRepository
- [x] Add test for IDNEnsNormalizer

### DIFF
--- a/app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/DefaultEnsRepository.kt
+++ b/app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/DefaultEnsRepository.kt
@@ -10,17 +10,20 @@ import pm.gnosis.model.Solidity
 import pm.gnosis.models.Transaction
 import pm.gnosis.utils.asEthereumAddress
 import pm.gnosis.utils.toHexString
+import java.net.IDN
+import java.util.*
 import javax.inject.Inject
 import javax.inject.Singleton
 
 // TODO Maybe move to Svalinn
 @Singleton
 class DefaultEnsRepository @Inject constructor(
+    private val ensNormalizer: EnsNormalizer,
     private val ethereumRepository: EthereumRepository
 ) : EnsRepository {
     override fun resolve(url: String): Single<Solidity.Address> =
         Single.fromCallable {
-            url.split(".").foldRight<String, ByteArray?>(null) { part, node ->
+            ensNormalizer.normalize(url).split(".").foldRight<String, ByteArray?>(null) { part, node ->
                 if (node == null && part.isEmpty()) ByteArray(32)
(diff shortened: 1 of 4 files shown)


app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/DefaultEnsRepository.kt
```diff
@@ -42,3 +45,12 @@ class DefaultEnsRepository @Inject constructor(
         private const val GET_RESOLVER = "0x0178b8bf"
     }
 }
+
```
DmitryBespalov: I don't see it here (maybe just don't understand the code), but we need to make sure that if the resolver address is zero, or if the resolved address is zero, then it is an error - we should show that address or resolver is not found, to prevent sending to the 0x000 address

app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/DefaultEnsRepository.kt
```diff
@@ -42,3 +45,12 @@ class DefaultEnsRepository @Inject constructor(
...
+interface EnsNormalizer {
+    fun normalize(name: String): String
+}
+
+@Singleton
+class IDNEnsNormalizer @Inject constructor() : EnsNormalizer {
+    override fun normalize(name: String) = IDN.toASCII(name, IDN.USE_STD3_ASCII_RULES).toLowerCase(Locale.getDefault())
```
DmitryBespalov: the IDN.toASCII can throw IllegalArgumentException if the input is invalid domain name (for example, starts with two hyphens "--"). This will be handled, right?

app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/DefaultEnsRepository.kt
rmeissner: Sending any call to the 0x0 address will fail and therefore have the same result currently (that no address could be resolved).

More detailed error messages can be adjusted in a future update, but not as part of this PR

app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/DefaultEnsRepository.kt
rmeissner: Errors are propagated

PR #504, opened 2019-12-02
Closes #413: Select payment token when recovering safe

### LINKED ISSUE #413: Token Payment: Recover Safe
In the process of recovering a Safe it should be possible to change the payment token for this Safe.

See epic.

### ToDos:
- [ ] Adjust recover Safe flow to allow changing token via token payment selection screen with tx estimation (#411)

elgatovital:
Closes #413.

- Add payment token selection screen to recovery flow

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/pm/gnosis/heimdall/ui/safe/recover/safe/submit/RecoveringSafeSubmitFragment.kt
+++ b/app/src/main/java/pm/gnosis/heimdall/ui/safe/recover/safe/submit/RecoveringSafeSubmitFragment.kt
@@ -4,9 +4,9 @@ import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
-import com.jakewharton.rxbinding2.view.clicks
 import io.reactivex.android.schedulers.AndroidSchedulers
 import io.reactivex.rxkotlin.plusAssign
+import io.reactivex.rxkotlin.subscribeBy
 import kotlinx.android.synthetic.main.layout_recovering_safe_submit.*
 import pm.gnosis.heimdall.R
 import pm.gnosis.heimdall.data.repositories.models.RecoveringSafe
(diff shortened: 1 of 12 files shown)


app/src/main/java/pm/gnosis/heimdall/ui/safe/recover/safe/submit/RecoveringSafeSubmitFragment.kt
```diff
@@ -45,65 +53,86 @@ class RecoveringSafeSubmitFragment : BaseFragment() {
...
-                layout_recovering_safe_submit_button.isEnabled = true
-                layout_recovering_safe_submit_retry.isEnabled = true
+                errorSnackbar(recover_safe_bottom_panel, it)
+                onLoading(false)
             })
     }
 
+    override fun onResume() {
+        super.onResume()
+        loadExecuteInfo()
```
rmeissner: Not sure if it is good to have this in `onResume` (might not be a big issue, but we only clear the disposable in `onStop`, not in `onPause`)

rmeissner: nvm, thought this is a long running observable .... but this is only a single, so it doesn't matter

app/src/main/java/pm/gnosis/heimdall/ui/tokens/payment/PaymentTokensActivity.kt
```diff
@@ -66,6 +73,20 @@ class PaymentTokensActivity : ViewModelActivity<PaymentTokensContract>() {
...
+            viewModel.paymentToken.observe(this, Observer {
+                payment_token_bottom_panel.forwardLabel = getString(R.string.pay_with, it.symbol)
+            })
+        }
+    }
+
+    override fun onStart() {
+        super.onStart()
+        disposables += payment_token_bottom_panel.forwardClicks.subscribeBy {
+            startActivity(intent.getParcelableExtra(EXTRA_NEXT_ACTION) as Intent)
```
rmeissner: I would feel safer if we check if `EXTRA_NEXT_ACTION` exists or returns `null` else it will crash here

elgatovital: if it is null, bottom panel is gone. But extra check won't hurt

PR #479, opened 2019-10-07
Fix misspelled 'create password' hint on create password screen

alexjg:
Changes proposed in this pull request:
- Fixes a tiny spelling error on the create password screen, the input textbox hint was "Create passwrod". Obviously it's a pretty inconsequential change but it makes the app feel unfinished.

@gnosis/mobile-devs

PR #505, opened 2019-12-02
Closes #479: Correct typo

elgatovital:
Closes #479.

- Fix misspelled "create password" hint 

@gnosis/mobile-devs


rmeissner:
was this updated in localize, or was this only an error in the android xml?

elgatovital:
currently, this string is missing in localize. I will add it now

PR #516, opened 2019-12-19
Feature/506 tx deeplink

### LINKED ISSUE #506: Add support to trigger transactions via deeplinks
It should be possible to create transactions via a deeplink.

It should follow https://github.com/ethereum/EIPs/blob/master/EIPS/eip-681.md (also the `gnosissafe` and `gs` scheme should be registered for the same format)

If no `from` is specified it is required to select a Safe.
If only 1 Safe is available use that, else show Safe selection.

rmeissner:
Closes #506.

- Add support to trigger transactions via ERC-67 deeplinks

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/pm/gnosis/heimdall/ui/deeplinks/DeeplinkActivity.kt
+++ b/app/src/main/java/pm/gnosis/heimdall/ui/deeplinks/DeeplinkActivity.kt
@@ -13,7 +13,10 @@ import kotlinx.android.synthetic.main.screen_wallet_connect_safe_selection.*
 import kotlinx.coroutines.CoroutineExceptionHandler
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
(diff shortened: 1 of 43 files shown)


app/src/main/java/pm/gnosis/heimdall/ui/deeplinks/DeeplinkActivity.kt
```diff
@@ -0,0 +1,252 @@
...
+import java.math.BigInteger
+import javax.inject.Inject
+import javax.inject.Singleton
+
+interface DeeplinkTransactionParser {
+    fun parse(link: String): Pair<SafeTransaction, String?>
+}
+
+@Singleton
+class AndroidDeeplinkTransactionParser @Inject constructor() : DeeplinkTransactionParser {
```
elgatovital: Would ERC67DeeplinkTransactionParser be a better name here?

PR #525, opened 2020-01-09
Closes #495: Remove Authenticator as 2FA option on creation

### LINKED ISSUE #495: Remove Authenticator as 2FA option on creation
See https://github.com/gnosis/safe/issues/253

### TODOs:
- [x] Show user that NFC is not enable
- [x] Do not display Authenticator in creation flow

elgatovital:
Closes #495.

- Add Get In Touch screen
- Use Get In Touch instead of feedback email in general settings
- Remove Authenticator option on creation
- Allow to select Authenticator when recovering safe

@gnosis/mobile-devs


app/src/main/java/pm/gnosis/heimdall/ui/safe/recover/safe/RecoverSafe2FAActivity.kt
```diff
@@ -2,15 +2,26 @@ package pm.gnosis.heimdall.ui.safe.recover.safe
...
 import pm.gnosis.model.Solidity
+import pm.gnosis.svalinn.common.utils.visible
 import pm.gnosis.utils.asEthereumAddress
 import pm.gnosis.utils.asEthereumAddressString
 
 class RecoverSafe2FAActivity : Select2FaActivity() {
 
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        //FIXME: allow to select Gnosis Authenticator when recovering for now
```
rmeissner: Why is that a `FIXME` ... makes it sound like a bug :P

app/src/main/java/pm/gnosis/heimdall/ui/settings/general/GeneralSettingsActivity.kt
```diff
@@ -72,7 +72,7 @@ class GeneralSettingsActivity : ViewModelActivity<GeneralSettingsContract>() {
 
         disposables += layout_general_settings_feedback_background.clicks()
             .subscribeBy(onNext = {
-                createFeedback()
```
rmeissner: can we remove `createFeedback`?

app/src/main/java/pm/gnosis/heimdall/ui/settings/general/GetInTouchActivity.kt
```diff
@@ -0,0 +1,78 @@
...
+
+        var intent = try {
+            try {
+                //Check for Telegram Messenger App
+                packageManager.getPackageInfo("org.telegram.messenger", 0)
+            } catch (e: Exception) {
+                //Check for Telegram X App
+                packageManager.getPackageInfo("org.thunderdog.challegram", 0)
+            }
+            Intent(Intent.ACTION_VIEW, Uri.parse("tg://resolve?domain=${getString(R.string.gnosis_telegram_id)}"))
```
rmeissner: I would actually just try to start this url and if no app can handle it, it will throw an error anyways.

Also can't the telegram app handle the `www.telegram.me` deeplink?

app/src/main/java/pm/gnosis/heimdall/ui/two_factor/Select2FaActivity.kt
```diff
@@ -38,16 +50,41 @@ open class Select2FaActivity : NfcActivity() {
...
+                startActivity(GetInTouchActivity.newIntent(this))
+            }
+        }
+
+        select_authenticator_keycard_description.apply {
+            val linkDrawable = ContextCompat.getDrawable(context, R.drawable.ic_external_link)!!
+            linkDrawable.setBounds(0, 0, linkDrawable.intrinsicWidth, linkDrawable.intrinsicHeight)
+            text = SpannableStringBuilder()
+                .appendText(getString(R.string.status_keycard_description),
+                    ForegroundColorSpan(Color.argb(if(nfcAvailable) 255 else 99, Color.red(currentTextColor), Color.green(currentTextColor), Color.blue(currentTextColor))))
```
rmeissner: Can we please put colors into the resources and load them from there

app/src/main/java/pm/gnosis/heimdall/ui/two_factor/Select2FaActivity.kt
```diff
@@ -38,16 +50,41 @@ open class Select2FaActivity : NfcActivity() {
...
+            select_authenticator_keycard_nfc.visible(true)
+            nfc_required_get_in_touch.setOnClickListener {
+                startActivity(GetInTouchActivity.newIntent(this))
+            }
+        }
+
+        select_authenticator_keycard_description.apply {
+            val linkDrawable = ContextCompat.getDrawable(context, R.drawable.ic_external_link)!!
+            linkDrawable.setBounds(0, 0, linkDrawable.intrinsicWidth, linkDrawable.intrinsicHeight)
+            text = SpannableStringBuilder()
```
rmeissner: We have a `setupLink` extension, this could probably be used (with some small adjustments)

app/src/main/java/pm/gnosis/heimdall/ui/two_factor/Select2FaActivity.kt
```diff
@@ -62,10 +99,12 @@ open class Select2FaActivity : NfcActivity() {
         startActivityForResult(intent, AUTHENTICATOR_REQUEST_CODE)
     }
 
-    private fun onSelected(type: AuthenticatorInfo.Type) {
+    protected fun onSelected(type: AuthenticatorInfo.Type) {
         selectedAuthenticator = type
```
rmeissner: Would it be better to make `selectedAuthenticator` nullable and have it as `null` by default (since it can be "nothing" now)

app/src/main/res/drawable/ic_telegram.xml
```diff
@@ -0,0 +1,22 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="28dp"
```
rmeissner: why 28dp? isn't the default 24dp (or something else that can be devided by 8) ... I would strongly prefer if we stick to the grid of 8 (or at least 4)

app/src/main/res/values/colors.xml
```diff
@@ -10,7 +10,7 @@
     <color name="black">#000000</color>
     <color name="white_alpha_70">#b3ffffff</color>
     <color name="black_alpha_70">#b3000000</color>
-    <color name="disabled_button">#9b9b9b</color>
+    <color name="disabled_button">#80008c73</color>
```
rmeissner: This changes the behavior in the whole app for quite some different buttons, is this intended

app/src/main/java/pm/gnosis/heimdall/ui/safe/recover/safe/RecoverSafe2FAActivity.kt
elgatovital: something that has to be changed in the future :) I'll change it to normal comment

app/src/main/res/drawable/ic_telegram.xml
elgatovital: those are the values from our mocks. I will use 24dp

app/src/main/res/values/colors.xml
elgatovital: according to mocks disabled buttons should be greenish with some transparency. This was intended.

app/src/main/java/pm/gnosis/heimdall/ui/two_factor/Select2FaActivity.kt
elgatovital: sure. It also makes sense to select nothing (leave radio button unchecked) if NFC is not available.

app/src/main/java/pm/gnosis/heimdall/ui/two_factor/Select2FaActivity.kt
elgatovital: this changes alpha of the text span. "Learn more" remains always opaque, but description changes transparency if an option is not available. I would extract alpha value to res/values

app/src/main/java/pm/gnosis/heimdall/ui/settings/general/GeneralSettingsActivity.kt
elgatovital: we could. I am not sure whether we still need to add a list of installed safes to the email body. If this is not needed anymore, then we can delete it.

rmeissner: it is an unused method, right? so I would remove it

app/src/main/java/pm/gnosis/heimdall/ui/two_factor/Select2FaActivity.kt
rmeissner: But they are just 2 different colors, just put them both in the resource file, right?

app/src/main/res/values/colors.xml
rmeissner: mocks = zeplin designs?

elgatovital: yes

app/src/main/java/pm/gnosis/heimdall/ui/settings/general/GetInTouchActivity.kt
elgatovital: it can, but android will show app chooser for the first time (choosing between browser and telegram)

PR #554, opened 2020-04-08
Final setup

jpalvarezl:
Final setup for optimal parallel working.


app/src/main/java/io/gnosis/safe/ui/base/BaseStateViewModel.kt
```diff
@@ -45,14 +32,12 @@ abstract class BaseStateViewModel<T : BaseStateViewModel.State>(
...
+            takeUnless { !forceViewAction && nextState === currentState }?.run {
+                stateChannel.send(nextState)
+            }
         } catch (e: Exception) {
-            // Could not submit update
             Timber.e(e)
         }
     }
 
-    protected fun safeLaunch(errorHandler: CoroutineExceptionHandler = coroutineErrorHandler, block: suspend CoroutineScope.() -> Unit) =
```
rmeissner, line 56: If you remove this then you can also remove `coroutineErrorHandler` ... I would strongly recomment to have something like this though as it makes error handling a log easier (but also depends what the plan for error handling is ;) )

app/src/main/java/io/gnosis/safe/ui/safe/overview/SafeOverviewFragment.kt
```diff
@@ -0,0 +1,23 @@
...
+import io.gnosis.safe.di.components.DaggerViewComponent
+import io.gnosis.safe.di.modules.ViewModule
+import io.gnosis.safe.ui.base.BaseFragment
+
+class SafeOverviewFragment : BaseFragment<FragmentSafeOverviewBinding>() {
+
+    override fun inject(component: ApplicationComponent) {
+        DaggerViewComponent.builder()
+            .applicationComponent(component)
+            .viewModule(ViewModule(context!!))
```
rmeissner, line 16: basically all fragments will use the same viewmodule so you can actually build the `DaggerViewComponent` (or at least the builder) in the `BaseFragment` to avoid that we have to do this everywhere again

app/src/main/java/io/gnosis/safe/ui/base/BaseStateViewModel.kt
jpalvarezl: I saw your comments after merging, sorry for that. I will restore this.

Issue #557, opened 2020-04-14
Safe persistency layer


elgatovital:
- Use room for persistency
   see https://github.com/gnosis/safe-android/issues/549 for db schema suggestion
- Use shared preferences for active safe
- Define DB, DAO'
- Add migration tests
- Add basic safe repository 
  - validate safe info
  - provide list of safes
  - provide currently selected safe
 
Ideally we can have this merged and published to not have to iterate for clean up for this PR: https://github.com/gnosis/svalinn-kotlin/pull/43

jpalvarezl:
@elgatovital I need some clarification, regarding the migration tests. It seems like in order to have a migration we need to have modified the database schema somehow and as this is the first iteration, there is really no way to do a migration as far as I understand. Am I missing something?

PR #561, opened 2020-04-16
Safe persistance

jpalvarezl:
Closes https://github.com/gnosis/safe-android/issues/557 .

- Added SafeDao and room implementation for the app
- Added basic use cases for the SafeRepository

@gnosis/mobile-devs


data/build.gradle
```diff
@@ -1,6 +1,45 @@
 apply from: '../buildsystem/android_library.gradle'
 apply from: '../buildsystem/test_dependencies.gradle'
+apply from: '../buildsystem/test_android_dependencies.gradle'
+apply plugin: 'kotlin-kapt'
+
+android {
+    defaultConfig {
```
elgatovital, line 7: maybe :contracts is more suitable for these constants, as they relate to contracts

jpalvarezl: I tried but the `:contracts` module is a "java-library", I would have to change it to an android one. Are you ok with that?
Also, if we need to access this constants from :app, the :data module will always have to act as mediator between :app and :contracts. If you are fine with that I will push the change.

data/src/main/java/io/gnosis/data/repositories/SafeRepository.kt
```diff
@@ -0,0 +1,70 @@
...
+            }
+
+    private suspend fun getSafeBy(address: Solidity.Address): Safe? = safeDao.loadByAddress(address)
+
+    companion object {
+
+        @VisibleForTesting
+        const val ACTIVE_SAFE = "prefs.string.active_safe"
+
+        val safeMasterCopy_0_0_2 = BuildConfig.SAFE_MASTER_COPY_0_0_2.asEthereumAddress()!!
```
elgatovital, line 55: constants should be upercase, what do you think?

data/src/test/java/io/gnosis/data/repositories/SafeRepositoryTest.kt
```diff
@@ -0,0 +1,207 @@
...
+        preferences = spyk()
+        val application = mockk<Application>().apply {
+            every { getSharedPreferences(any(), any()) } returns preferences
+        }
+        val preferencesManager = PreferencesManager(application)
+        safeRepository = SafeRepository(safeDao, preferencesManager, ethereumRepository)
+    }
+
+    @Test
+    fun `getSafes - should return safes in the database`() = runBlocking {
```
elgatovital, line 37: should we use test names without spaces here?

data/build.gradle
elgatovital: I'm fine with the way it is right now. Having those constants directly in :app would also be good

data/src/main/java/io/gnosis/data/repositories/SafeRepository.kt
jpalvarezl: Sure, let's make them all SCREAMING_SNAKE_CASING?

data/src/test/java/io/gnosis/data/repositories/SafeRepositoryTest.kt
jpalvarezl: you mean like in instrumentation tests?

elgatovital: I was thinking about having the same naming schema for all tests, but having spaces improves readability a lot. Let's use the same structure but spaces for unit tests and underscores for instrumentation.

PR #563, opened 2020-04-20
Removed usage of VisibleForTesting annotation

jpalvarezl:
- Remove visible for testing
- shared prefs const are private

@gnosis/mobile-devs


rmeissner:
For context, this change was done because of some internal comment I made to @jpalvarezl 

```
Exposing state for tests seems wrong, as they are normally not only exposed in tests but also in other places (if I get it correctly the annotation will prevent this, similar to @hide, right?). Also the pref key strings should probably be hardcoded in the tests so in case we by accident rename the string the test should fail (since this is quite critical that we do not just change pref strings)
```

PR #565, opened 2020-04-21
Add basic style definitions

elgatovital:
Add basic styles for buttons & text

@gnosis/mobile-devs


app/src/main/res/values/styles.xml
```diff
@@ -17,11 +17,107 @@
...
+        <item name="android:textStyle">normal</item>
+    </style>
+
+    <style name="TextLight" parent="Text">
+        <item name="fontFamily">sans-serif</item>
+        <item name="android:lineSpacingExtra">6sp</item>
+        <item name="android:textColor">@color/dark_grey</item>
+        <item name="android:textSize">14sp</item>
+        <item name="android:textStyle">normal</item>
+    </style>
```
jpalvarezl, line 72: I think I will probably add additionally a `TextButton` style in my PR that inherits from TextDark problably. I guess so that we can add the ripple effect on click too.

app/src/main/res/values/dimens.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <dimen name="toolbar_shadow_size">1dp</dimen>
```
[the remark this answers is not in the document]
elgatovital: it's artificial shadow height. You can have elevation for the component itself. But if used only in view that acts as a shadow, size is a better name I think

app/src/main/res/values/styles.xml
elgatovital: Can't you use primary button style?

app/src/main/res/values/styles.xml
```diff
@@ -24,4 +24,77 @@
...
+
+    <style name="Toolbar">
+        <item name="theme">?attr/actionBarTheme</item>
+        <item name="background">@color/white</item>
+        <item name="android:layout_width">match_parent</item>
+        <item name="android:layout_height">?attr/actionBarSize</item>
+    </style>
```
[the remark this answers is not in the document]
elgatovital: no, we don't have actual toolbar. Just a container with artificial shadow

app/src/main/res/values/styles.xml
jpalvarezl: I suppose I could but I am concerned about the background being set. Some situations (like the "back" or "next" button on the toolbar) may need a transparent background.

PR #575, opened 2020-04-23
Safe address input

### LINKED ISSUE #559: Safe input
- Add screens for enterning new safe
  - enter address: https://zpl.io/adJk7oe
  - enter name https://zpl.io/amZ4nzK

Assumption: when back button is clicked whole flow is canceled and safe input needs to be started from the beginning.

- validate entered safe address
- Input methods: https://zpl.io/VD0nzY3
  - separate module for qr scanner

jpalvarezl:
Closes #559

- Added ENS, clipboard, QR scanning for safe address addition
- Empty address/local name validation as well as GetStorageAt for address
- ~~Unit tests missing for EnsRepo and viewModels.~~ 

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/di/TypeAdapters.kt
```diff
@@ -0,0 +1,66 @@
...
+import io.gnosis.safe.utils.parseToBigInteger
+import pm.gnosis.crypto.utils.asEthereumAddressChecksumString
+import pm.gnosis.model.Solidity
+import pm.gnosis.models.Wei
+import pm.gnosis.utils.asEthereumAddress
+import pm.gnosis.utils.hexAsBigInteger
+import pm.gnosis.utils.toHexString
+import java.math.BigInteger
+
+class WeiAdapter {
```
rmeissner, line 15: Should we use the ones in svalinn that @biafra23 added

app/src/main/java/io/gnosis/safe/helpers/AddressHelper.kt
```diff
@@ -0,0 +1,35 @@
...
+//TODO legacy handles name too
+class AddressHelper
+@Inject constructor() {
+    fun populateAddressInfo(
+        addressView: TextView,
+        imageView: BlockiesImageView?,
+        address: Solidity.Address
+    ) {
+
+        GlobalScope.launch(Dispatchers.Default) {
```
rmeissner, line 23: I think it is discouraged to use GlobalScope (need to check this again)

app/src/main/java/io/gnosis/safe/helpers/AddressTooltip.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    init {
+        setBackgroundDrawable(ColorDrawable(Color.TRANSPARENT))
+        isOutsideTouchable = true
+        isFocusable = false
+
+        contentView = LayoutInflater.from(context).inflate(R.layout.popup_address_tooltip, null)
+
+        //make first & last 4 characters bold
+        val addressString=  SpannableStringBuilder(address.substring(0, address.length / 2) + "\n" + address.substring(address.length / 2, address.length))
```
rmeissner, line 29: Should this be a utils (we will use this in many places, right?)

app/src/main/java/io/gnosis/safe/ui/base/BaseDialog.kt
```diff
@@ -0,0 +1,5 @@
+package io.gnosis.safe.ui.base
+
+import androidx.fragment.app.DialogFragment
+
+abstract class BaseDialog : DialogFragment()
```
rmeissner, line 5: Are we planning to add.stuff here?

app/src/main/java/io/gnosis/safe/utils/AppUtils.kt
```diff
@@ -0,0 +1,40 @@
...
+            onQrCodeResult(data.getStringExtra(QRCodeScanActivity.RESULT_EXTRA))
+        } else if (resultCode == Activity.RESULT_CANCELED) {
+            onCancelledResult?.invoke()
+        }
+        return true
+    }
+    return false
+}
+
+fun handleAddressBookResult(
```
rmeissner, line 23: Should we remove this, not sure when address book is planned, but might be a little bit in the future

app/src/main/java/io/gnosis/safe/utils/StringUtils.kt
```diff
@@ -0,0 +1,56 @@
...
+}
+
+fun String.underline(): CharSequence =
+    SpannableString(this).also {
+        it.setSpan(UnderlineSpan(), 0, length, 0)
+    }
+
+fun String.parseToBigIntegerOrNull(): BigInteger? =
+    nullOnThrow { parseToBigInteger() }
+
```
rmeissner, line 30: Should we use the svalinn libs?

app/src/main/java/io/gnosis/safe/di/Repositories.kt
```diff
@@ -0,0 +1,23 @@
...
+import dagger.Lazy
+import io.gnosis.data.repositories.EnsRepository
+import io.gnosis.data.repositories.SafeRepository
+import pm.gnosis.ethereum.EthereumRepository
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class Repositories
+@Inject constructor(
```
rmeissner, line 12: Just out of curiosity why do we need this intermediate wrapper?

app/src/main/java/io/gnosis/safe/di/TypeAdapters.kt
biafra23: We would need a svalinn release first. I would postpone this. I can go over the app and replace its usage as soon as there is a release.

jpalvarezl: But the PR for this change was already merged, wasn't it? If you want we can do a release for svalinn tomorrow and then I can replace this class with your changes.

app/src/main/java/io/gnosis/safe/helpers/AddressHelper.kt
jpalvarezl: It is, this was copied from Kouban. I can have a closer look.

app/src/main/java/io/gnosis/safe/helpers/AddressTooltip.kt
jpalvarezl: As a matter of fact we do have a utils for this 🤦 I will replace with it.

app/src/main/java/io/gnosis/safe/ui/base/BaseDialog.kt
jpalvarezl: I think Vitaly has something here in his branch. Tomorrow we will merge stuff and sync. I will probably end up getting rid of this anyhow.

app/src/main/java/io/gnosis/safe/utils/AppUtils.kt
jpalvarezl: I left all the places where the logic for the address book is implemented commented out, so maybe it doesn't hurt to have this ready. I think the `AddressInputHelper` is using this method anyway, even though it will never get called in the current implementation.

app/src/main/java/io/gnosis/safe/utils/StringUtils.kt
jpalvarezl: I will check which module has this utility method and replace with the svalinn one.

app/src/main/java/io/gnosis/safe/di/Repositories.kt
jpalvarezl: Well, "need" is a strong word. All the dependencies are injected using `dagger.Lazy<T>` which is nice, but our repos at least so far have 0 internal state so there is no performance gain there. Maybe eventually there will be. Also, this way we only need one method in our `ApplicationComponent` which reduces clutter. As, unlike for the legacy app, we don't use contructor injection for repos, since we have a separate module and we don't need dagger in that one just for that. Here we have a `RepositoryModule` and only the `Repositories` class gets provided by the `ApplicationComponent`.

rmeissner: Afaik dagger injection is always lazy (even if you do not use the lazy step in between). So classes injected via dagger will only be isntantiated when they are injected. Regarding the methods in the ApplicationComponent, it is just pushing them somewhere else, right? in most of the the repos so far you only use one of these repositories in that class.

Another point (would need to look into to be hones), but if you are not exposing the injected class outside of dependency graph you do not need the method in the ApplicationComponent. So for example in the legacy app we had that view model factory and that was a singleton in the same graph as the repositories and it was responsible for the view models. So actually it was not required to expose the repositories in the application component via methods (for reference https://github.com/gnosis/safe-android/blob/safe_v1/app/src/main/java/pm/gnosis/heimdall/di/components/ApplicationComponent.kt) 

Note on the side: the only reason I requested changes to the PR were the missing unit tests, the rest are more comments or out of pure interest (forgot to mention that 😬 )

app/src/main/java/io/gnosis/safe/utils/StringUtils.kt
rmeissner: I think it was/is added in @biafra23 svalinn PR ... we should merge that tomorrow and maybe do a release

qrscanner/src/main/java/io/gnosis/safe/qrscanner/QRCodeDecoder.kt
```diff
@@ -0,0 +1,33 @@
...
+        var size = width.coerceAtMost(height)
+        size = (size.toDouble() * ReticleView.FRAME_SCALE).toInt()
+
+        val halfSize = size / 2
+
+        val left = centerX - halfSize
+        val top = centerY - halfSize
+
+        val source = PlanarYUVLuminanceSource(data, width, height, left, top, size, size, false)
+        return nullOnThrow { source.decode() } ?: source.invert().decode()
```
rmeissner, line 31: Maybe we could a comment to this line. So we catch it once and invert the image, because some websites use white QR codes on black background which is not supported by the default `.decode`. This was the workaround that worked the best for me. If anyone wants to thest this they can use uniswap in dark mode (there they use an inverted QR code)

qrscanner/src/main/java/io/gnosis/safe/qrscanner/QRCodeScanActivity.kt
```diff
@@ -0,0 +1,91 @@
...
+    override fun onPause() {
+        super.onPause()
+        if (videographer.isOpen) videographer.close()
+    }
+
+    override fun onRequestPermissionsResult(requestCode: Int, permissions: Array<out String>, grantResults: IntArray) {
+        handlePermissionsResultRequest(requestCode, CAMERA_REQUEST_CODE, grantResults, onPermissionDenied = { finish() })
+    }
+
+    // TODO: We can use Rx to listen for successful events and finish instead of passing a callback
```
rmeissner, line 49: Rx here is probably not on our roadmap anymore ;)

app/src/main/java/io/gnosis/safe/di/Repositories.kt
jpalvarezl: Ah yes, but that's because you could annotate the constructors of the repositories with @Inject , I think. Like in there: https://github.com/gnosis/safe-android/blob/safe_v1/app/src/main/java/pm/gnosis/heimdall/data/repositories/impls/DefaultEnsRepository.kt

In our case, since the repository classes are all in the :data module, and we are not really pulling dagger into it (as we may want to extract at some point) we just have provider methods in the `RepositoryModule` defined here: https://github.com/gnosis/safe-android/blob/feature/address_input/app/src/main/java/io/gnosis/safe/di/modules/RepositoryModule.kt

I guess in a scenario where we do extract this into a dependency pulled from jitpack, then you could just have a provider method in your app component of the specific repos you want to pull in.

And yeah no worries about requesting changes, this by no means in a mergeable state :) I should have requested review once it was properly finished. But there is a lot of feedback I can already work on anyhow. Thanks!

rmeissner: That should be unrelated from the annotations, as these annotations just autogenerate code that has to be handwritten here. It depends highly on where the view models are created. I cannot really check in detail right now, so hard to tell ;)

rmeissner: As far as I can tell it should also work with the current setup, as you keep using the ViewModelFactory 🤔

app/src/main/java/io/gnosis/safe/di/TypeAdapters.kt
biafra23: I created a release. The version is v0.12.0

app/src/main/java/io/gnosis/safe/di/Repositories.kt
jpalvarezl: I think I will create a task for the next sprint to revise the dagger setup. I think we've mixed things the wrong in some places.

app/src/main/java/io/gnosis/safe/utils/ViewUtils.kt
```diff
@@ -0,0 +1,32 @@
...
+
+enum class DrawablePosition(val index: Int) {
+    DRAWABLE_LEFT(0),
+    DRAWABLE_TOP(1),
+    DRAWABLE_RIGHT(2),
+    DRAWABLE_BOTTOM(3)
+}
+
+@SuppressLint("ClickableViewAccessibility")
+inline fun TextView.setOnCompoundDrawableClicked(
```
rmeissner, line 18: Is this used?

jpalvarezl: Not anymore. In the address inputs the clicks were confined the the compound drawable but now I made the entire view clickable as it is no longer editable by hand anyhow. I will remove this as well in the cleanup tasks that I mentioned earlier.

app/src/main/java/io/gnosis/safe/di/Repositories.kt
jpalvarezl: Just for tying up the reference: https://github.com/gnosis/safe-android/issues/576

app/src/test/java/io/gnosis/safe/ui/safe/add/AddSafeNameViewModelTest.kt
```diff
@@ -0,0 +1,119 @@
...
+    private val repositories = mockk<Repositories>().apply {
+        every { safeRepository() } returns safeRepository
+    }
+
+    private lateinit var viewModel: AddSafeNameViewModel
+
+    @Before
+    fun setup() {
+        viewModel = AddSafeNameViewModel(repositories, appDispatchers)
+        Dispatchers.setMain(TestCoroutineDispatcher())
```
rmeissner, line 37: Would it make sense to have a base class for the dispatchers setup?

jpalvarezl: Yeah, I think so 👍

PR #579, opened 2020-04-29
Closes #550: internal release

### LINKED ISSUE #550: Internal release
https://github.com/gnosis/safe/issues/300

biafra23:
Closes #550 .

- Added Firebase distribution gradle plugin

**Attention**! Build server needs new google-services.json and a new FIREBASE_TOKEN as an environment variable.

You also need a copy of the `google-services.json` file. You can find the newest version here: https://console.firebase.google.com/u/0/project/safe-firebase-staging/settings/general/android:io.gnosis.safe.internal

@gnosis/mobile-devs


app/build.gradle
```diff
@@ -3,11 +3,12 @@ apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-android-extensions'
 apply plugin: 'kotlin-kapt'
 apply plugin: 'androidx.navigation.safeargs.kotlin'
+apply plugin: 'com.google.gms.google-services'
```
elgatovital, line 6: shouldn't this plugin be applied as last at the very bottom?

jpalvarezl: I was under the same impression but it seems like it might no longer be necessary: https://firebase.google.com/docs/android/setup#console

jpalvarezl:
> You also need a copy of the `google-services.json` file. You can find the newest version here: https://console.firebase.google.com/u/0/project/safe-firebase-staging/settings/general/android:io.gnosis.safe.internal

This should probably be added to the repo README file.

biafra23:
> > You also need a copy of the `google-services.json` file. You can find the newest version here: https://console.firebase.google.com/u/0/project/safe-firebase-staging/settings/general/android:io.gnosis.safe.internal
> 
> This should probably be added to the repo README file.

The link is only accessible for Gnosis accounts. External users need to generate their own firebase project and `google-services.json`

jpalvarezl:
> > > You also need a copy of the `google-services.json` file. You can find the newest version here: https://console.firebase.google.com/u/0/project/safe-firebase-staging/settings/general/android:io.gnosis.safe.internal
> > 
> > 
> > This should probably be added to the repo README file.
> 
> The link is only accessible for Gnosis accounts. External users need to generate their own firebase project and `google-services.json`

Correct, but it is also useful info for future employees. Otherwise, we have to go comment diving in github. Well, is not terribly important, but then we don't have to really on our memory in the long term.

PR #580, opened 2020-04-30
Basic safe overview screens

### LINKED ISSUE #549: Add Safes, remove Safes and switching between them
https://github.com/gnosis/safe/issues/285

### LINKED ISSUE #558: Safe selection screen setup
- No safe selected (initial state)
   - Toolbar without data: https://zpl.io/aMgdlBz
   - Empty content area for now
- Safe selected
  - Updated toolbar:  https://zpl.io/aB0NNwL
    - ugly down arrow in toolbar opens list of safe
  - Load btn
  - label
- List of safes: https://zpl.io/aMgvqjz
- Define styles

### LINKED ISSUE #560: Remove safe
Specify ui details
- remove action in toolbar?
- remove using swipe to delete in list of safes?
- long press + edit mode in list of safes?

elgatovital:
Closes #549 
Closes #558 
Closes #560 

- navigation for safe overview screens
- basic fragments, layouts, and styles
- add / remove / switch safe

@gnosis/mobile-devs


app/build.gradle
```diff
@@ -162,9 +162,11 @@ dependencies {
     implementation project(":data")
     implementation project(":qrscanner")
 
+    implementation "com.google.code.findbugs:jsr305:$versions.findbugs"
```
rmeissner, line 165: Why do we need this? It would be nice to have a little bit more info on that as I don't see any annotation in this PR that uses this. Also I assume we would need findbugs for that (which is deprecated).

app/src/main/java/io/gnosis/safe/ui/safe/NoSafeFragment.kt
```diff
@@ -1,28 +1,29 @@
...
-        FragmentSafeOverviewBinding.inflate(inflater, container, false)
+    override fun inflateBinding(inflater: LayoutInflater, container: ViewGroup?): FragmentNoSafesBinding =
+        FragmentNoSafesBinding.inflate(inflater, container, false)
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
-        binding.addSafeButton.setOnClickListener {
-            findNavController().navigate(SafeOverviewFragmentDirections.actionSafeOverviewFragmentToAddSafeNav())
+        binding.loadSafeButton.setOnClickListener {
+            Navigation.findNavController(activity as FragmentActivity, R.id.safe_overview_root).navigate(SafeOverviewFragmentDirections.actionSafeOverviewFragmentToAddSafeNav())
```
jpalvarezl, line 26: Just calling `findNavController()` doesn't work?

app/src/main/java/io/gnosis/safe/ui/safe/SafeSelectionDialog.kt
```diff
@@ -0,0 +1,277 @@
...
+    lateinit var viewModel: SafeSelectionViewModel
+
+    @Inject
+    lateinit var adapter: SafeSelectionAdapter
+
+    override fun inject() {
+        DaggerViewComponent.builder()
+            .viewModule(ViewModule(context!!))
+            .applicationComponent(HeimdallApplication[context!!])
+            .build()
```
jpalvarezl, line 48: Building the view component can be pushed to the base class like it is for fragments.

app/src/main/java/io/gnosis/safe/ui/safe/SafeSelectionDialog.kt
```diff
@@ -0,0 +1,277 @@
...
+        override var viewAction: BaseStateViewModel.ViewAction?
+    ) : SafeSelectionState()
+
+    data class AddSafeState(
+        override var viewAction: BaseStateViewModel.ViewAction?
+    ) : SafeSelectionState()
+}
+
+
+class SafeSelectionViewModel @Inject constructor(
```
jpalvarezl, line 110: I think we can put this in a separate source file like we usually do with viewModels. The viewModel and the states emitted by it.

app/src/main/java/io/gnosis/safe/ui/safe/SafeSelectionDialog.kt
```diff
@@ -0,0 +1,277 @@
...
+    var activeSafe: Safe? = null
+        set(value) {
+            field = value
+            notifyAllChanged()
+        }
+
+    fun setItems(items: List<Any>, activeSafe: Safe?) {
+        this.activeSafe = activeSafe
+        this.items.clear()
+        this.items.addAll(items)
```
jpalvarezl, line 185: This method sets active safe as a side effect. Would it make more sense to decouple that logic?

app/src/main/java/io/gnosis/safe/ui/safe/SafeSelectionDialog.kt
```diff
@@ -0,0 +1,277 @@
...
+            )
+        }
+    }
+
+    override fun getItemViewType(position: Int): Int {
+        val item = items[position]
+        return when (item) {
+            is AddSafeHeader -> SafeSelectionViewTypes.HEADER_ADD_SAFE.ordinal
+            is Safe -> SafeSelectionViewTypes.SAFE.ordinal
+            else -> throw UnsupportedViewType(item.toString())
```
jpalvarezl, line 222: Even though we are no using the super generic solution, defining a ViewHolder factory would make your adapter easier to digest.

jpalvarezl: Same goes for `OnCreateViewHolder`

app/src/main/java/io/gnosis/safe/ui/safe/settings/SafeSettingsFragment.kt
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    override fun onResume() {
+        super.onResume()
+        viewModel.loadSafe()
+    }
+}
+
+
+class SafeSettingsViewModel @Inject constructor(
```
jpalvarezl, line 70: Separate class for this too would be nice.

app/src/main/java/io/gnosis/safe/ui/safe/settings/SafeSettingsFragment.kt
```diff
@@ -0,0 +1,112 @@
...
+    repositories: Repositories,
+    appDispatchers: AppDispatchers
+) : BaseStateViewModel<SafeSettingsState>(appDispatchers) {
+
+    private val safeRepository = repositories.safeRepository()
+
+    private lateinit var safe: Safe
```
jpalvarezl, line 77: This is unused as far as I can see

app/src/main/java/io/gnosis/safe/ui/safe/settings/SafeSettingsFragment.kt
```diff
@@ -0,0 +1,112 @@
...
+            emit(event)
+    }
+
+    override fun initialState(): SafeSettingsState = SafeSettingsState.SafeLoading(null)
+
+    fun loadSafe() {
+
+        safeLaunch {
+
+            safe = safeRepository.getActiveSafe()!!
```
jpalvarezl, line 90: Do you need this bang operator? If there is no safe selected shouldn't we have a an empty state ready for that situation in settings?

app/src/main/java/io/gnosis/safe/ui/safe/NoSafeFragment.kt
elgatovital: no, it will nav graph for current nav host. But we need main/global nav graph here

app/src/main/java/io/gnosis/safe/ui/safe/SafeSelectionDialog.kt
elgatovital: I think short view models that are used only in a fragment can coexist in a same file with fragment. Then everything related is in one file and you don't have to jump around. But it doesn't matter to me. We have to agree on one approach and stick to it

app/src/main/java/io/gnosis/safe/ui/safe/SafeSelectionDialog.kt
elgatovital: I prefer not to do it in this pr

app/src/main/java/io/gnosis/safe/ui/safe/settings/SafeSettingsFragment.kt
elgatovital: safe repository give nullable object back. But inside safe settings we can be sure, that there is a safe available. Otherwise, navigation here wouldn't be possible.

app/src/main/java/io/gnosis/safe/ui/safe/SafeSelectionDialog.kt
jpalvarezl: Well this file is nearly 300 lines long if you don't want to split up the viewModel then something else.

app/src/main/java/io/gnosis/safe/utils/NavUtils.kt
```diff
@@ -0,0 +1,14 @@
+package io.gnosis.safe.utils
+
+import androidx.fragment.app.FragmentManager
+import androidx.navigation.NavDirections
+import androidx.navigation.fragment.NavHostFragment
+
+fun navigateFromChild(childFragmentManager: FragmentManager, directions: NavDirections) {
+    try {
+        (childFragmentManager.primaryNavigationFragment as NavHostFragment).navController.navigate(directions)
```
jpalvarezl, line 9: Maybe write this as an extension `fun FragmentManager.getChildNavController(): NavController?`. Also it's probably better to use a safe cast `as?` and if you still need the try and catch please put the exception to `Timber`

app/src/main/java/io/gnosis/safe/ui/safe/SafeOverviewViewModel.kt
```diff
@@ -0,0 +1,50 @@
...
+import javax.inject.Inject
+
+class SafeOverviewViewModel @Inject constructor(
+    repositories: Repositories,
+    appDispatchers: AppDispatchers
+) : BaseStateViewModel<SafeOverviewState>(appDispatchers) {
+
+    private val safeRepository = repositories.safeRepository()
+
+    private var activeSafe: Safe? = null
```
jpalvarezl, line 18: This could lead to inconsistent states.

app/src/main/java/io/gnosis/safe/ui/safe/SafeOverviewViewModel.kt
```diff
@@ -0,0 +1,50 @@
...
+    fun loadSafe() {
+        safeLaunch {
+            activeSafe = safeRepository.getActiveSafe()
+            updateState { SafeOverviewState.ActiveSafe(activeSafe, null) }
+        }
+    }
+
+    fun removeSafe() {
+        safeLaunch {
+            activeSafe?.let {
```
jpalvarezl, line 36: I think is better to get rid of the mutable internal state and use here `safeRepository.getActiveSafe()` as the single source of truth for the currently active safe.

app/src/main/java/io/gnosis/safe/ui/safe/SafeOverviewFragment.kt
```diff
@@ -0,0 +1,95 @@
...
+        FragmentSafeOverviewBinding.inflate(inflater, container, false)
+
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+
+        with(findNavController(activity!!, R.id.safe_overview_content)) {
+            binding.bottomNavigation.setupWithNavController(this)
+            addOnDestinationChangedListener { _, destination, _ ->
```
jpalvarezl, line 40: Empty callback for listener

app/src/main/java/io/gnosis/safe/ui/safe/SafeSelectionDialog.kt
jpalvarezl: What I suggest is to do the following, and using the package structure as guidance:

ui.safe.
            add.
                  //...
            select.
                 Dialog
                 Adapter
                 ViewModel
           overview
                 fragment
                 viewModel
           Balances. //this doesn't need to be on your PR
                .nfts
                .fts

app/src/main/java/io/gnosis/safe/ui/base/BaseBottomSheetDialogFragment.kt
```diff
@@ -0,0 +1,49 @@
...
+import androidx.viewbinding.ViewBinding
+import com.google.android.material.bottomsheet.BottomSheetDialog
+import com.google.android.material.bottomsheet.BottomSheetDialogFragment
+import io.gnosis.safe.HeimdallApplication
+import io.gnosis.safe.R
+import io.gnosis.safe.di.components.ApplicationComponent
+
+abstract class BaseBottomSheetDialogFragment<T : ViewBinding> : BottomSheetDialogFragment() {
+
+    private var _binding: T? = null
```
rmeissner, line 17: why do we start a variable name with `_` isn't that super uncommon for our style guidelines? (shouldn't we use something like `internalBinding` or try to refactor it so that we don't need this dirty hack)

elgatovital: It's common in kotlin to use underscore for backing properties. See https://kotlinlang.org/docs/reference/coding-conventions.html

app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
```diff
@@ -1,13 +1,87 @@
...
+import pm.gnosis.utils.asEthereumAddressString
 
-class StartActivity : BaseActivity() {
+class StartActivity : BaseActivity(), SafeOverviewNavigationHandler {
+
+    private val toolbarBinding by lazy {
+        ToolbarSafeOverviewBinding.bind(findViewById(R.id.toolbar_container))
+    }
+    private val toolbar by lazy { findViewById<View>(R.id.toolbar) }
+    private val navBar by lazy { findViewById<BottomNavigationView>(R.id.nav_bar) }
```
jpalvarezl, line 24: I wonder if you can access these 2 from the `toolbarBinding`

app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
```diff
@@ -1,13 +1,87 @@
...
+        setupNav()
+    }
+
+    private fun setupNav() {
+        val navController = Navigation.findNavController(this, R.id.nav_host)
+        navBar.setupWithNavController(navController)
+        navController.addOnDestinationChangedListener { _, destination, _ ->
+            if (isFullscreen(destination.id)) {
+                toolbar.visibility = View.GONE
+                navBar.visibility = View.GONE
```
jpalvarezl, line 41: Did you decide to not use motion layout? I suppose that here you would have `transitionToEnd` or `start` depending on how you defined it.

app/src/main/java/io/gnosis/safe/ui/base/BaseStateViewModel.kt
```diff
@@ -15,9 +20,11 @@ data class AppDispatchers(
 )
 
 abstract class BaseStateViewModel<T>(private val dispatchers: AppDispatchers) : ViewModel() where T : BaseStateViewModel.State {
-    abstract val state: LiveData<T>
 
-//    private val errorHandler = SimpleLocalizedException.networkErrorHandlerBuilder(context).build()
+    val state: LiveData<T> = liveData {
+        for (event in stateChannel.openSubscription())
```
jpalvarezl, line 25: We could still provide a callback here. As in the legacy app in some places behaviour is injected here. Like an `open fun onStateSubscribed()` with an empty default implementation to not have to do it everywhere.

app/src/main/java/io/gnosis/safe/ui/safe/balances/SafeBalancesViewModel.kt
```diff
@@ -0,0 +1,46 @@
...
+    override fun initialState(): SafeBalancesState = SafeBalancesState.SafeLoading(null)
+
+    init {
+        safeLaunch {
+            safeRepository.activeSafeFlow().collect { safe ->
+                updateState {
+                    SafeBalancesState.ActiveSafe(
+                        safe, if (safe == null)
+                            ViewAction.NavigateTo(
+                                SafeBalancesFragmentDirections.actionSafeBalancesFragmentToNoSafeFragment()
```
jpalvarezl, line 26: So for now I think it's fine, but I think it would be cool to have  viewHolder that acts as an empty state, so anywhere we have list we can use it.

app/src/main/java/io/gnosis/safe/ui/safe/empty/NoSafeViewModel.kt
```diff
@@ -0,0 +1,40 @@
...
+    appDispatchers: AppDispatchers
+) : BaseStateViewModel<SafesState>(appDispatchers) {
+
+    private val safeRepository = repositories.safeRepository()
+
+    override fun initialState() = SafesState(null, 0, ViewAction.Loading(true))
+
+    init {
+        safeLaunch {
+            safeRepository.activeSafeFlow().collect { safe ->
```
jpalvarezl, line 21: Very nice

app/src/main/java/io/gnosis/safe/ui/safe/empty/NoSafeViewModel.kt
```diff
@@ -0,0 +1,40 @@
...
+                val safeNum = safeRepository.getSafes().count()
+                updateState {
+                    SafesState(
+                        safe, safeNum, if (safe == null) ViewAction.None else ViewAction.NavigateTo(
```
jpalvarezl, line 25: Not important now, but I think it's more readable to extract this into a `val viewAction` since `updateState { }` is a callback after all

app/src/main/java/io/gnosis/safe/ui/safe/selection/SafeSelectionDialog.kt
```diff
@@ -0,0 +1,97 @@
...
+
+class SafeSelectionDialog : BaseBottomSheetDialogFragment<DialogSafeSelectionBinding>() {
+
+    @Inject
+    lateinit var viewModel: SafeSelectionViewModel
+
+    @Inject
+    lateinit var adapter: SafeSelectionAdapter
+
+    var navHandler: SafeOverviewNavigationHandler? = null
```
jpalvarezl, line 31: Is this what you were referring as duplicate code? We could somehow refactor this into a helper class "ActiveSafeHelper" but, that's something we could do in a more tech debt oriented sprint.

app/src/main/java/io/gnosis/safe/ui/safe/selection/SafeSelectionViewModel.kt
```diff
@@ -0,0 +1,85 @@
...
+
+    fun addSafe() {
+        safeLaunch {
+            updateState {
+                SafeSelectionState.AddSafeState(
+                    ViewAction.NavigateTo(
+                        object : NavDirections {
+                            override fun getArguments() = Bundle()
+                            override fun getActionId() = R.id.action_to_add_safe_nav
+                        }
```
jpalvarezl, line 70: There is no class that matches this action generated by the safeargs dependency?

app/src/main/java/io/gnosis/safe/ui/safe/settings/SafeSettingsFragment.kt
```diff
@@ -0,0 +1,66 @@
...
+        component.inject(this)
+    }
+
+    override fun inflateBinding(inflater: LayoutInflater, container: ViewGroup?): FragmentSafeSettingsBinding =
+        FragmentSafeSettingsBinding.inflate(inflater, container, false)
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        binding.button.setOnClickListener {
+            viewModel.removeSafe()
```
jpalvarezl, line 32: Just a thought, we should work on debounced clicks with coroutines. This could lead to duplicates when we get deeper with DAOs.

app/src/main/java/io/gnosis/safe/ui/safe/settings/SafeSettingsViewModel.kt
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    fun removeSafe() {
+        safeLaunch {
+            val safe = safeRepository.getActiveSafe()!!
+            safeRepository.removeSafe(safe)
+            updateState {
+                SafeSettingsState.SafeRemoved(
+                    ViewAction.NavigateTo(
+                        SafeSettingsFragmentDirections.actionSafeSettingsFragmentToNoSafeFragment()
```
jpalvarezl, line 34: This is good default assumed behaviour. I hope it syncs with what product decides about it xD

app/src/main/java/io/gnosis/safe/utils/NavUtils.kt
```diff
@@ -0,0 +1,14 @@
+package io.gnosis.safe.utils
+
+import androidx.fragment.app.FragmentManager
+import androidx.navigation.NavDirections
+import androidx.navigation.fragment.NavHostFragment
+
+fun navigateFromChild(childFragmentManager: FragmentManager, directions: NavDirections) {
```
jpalvarezl, line 7: Do we need this anymore?

app/src/main/res/layout/toolbar_safe_overview.xml
```diff
@@ -0,0 +1,57 @@
...
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintTop_toTopOf="parent"
+        app:srcCompat="@drawable/ic_circle_dropdown_24dp" />
+
+</androidx.constraintlayout.widget.ConstraintLayout>
+
+
+
+
+
```
jpalvarezl, line 57: That's a lot of white space :P

app/src/test/java/io/gnosis/safe/MainCoroutineScopeRule.kt
```diff
@@ -0,0 +1,68 @@
...
+ * ```
+ * mainCoroutineRule.pauseDispatcher()
+ * // do some coroutines
+ * mainCoroutineRule.advanceUntilIdle() // run all pending coroutines until the dispatcher is idle
+ * ```
+ *
+ * By default, [MainCoroutineScopeRule] will be in a *resumed* state.
+ *
+ * @param dispatcher if provided, this [TestCoroutineDispatcher] will be used.
+ */
```
jpalvarezl, line 47: Nice documentation. I will replace the ones I've done with just `runBlocking` at some point.

buildsystem/versions.gradle
```diff
@@ -10,7 +10,7 @@ ext {
             // Runtime dependencies
             androidx_app_compat       : '1.1.0',
             androidx_card_view        : '1.0.0',
-            androidx_constraint_layout: '1.1.3',
+            androidx_constraint_layout: '2.0.0-beta4',
```
jpalvarezl, line 13: I would normally be against using anything that is not at least a release candidate, but i suppose we don't use a lot of the newer API that they are adding so it should be safe, right?

data/src/main/java/io/gnosis/data/repositories/SafeRepository.kt
```diff
@@ -19,6 +22,21 @@ class SafeRepository(
     private val ethereumRepository: EthereumRepository
 ) {
 
+    val keyFlow = callbackFlow {
+        val listener = SharedPreferences.OnSharedPreferenceChangeListener { _, key -> offer(key) }
+        preferenceManager.prefs.registerOnSharedPreferenceChangeListener(listener)
+        awaitClose { preferenceManager.prefs.unregisterOnSharedPreferenceChangeListener(listener) }
```
jpalvarezl, line 28: Nice :) My only concern is that this is a cold flow, and here it seems like a hot observable makes more sense. But it it works it works. I think we will find out when you add the tests to this.

data/src/main/java/io/gnosis/data/repositories/SafeRepository.kt
```diff
@@ -19,6 +22,21 @@ class SafeRepository(
...
+    }
+
+    fun activeSafeFlow() =
+        keyFlow
+            .filter { it == ACTIVE_SAFE }
+            .onStart { emit(ACTIVE_SAFE) }
+            .map { getActiveSafe() }
+            .conflate()
```
jpalvarezl, line 36: Could `keyFlow` be private? I would prefer it if it were.

buildsystem/versions.gradle
elgatovital: I have reverted to the older version, as we currently don't use motion layout

app/src/main/java/io/gnosis/safe/ui/safe/add/AddSafeNameViewModel.kt
```diff
@@ -34,12 +36,7 @@ class AddSafeNameViewModel
         }
     }
 
-    override val state: LiveData<State> = liveData {
```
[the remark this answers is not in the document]
elgatovital: sure, it still can be overridden or we can add hooks if necessary at some later point. Right now it has the same behavior in all of the view models. That's why I moved it to base

app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
elgatovital: toolbarBinding is toolbar container for accessing views, that display safe data. Navbar is bottom nav view, and toolbar is a container that has both toolbar container (toolbarBinding) and it's artificial shadow. I don't think it is possible to access outer elements from binding.

app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
elgatovital: yes, basically it is same logic, but without animated transitions. We could add motion scene at some later point.

app/src/main/java/io/gnosis/safe/ui/base/BaseStateViewModel.kt
```diff
@@ -22,10 +22,14 @@ data class AppDispatchers(
 abstract class BaseStateViewModel<T>(private val dispatchers: AppDispatchers) : ViewModel() where T : BaseStateViewModel.State {
 
     val state: LiveData<T> = liveData {
+        onStateSubscribed()
         for (event in stateChannel.openSubscription())
             emit(event)
     }
 
+    fun onStateSubscribed() {
```
jpalvarezl, line 30: Should be `open` 👍

elgatovital:
@rmeissner can I merge now and start adding tests? (https://github.com/gnosis/safe-android/issues/585)

jpalvarezl:
> @rmeissner can I merge now and start adding tests? (#585)

Ping in slack is more effective :D

PR #584, opened 2020-05-05
#551 Terms of Use and Privacy policy

### LINKED ISSUE #551: Accept terms and privacy policy
https://github.com/gnosis/safe/issues/292

biafra23:
Closes #551 .

- Add a screen showing Links to open Terms of Use and Privacy policy
- Only after agreeing to it, the user can continue to add a safe
- I added a style for `TextDark.Bold`
-  I changed the font-family for PrimaryButton from `sans-serif-medium` to `sans-serif` and added a textStyle of bold instead. This matches the Designs in Zeplin

Design: https://app.zeplin.io/project/5e9efc30726e106958cc43de/screen/5ea803ab22315c2586ffe275

Screenshot: 

@gnosis/mobile-devs


app/src/test/java/io/gnosis/safe/ui/safe/overview/SafeOverviewViewModelTest.kt
```diff
@@ -0,0 +1,62 @@
...
+        viewModel.termsBottomSheetDialog = mockk(relaxed = true)
+        val callback: () -> Unit = mockk(relaxed = true)
+
+        viewModel.checkTerms(callback)
+
+        verify { viewModel.termsBottomSheetDialog.show() }
+        verify(exactly = 0) { callback.invoke() }
+    }
+
+    // TODO: Test if advance is called when user clicks agree
```
rmeissner, line 58: Should we create an issue for these, or will you add them later to this PR?

biafra23: I was hoping for tips on how to test these in this PR and then I would add them immediately.

app/src/main/java/io/gnosis/safe/ui/safe/overview/TermsBottomSheetDialog.kt
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        bottom_sheet_terms_and_conditions_reject.setOnClickListener {
+            onReject()
+        }
+    }
+
+    private fun onAgree() {
+        dismiss()
+        onUserClicksAgree()
```
jpalvarezl, line 34: I would say that maybe we are better off decoupling the `terms` from the `SafeOverviewViewModel` and have a separate `TermsViewModel` so later on we can easily show these from any context. At some point it may be required that we show these when the SafeOverviewViewModel is not even instanced.

app/src/main/java/io/gnosis/safe/ui/safe/overview/SafeOverviewViewModel.kt
```diff
@@ -0,0 +1,39 @@
+package io.gnosis.safe.ui.safe.overview
+
+import androidx.lifecycle.ViewModel
+import pm.gnosis.svalinn.common.PreferencesManager
+import pm.gnosis.svalinn.common.utils.edit
+import javax.inject.Inject
+
+class SafeOverviewViewModel
```
jpalvarezl, line 8: I know is just naming, but I would rename this to `TermsViewModel` and put under the package `io.gnosis.safe.ui.terms` as these don't have to be coupled with the overview screen necessarily.

jpalvarezl: Of course along with the bottom sheet fragment. Similarly to what we have for the EnsInputDialog.

jpalvarezl:
Ah also, could you change the title of the PR, as it will become part of the commit message. I think is not necessary to mention the issue or whether if it is a fix or a feature. But a cleaner title would be nice, as we squash merge.

app/src/main/java/io/gnosis/safe/ui/safe/overview/SafeOverviewViewModel.kt
biafra23: As soon as PR#580 is merged, I will have to change this.

biafra23:
Since this should be triggered from the launch screen. Should I just create the launch screen as part of this PR? https://app.zeplin.io/project/5e9efc30726e106958cc43de/screen/5ea803abd57b80237b6ef01d

app/src/main/java/io/gnosis/safe/ui/safe/overview/SafeOverviewViewModel.kt
```diff
@@ -0,0 +1,39 @@
...
+@Inject
+constructor(
+    private val preferencesManager: PreferencesManager
+) : ViewModel() {
+
+    lateinit var termsBottomSheetDialog: TermsBottomSheetDialog
```
jpalvarezl, line 14: Let's not keep reference to views in the viewModels. This can lead to leaks and is a bit of an anti pattern under MVVM. Then for testing, you will have through 7 different types of mocking hell to verify something that you could just do by emitting a data class instance as a view state update. I can show you around the app if you want an example.

app/src/main/java/io/gnosis/safe/ui/safe/overview/SafeOverviewViewModel.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    fun checkTerms(advance: () -> Unit) {
+        if (getTermsAgreed()) {
+            advance()
+        } else {
+            termsBottomSheetDialog.onUserClicksAgree = {
+                advance()
+                setTermsAgreed(true)
+            }
+            termsBottomSheetDialog.show()
```
jpalvarezl, line 32: This should happen in the view consuming states from the viewModel.

jpalvarezl:
Yeah I think you could do that. Right now we just put an artificial delay to get right into the main nav stuff. You could inject the TermsViewModel there and get stuff shown if required.

biafra23:
@jpalvarezl Please have another look. I think I covered everything.

app/src/main/java/io/gnosis/safe/ui/safe/terms/TermsViewModel.kt
```diff
@@ -0,0 +1,31 @@
...
+
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.ViewModel
+import javax.inject.Inject
+
+class TermsViewModel @Inject constructor(
+    private val termsChecker: TermsChecker
+) : ViewModel() {
+
+    lateinit var advance: () -> Unit
```
jpalvarezl, line 11: I seems to me you don't need this `var`. You update it, if `termsChecker.getTermsAgreed` returns false, but when it returns true, you call the method passed in the argument. `invoke()` gets never called for this reference, as far as I can see.

app/src/main/java/io/gnosis/safe/ui/safe/terms/TermsBottomSheetDialog.kt
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    private fun inject() {
+        DaggerViewComponent.builder()
+            .applicationComponent(HeimdallApplication[context])
+            .viewModule(ViewModule(context))
+            .build().inject(this)
+    }
+
+    fun checkTerms(function: () -> Unit) = viewModel.checkTerms { function() }
```
jpalvarezl, line 62: Hmm so here you are passing the context from the SplashActivity, to the BottomSheetDialog and then into the viewModel, and as you are storing a reference to the method, you still have a reference to a view in the viewModel. I would suggest looking at the `EnsInputDialog`. If you reverse the direction of the communication between classes, you won't have this problem.

The viewModel notifies that the terms have been agreed to, the dialog triggers an `onAgreed` callback to which it holds a reference and then the calling context (SplashActivity) gets notified.

app/src/main/java/io/gnosis/safe/ui/safe/terms/TermsViewModel.kt
```diff
@@ -0,0 +1,31 @@
...
+        if (termsChecker.getTermsAgreed()) {
+            advance()
+        } else {
+            this.advance = advance
+            showTermsBottomSheet()
+        }
+    }
+
+    fun onAgreeClicked() {
+        termsChecker.setTermsAgreed(true)
```
jpalvarezl, line 28: Here you could use your mutableLive data to communicate that the terms have been agreed to and remove the need to store a reference to `advance`

app/src/main/res/layout/activity_splash.xml
```diff
@@ -1,7 +1,22 @@
...
-    android:fitsSystemWindows="true" />
+    android:fitsSystemWindows="true" >
+
+    <TextView
+        android:id="@+id/continue_button"
+        style="@style/PrimaryButton"
+        android:layout_width="0dp"
+        android:layout_marginEnd="@dimen/default_horizontal_margin"
+        android:layout_marginStart="@dimen/default_horizontal_margin"
+        android:text="Get started"
```
jpalvarezl, line 17: Please extract to `value/strings.xml`

app/src/main/res/navigation/main_nav.xml
```diff
@@ -60,11 +58,15 @@
...
+            app:destination="@id/navigation"
             app:launchSingleTop="true"
             app:popUpTo="@id/safeBalancesFragment"
             app:popUpToInclusive="false"/>
 
     </fragment>
+    <navigation android:id="@+id/navigation"
+        app:startDestination="@id/add_safe_nav">
+        <include app:graph="@navigation/add_safe_nav" />
+    </navigation>
```
jpalvarezl, line 70: Why is this additional level of nesting necessary?

app/src/main/res/values/dimens.xml
```diff
@@ -3,5 +3,7 @@
     <dimen name="toolbar_shadow_size">1dp</dimen>
     <dimen name="primary_button_radius">16dp</dimen>
     <dimen name="safe_blockie_size">36dp</dimen>
+    <dimen name="default_horizontal_margin">16dp</dimen>
+    <dimen name="default_line_spacing">6sp</dimen>
```
jpalvarezl, line 7: Spacing should be probably in `dp`. I would also ask you that even if you see a specific value that is not a multiple of 8, you truncate it to the closest one, we otherwise can't ensure correct alignment of views in layouts.

app/src/main/res/values/styles.xml
```diff
@@ -29,8 +29,10 @@
         <item name="android:paddingEnd">16dp</item>
         <item name="android:textColor">@drawable/primary_button_text</item>
         <item name="android:textSize">16sp</item>
-        <item name="fontFamily">sans-serif-medium</item>
+        <item name="fontFamily">sans-serif</item>
```
jpalvarezl, line 32: This is effectively changing all buttons using this style. Did you do so because of the mocks? I would add this value locally where needed.

app/src/main/res/values/styles.xml
```diff
@@ -29,8 +29,10 @@
...
+        <item name="android:textStyle">bold</item>
```
jpalvarezl, line 33: This should also be moved to where you added your feature and where exactly you saw this style. I am not sure we want his a the app wide default value.

app/src/main/res/values/styles.xml
```diff
@@ -29,8 +29,10 @@
...
         <item name="textAllCaps">false</item>
+        <item name="android:lineSpacingExtra">@dimen/default_line_spacing</item>
```
jpalvarezl, line 35: This one should be definitely be moved as well or maybe removed. Why do you need line spacing in a button? Buttons should have 1 line of text and in the worst case scenario ellipse. Maybe use padding to adjust to match the mocks but please also do so in the button in that particular view.

app/src/main/res/values/styles.xml
```diff
@@ -94,7 +101,7 @@
 
     <style name="ToolbarTitle">
         <item name="fontFamily">sans-serif</item>
-        <item name="android:lineSpacingExtra">6sp</item>
+        <item name="android:lineSpacingExtra">@dimen/default_line_spacing</item>
```
jpalvarezl, line 104: I don't think this should be here. Is it really necessary for the toolbar?

app/src/main/res/values/styles.xml
```diff
@@ -55,23 +57,28 @@
...
         <item name="fontFamily">sans-serif-medium</item>
-        <item name="android:lineSpacingExtra">6sp</item>
+        <item name="android:lineSpacingExtra">@dimen/default_line_spacing</item>
         <item name="android:textColor">@color/gnosis_dark_blue</item>
         <item name="android:textSize">16sp</item>
         <item name="android:textStyle">normal</item>
     </style>
 
+    <style name="TextDark.Bold" parent="TextDark">
+        <item name="fontFamily">sans-serif</item>
```
jpalvarezl, line 75: This should be moved as well to where it's used.  I am not sure we are getting much value for defining a style that only sets 2 variables.

app/src/test/java/io/gnosis/safe/ui/safe/terms/TermsCheckerTest.kt
```diff
@@ -0,0 +1,55 @@
...
+
+class TermsCheckerTest {
+    private lateinit var preferences: TestPreferences
+    private lateinit var preferencesManager: PreferencesManager
+
+    @Before
+    fun setup() {
+        preferences = spyk()
+        val application = mockk<Application>().apply {
+            every { getSharedPreferences(any(), any()) } returns preferences
```
jpalvarezl, line 23: There is a test util for shared preferences in svalinn that would be better suited for this. I would also ask you to verify mock calls as they are used. Check the mockk documentation and tests that are already merged so you get a clearer idea of what would be nice to have

app/src/test/java/io/gnosis/safe/ui/safe/terms/TermsCheckerTest.kt
```diff
@@ -0,0 +1,55 @@
...
+import junit.framework.Assert.assertFalse
+import junit.framework.Assert.assertTrue
+import org.junit.Before
+import org.junit.Test
+import pm.gnosis.svalinn.common.PreferencesManager
+import pm.gnosis.tests.utils.TestPreferences
+
+class TermsCheckerTest {
+    private lateinit var preferences: TestPreferences
+    private lateinit var preferencesManager: PreferencesManager
```
jpalvarezl, line 17: `TestPreferences` from svalinn test utils. Check the `SafeRepositoryTest.kt` for reference, if you need.

app/src/test/java/io/gnosis/safe/ui/safe/terms/TermsViewModelTest.kt
```diff
@@ -0,0 +1,73 @@
...
+import junit.framework.Assert.assertTrue
+import org.junit.Before
+import org.junit.Test
+import pm.gnosis.svalinn.common.PreferencesManager
+import pm.gnosis.tests.utils.TestPreferences
+
+class TermsViewModelTest {
+
+    private lateinit var preferences: TestPreferences
+    private lateinit var preferencesManager: PreferencesManager
```
jpalvarezl, line 19: `TestPreferences` from svalinn.

app/src/main/java/io/gnosis/safe/ui/safe/terms/TermsChecker.kt
```diff
@@ -0,0 +1,21 @@
+package io.gnosis.safe.ui.safe.terms
+
+import pm.gnosis.svalinn.common.PreferencesManager
+import pm.gnosis.svalinn.common.utils.edit
+import javax.inject.Inject
+
+class TermsChecker @Inject constructor(
+    private val preferencesManager: PreferencesManager
+) {
+    fun setTermsAgreed(value: Boolean) {
```
jpalvarezl, line 10: As you can see in the `SafeRepository`, any interaction with sharedPreferences is assumed to be asynchronous. I would ask you to add `suspend` to both this and the `getTermsAgreed` method.

app/src/test/java/io/gnosis/safe/ui/safe/terms/TermsCheckerTest.kt
```diff
@@ -0,0 +1,55 @@
...
+    fun setup() {
+        preferences = spyk()
+        val application = mockk<Application>().apply {
+            every { getSharedPreferences(any(), any()) } returns preferences
+        }
+        preferencesManager = PreferencesManager(application)
+    }
+
+    @Test
+    fun `When terms agreed preferences should contain true`() {
```
jpalvarezl, line 29: This is something we should have probably communicated to you sooner, but the convention for naming tests is:
For unit tests:
`methodName (input and precondition brief description) should <succeed|fail|returns|throws (type|value)>`
For instrumented tests we use underscores instead of whitespace and double underscore to replace parenthesis for the input section. If it's unclear you can see other test or let's have a chat.

app/src/main/res/navigation/main_nav.xml
biafra23: I don't remember touching this file.

app/src/main/res/values/dimens.xml
biafra23: I took this value from Sashas Layout. IIRC we had this in th stayles already. MAybe this is sp because it refers to fonts?

app/src/main/res/values/styles.xml
biafra23: I just took this from the Zeplin layout. You are probably right.

app/src/main/res/values/styles.xml
biafra23: I took this from the Zeplin layout. The title is multiline.

biafra23: I did not add this. I just extracted it to a dimension. Maybe @elgatovital can help?

app/src/main/java/io/gnosis/safe/HeimdallApplication.kt
```diff
@@ -20,7 +22,9 @@ class HeimdallApplication : MultiDexApplication(), ComponentProvider {
 
     override fun onCreate() {
         super.onCreate()
-
+        if (BuildConfig.DEBUG) {
+            Timber.plant(DebugTree())
+        }
```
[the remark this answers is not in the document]
biafra23: https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe/318

app/src/main/java/io/gnosis/safe/ui/safe/terms/TermsViewModel.kt
biafra23: It gets called in two cases:
1. Terms have already been agreed and that has been stored in preferences
2. When the user clicks on the agree button

app/src/test/java/io/gnosis/safe/ui/safe/terms/TermsCheckerTest.kt
biafra23: I tried to use the same way as `SafeRepositoryTest`.

app/src/main/java/io/gnosis/safe/di/components/ApplicationComponent.kt
```diff
@@ -37,6 +39,10 @@ interface ApplicationComponent {
 
     fun appDispatchers(): AppDispatchers
 
+    fun preferencesManager(): PreferencesManager
```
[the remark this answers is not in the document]
biafra23: It's not. Removed in 851fdcb1605625e9fb73ea5ada564570d9894960

app/src/main/java/io/gnosis/safe/ui/safe/terms/TermsChecker.kt
```diff
@@ -0,0 +1,21 @@
+package io.gnosis.safe.ui.safe.terms
+
+import pm.gnosis.svalinn.common.PreferencesManager
+import pm.gnosis.svalinn.common.utils.edit
+import javax.inject.Inject
+
+class TermsChecker @Inject constructor(
```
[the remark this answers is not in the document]
biafra23: Done in dd8ac8c6b4ea7e8c53b32b90f942699a7aadb031

app/src/main/res/navigation/main_nav.xml
biafra23: It seems this was merged in from master.

app/src/main/res/values/styles.xml
biafra23: I did so, because all primary buttons (I could find) are using this `fontFamily` and `textStyle` bold.

app/src/main/res/values/styles.xml
biafra23: All primary buttons have this textStyle. Isn't that the main use case for styles? Define them once, and apply them in several places.

app/src/main/res/values/styles.xml
biafra23: If we can apply this style in more than one place, it makes sense to have it. Otherwise we will have to change this style in several places. 
This style is used (for example) also in the name of the safe in safe settings.

app/src/test/java/io/gnosis/safe/ui/safe/terms/TermsCheckerTest.kt
biafra23: Done in 14e215271c75c8b1d0a0a002d38a06bd6af89c49

app/src/main/res/values/styles.xml
biafra23: Fixed in cf62d47e2a713d1b7b9de2f41c71e15e0cd55eda

app/src/main/java/io/gnosis/safe/ui/safe/terms/TermsChecker.kt
biafra23: For writing to SharePreferences that is definitely not necessary because `SharedPreferences.edit()` uses `Editor.apply()`. For reading it's not clear, whether it is recommended. Some say so, some say so. 

If I use suspend there, then I also have to inherit from BaseStateViewModel and use AppDispatchers, right?

biafra23: `suspend` on `setActiveSafe()` is redundant (according to Android Studio) and on `getActiveSafe()` it is only necessary because of the (network) call to `getSafeBy()`. Not because of the interaction with the shared preferences.

app/src/main/java/io/gnosis/safe/ui/safe/terms/TermsViewModel.kt
biafra23: advance is removed in 09ed4521d64e0e7151f5e6bdeb4794b9d1622861

app/src/main/java/io/gnosis/safe/ui/safe/terms/TermsBottomSheetDialog.kt
```diff
@@ -0,0 +1,77 @@
...
+import androidx.appcompat.app.AppCompatActivity
+import androidx.lifecycle.Observer
+import com.google.android.material.bottomsheet.BottomSheetDialog
+import io.gnosis.safe.HeimdallApplication
+import io.gnosis.safe.R
+import io.gnosis.safe.di.components.DaggerViewComponent
+import io.gnosis.safe.di.modules.ViewModule
+import io.gnosis.safe.ui.safe.terms.TermsViewModel.ViewAction.ShowBottomSheet
+import io.gnosis.safe.ui.safe.terms.TermsViewModel.ViewAction.TermsAgreed
+import kotlinx.android.synthetic.main.bottom_sheet_terms_and_conditions.*
```
elgatovital, line 14: can we use view binding and not use synthetics at all?

elgatovital:
looks good except for synthetics. I think it is better to use view bindings instead. The code will be not cluttered with long view ids. And we use view bindings in the project already.

app/src/main/java/io/gnosis/safe/ui/safe/terms/TermsBottomSheetDialog.kt
biafra23: Why not in this case? They are used in several other Fragments and Dialogs.

biafra23: Changed in 47e40600dac8a57c30e85bdd42fcddafd5263afc

biafra23: @elgatovital So as I understand it,: We use view bindings in all the new code. And at some point get rid of the synthetics in existing code?

app/src/test/java/io/gnosis/safe/ui/safe/terms/TermsViewModelTest.kt
```diff
@@ -0,0 +1,80 @@
...
+
+        viewModel.state.test().assertValues(
+            TermsViewModel.TermsOfUseState(TermsViewModel.ViewAction.ShowBottomSheet)
+        )
+
+        viewModel.onAgreeClicked()
+
+        viewModel.state.test().assertValues(
+            TermsViewModel.TermsOfUseState(TermsViewModel.ViewAction.TermsAgreed)
+        )
```
jpalvarezl, line 78: This unit tests seems to be testing 2 things. I think it would be best to separate it in 2. If you can't, I would say it's a hint that the code could be coupled in a way that would represent a code smell. Also, please add `verify` blocks.

app/src/main/java/io/gnosis/safe/ui/safe/terms/TermsBottomSheetDialog.kt
```diff
@@ -0,0 +1,79 @@
...
+    private fun inject() {
+        DaggerViewComponent.builder()
+            .applicationComponent(HeimdallApplication[context])
+            .viewModule(ViewModule(context))
+            .build().inject(this)
+    }
+
+    fun checkTerms(function: () -> Unit) {
+        viewModel.checkTerms()
+        advance = function
```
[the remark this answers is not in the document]
biafra23: When I decouple showing the Dialog and setting the onAgreeClickedLlistener. Then I don't need a Viewmodel anymore and part of the logic is untestable because it is in the Activity. See: ccf100aff95dbd19d93282b2e19efa4460a1c2cc

app/src/test/java/io/gnosis/safe/ui/safe/terms/TermsCheckerTest.kt
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Test
+    fun `setTermsAgreed (true) should save true under TERMS_AGREED`() {
+        val termsChecker = TermsChecker(preferencesManager)
+        runBlocking {
+            termsChecker.setTermsAgreed(true)
+
+            assertTrue(preferences.getBoolean(TERMS_AGREED, false))
+        }
```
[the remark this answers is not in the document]
biafra23: What is the advantage of testing the right mock methods were called? Does it matter how TermsChecker interacts with the preferences? Isn't that testing an implementation detail?

app/src/main/res/layout/bottom_sheet_terms_and_conditions.xml
```diff
@@ -0,0 +1,138 @@
...
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:background="@color/white"
+    android:fitsSystemWindows="true">
+
+    <ImageView
+        android:id="@+id/bottom_sheet_terms_and_conditions_first_bullet"
```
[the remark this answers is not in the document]
biafra23: How do I indent the second line correctly?

PR #586, opened 2020-05-05
Transaction adapter list with generic implementation

### LINKED ISSUE #572: Create adapter & view holders for tx list

jpalvarezl:
Closes  #572.

- Defined transaction type models
- Defined factory and view holders
- Added generic base adapter for handling diverse view types

@gnosis/mobile-devs


jpalvarezl:
I am not exactly sure if this task covers the full scope of what was expected in the issue. 

I added some models as type holders but they don't have any significant fields so far.

app/src/main/java/io/gnosis/safe/ui/adapter/BaseAdapter.kt
```diff
@@ -0,0 +1,45 @@
...
+import androidx.recyclerview.widget.RecyclerView
+
+class BaseAdapter<VH, T>(
+    private val factory: BaseFactory<VH>
+) : RecyclerView.Adapter<VH>() where VH : BaseViewHolder<T> {
+
+    private val items = mutableListOf<T>()
+
+    @Deprecated("Unsafe")
+    fun setItemsUnsafe(items: List<T>) {
```
rmeissner, line 14: Do we plan on using the DiffUtils?

jpalvarezl: Absolutely. We already have one adapter that makes use of it afaik. This is a bit of a naive implementation. I just need to verify what's the scope of the issue defined xD

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionViewHolderFactory.kt
```diff
@@ -0,0 +1,76 @@
...
+import io.gnosis.safe.databinding.EntryTransactionCancelledBinding
+import io.gnosis.safe.databinding.EntryTransactionFailedBinding
+import io.gnosis.safe.databinding.EntryTransactionPendingBinding
+import io.gnosis.safe.databinding.EntryTransactionSuccessBinding
+import io.gnosis.safe.ui.adapter.BaseFactory
+import io.gnosis.safe.ui.adapter.BaseViewHolder
+import io.gnosis.safe.ui.adapter.UnsupportedViewType
+
+enum class TransactionViewType {
+    PENDING, SUCCESS, CANCELLED, FAILED
```
biafra23, line 16: Should we have different ViewHolders based on the state (PENDING, SUCCESS, CANCELLED, FAILED) of the transaction or the type (custom_transaction, settings_change, transfer, creation_call(?)) of the transaction?

app/src/main/java/io/gnosis/safe/ui/base/Adapter.kt
```diff
@@ -44,11 +44,11 @@ abstract class Adapter<T, VH : Adapter.ViewHolder<T>> : RecyclerView.Adapter<VH>
     }
 
     data class Data<out T>(
```
elgatovital, line 46: what is the intention of the Data class?

jpalvarezl: I think it's a form of change dispatcher. I haven't looked at it closely, it's from the old app.

elgatovital: I think we do not need it. Also, we would we want to expose diff result. ParentId is also something very special and should not be generalized.

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionViewHolderFactory.kt
elgatovital: I would say type

elgatovital: our tx layout segregation is also based on type

jpalvarezl: This is an outdated change.

app/src/main/java/io/gnosis/safe/ui/base/Adapter.kt
jpalvarezl: So the scope of this PR was to define the adapter and viewHolders for TX list. I don't want to start refactoring code that is from the legacy app in here. This Adapter is also extended by the `LifecycleAdapter` so I am sorry but I don't want this to blow up like other tasks have. If we want to get rid of this, then I say we do so in a different issue.

data/src/main/java/io/gnosis/data/models/Transaction.kt
```diff
@@ -0,0 +1,11 @@
+package io.gnosis.data.models
+
+sealed class Transaction(open val nonce: String) {
+
+    data class ChangeMastercopy(override val nonce: String) : Transaction(nonce)
```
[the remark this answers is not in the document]
jpalvarezl: I put that just because data classes need at least one member variable/value. `nonce` seemed like a good candidate as all the subtypes will have it for sure.

jpalvarezl:
Thanks @biafra23, I will wait in case @elgatovital has any additional comments before merging.

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionViewHolderFactory.kt
```diff
@@ -0,0 +1,103 @@
...
+    override fun viewTypeFor(item: Transaction): Int =
+        when (item) {
+            is Transaction.ChangeMastercopy -> TransactionViewType.CHANGE_MASTERCOPY
+            is Transaction.SettingsChange -> TransactionViewType.SETTINGS_CHANGE
+            is Transaction.Transfer -> TransactionViewType.TRANSFER
+            is Transaction.ChangeMastercopyQueued -> TransactionViewType.CHANGE_MASTERCOPY_QUEUED
+            is Transaction.SettingsChangeQueued -> TransactionViewType.SETTINGS_CHANGE_QUEUED
+            is Transaction.TransferQueued -> TransactionViewType.TRANSFER_QUEUED
+            else -> throw UnsupportedViewType(javaClass.name)
+        }.ordinal
```
biafra23, line 51: Do we risk breaking the code when (accidentally) reordering the enum? Should we detect this with a unit test?

jpalvarezl: There is no risk of that. `ordinal` returns the index within the enum whatever it may be at the time of compilation. As long as we use `ordinal` instead of hard coded indexes everything should be safe. 
Unit tests would be a bit hard at the moment. `layout` calls static methods from `ViewBinding` classes. 
I suppose I could try testing, see that the instances of viewHolders returned conform to the expected type, etc.

jpalvarezl: @biafra23 I think that the tests might not work. `ViewBinding` doesn't seem to fly with unit tests. I am afraid this might not work on build kite. I suppose when we start with UI testing this would be feasible to test.

app/src/main/java/io/gnosis/safe/ui/base/Adapter.kt
elgatovital: It's not about blowing things up, but about not having code that is not needed. This just leads to technical debt and over-engineering. You can create an issue for this (Base Adapter, DiffUtils), link it in this pr, and merge. Everything else looks good.

jpalvarezl:
The issue for evaluating the fitness of `Data` class as an update dispatcher: https://github.com/gnosis/safe-android/issues/610

PR #595, opened 2020-05-11
Add tracking

### LINKED ISSUE #552: Tracking
https://github.com/gnosis/safe/issues/299

elgatovital:
Closes #552.

- Add tracker
- Track safe addition & removal
- Logging of push settings changes will be done when implementing push notifications

@gnosis/mobile-devs


rmeissner:
Does this include screen tracking and the default user properties?

Also would it be possible to add tests for this?

elgatovital:
> Does this include screen tracking and the default user properties?
> 
> Also would it be possible to add tests for this?

Not yet. Screen tracking & tests are coming shortly

jpalvarezl:
Could you also rename the PR title, please :)

app/src/main/java/io/gnosis/safe/Tracker.kt
```diff
@@ -0,0 +1,76 @@
+package io.gnosis.safe
+
+import android.content.Context
+import android.os.Bundle
+import com.google.firebase.analytics.FirebaseAnalytics
+import com.google.firebase.crashlytics.FirebaseCrashlytics
+import io.gnosis.safe.di.ApplicationContext
+import javax.inject.Inject
+
+class Tracker @Inject constructor(@ApplicationContext context: Context) {
```
jpalvarezl, line 10: Should this be marked as `@Singleton` ?

app/src/main/java/io/gnosis/safe/ui/safe/add/AddSafeNameViewModel.kt
```diff
@@ -32,6 +32,7 @@ class AddSafeNameViewModel
                 updateState { AddSafeNameState(ViewAction.ShowError(it)) }
             }.onSuccess {
                 updateState { AddSafeNameState(ViewAction.CloseScreen) }
+                tracker.logSafeAdd(safeRepository.getSafes().count())
```
jpalvarezl, line 35: Generally, I would say tracking belongs in the UI but I see that here there is really no work around that. Maybe passing it as part of the `State` update? Is that something we should try to establish a pattern around?

app/src/main/java/io/gnosis/safe/ui/safe/settings/SafeSettingsViewModel.kt
```diff
@@ -26,16 +28,22 @@ class SafeSettingsViewModel @Inject constructor(
 
     fun removeSafe() {
         safeLaunch {
-            val safe = safeRepository.getActiveSafe()
-            safe?.let {
-                safeRepository.removeSafe(safe)
+            runCatching {
```
jpalvarezl, line 31: Why is this in a `runCatching` block?

app/src/test/java/io/gnosis/safe/ui/safe/add/AddSafeNameViewModelTest.kt
```diff
@@ -33,7 +35,7 @@ class AddSafeNameViewModelTest {
 
     @Before
     fun setup() {
-        viewModel = AddSafeNameViewModel(repositories, appDispatchers)
+        viewModel = AddSafeNameViewModel(repositories, appDispatchers, tracker)
```
jpalvarezl, line 38: It would be good to add `verify { }` blocks for tracker methods.

app/src/main/java/io/gnosis/safe/ui/safe/settings/SafeSettingsViewModel.kt
elgatovital: no reason really. In case something goes wrong with removing

app/src/main/java/io/gnosis/safe/ui/safe/add/AddSafeNameViewModel.kt
elgatovital: why should it be part of the UI? We could also track some background actions or events. Why can't we have tracking logic in view models?

app/src/main/java/io/gnosis/safe/Tracker.kt
elgatovital: It's provided as a singleton in a module. Is @Singleton annotation necessary? Or you mean it would improve readability?

rmeissner: The annotation changes the generated code and in some cases does some checking. Easiest to see that if you look at the generated code. The singleton should be cached and only be created once, while the unscoped one will be used like a factory and be created each time again

app/src/main/java/io/gnosis/safe/ui/safe/add/AddSafeNameViewModel.kt
jpalvarezl: Well maybe viewModels are fine, but I would draw the line on the `data` module. Adding tracking indiscriminately will make it harder to extract functionality in the future. So at some point we would like to move something to svalinn, this could get on the way.

app/src/main/java/io/gnosis/safe/Tracker.kt
jpalvarezl: ☝️ exactly what Richard said. Particularly, because it could be handy as well eventually to have a reference to `this` in a `companion object` so that we can track in viewHolders when constructor or member injection is not posible.

app/src/main/java/io/gnosis/safe/Tracker.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    object Param {
+        val NUM_SAFES = "num_safes"
+        val PUSH_INFO = "push_info"
+    }
+
+    object ParamValues {
+        val PUSH_ENABLED = "enabled"
+        val PUSH_DISABLED = "disabled"
+    }
```
jpalvarezl, line 59: I think for now it's ok, but we could provide something with callbacks whereby we build the event and just submit it leveraging polymorphism. Maybe it's overkill too xD. But it would be nice to have something like: 
```suggestion
    } object PUSH_INFO: Trackable {
       override val name = "push_info"
       //...
    }
    
   interface Trackable {
   val name: String
   }
```

elgatovital:
manual screen tracking is addressed here: https://github.com/gnosis/safe-android/issues/599

app/src/main/java/io/gnosis/safe/ui/safe/add/AddSafeNameViewModel.kt
```diff
@@ -12,7 +11,8 @@ import javax.inject.Inject
 class AddSafeNameViewModel
 @Inject constructor(
     repositories: Repositories,
-    appDispatchers: AppDispatchers
+    appDispatchers: AppDispatchers,
+    private val tracker: Tracker
```
jpalvarezl, line 15: Maybe it would be good to have this as a singleton already to not have to inject it everywhere.

app/src/main/java/io/gnosis/safe/Tracker.kt
```diff
@@ -0,0 +1,60 @@
...
+            firebaseAnalytics.logEvent(name, bundle)
+        } catch (e: Exception) {
+            logException(e)
+        }
+    }
+
+    private fun logException(exception: Exception) {
+        FirebaseCrashlytics.getInstance().recordException(exception)
+    }
+
```
jpalvarezl, line 46: Singleton implementation suggestion part2:
```suggestion
companion object {
      private lateinit var tracker: Tracker
      
      fun get(): Tracker = tracker 
}
```

app/src/main/java/io/gnosis/safe/Tracker.kt
```diff
@@ -0,0 +1,60 @@
...
+import com.google.firebase.crashlytics.FirebaseCrashlytics
+import io.gnosis.safe.di.ApplicationContext
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class Tracker @Inject constructor(@ApplicationContext context: Context) {
+
+    private val firebaseAnalytics: FirebaseAnalytics = FirebaseAnalytics.getInstance(context)
+
```
jpalvarezl, line 15: Singleton implementation suggestion part1:
```suggestion
init {
 this.tracker = this
}
```

app/src/main/java/io/gnosis/safe/ui/safe/add/AddSafeNameViewModel.kt
```diff
@@ -32,6 +32,7 @@ class AddSafeNameViewModel
                 updateState { AddSafeNameState(ViewAction.ShowError(it)) }
             }.onSuccess {
                 updateState { AddSafeNameState(ViewAction.CloseScreen) }
+                tracker.setNumSafes(safeRepository.getSafes().count())
```
jpalvarezl, line 35: With a singleton implementation I suggested then we could
```suggestion
                Tracker.get().setNumSafes(safeRepository.getSafes().count())
```

and get rid of the `tracker` in the constructor

PR #596, opened 2020-05-11
Coins balance fragment

### LINKED ISSUE #553: View ETH and ERC20 token balance
- User can see their ETH balance, ETH is always on top
- User can see the balance of all ERC20 tokens with balance > 0, tokens are ordered by their names, ascending
- User sees fiat value in USD per asset
- ~User can select the reference fiat currency from a list somewhere in the app~
  - Selecting the fiat currency will be part of a separate epic. (#304)

jpalvarezl:
Closes #553 .

- Added viewPager with Coins and Collectibles (collectibles has a duplicate of coins as a placeholder for now)
- Simplified implementation of TokenRepository to not use the RelayService
- Storing ERC20 token info into local DAO as PoC for migration tests (could be useful in the future)

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/safe/balances/coins/CoinBalanceAdapter.kt
```diff
@@ -0,0 +1,42 @@
...
+    override fun onBindViewHolder(holder: CoinBalanceViewHolder, position: Int) {
+        holder.bind(balances[position])
+    }
+}
+
+class CoinBalanceViewHolder(private val viewBinding: ItemCoinBalanceBinding) : RecyclerView.ViewHolder(viewBinding.root) {
+
+    fun bind(balanceItem: Balance) {
+        with(viewBinding) {
+            logoImage.loadTokenLogo(Picasso.get(), balanceItem.token.logoUrl)
```
rmeissner, line 36: This is a little bit pattern mixing: dependency injection vs java singletons

app/src/main/java/io/gnosis/safe/ui/safe/balances/SafeBalancesFragment.kt
```diff
@@ -47,5 +69,17 @@ class SafeBalancesFragment : SafeOverviewBaseFragment<FragmentSafeBalancesBindin
...
+            adapter?.notifyItemChanged(currentItem)
+        }
     }
 }
+
+class BalancesPagerAdapter(fragment: Fragment) : FragmentStateAdapter(fragment) {
+
+    override fun getItemCount(): Int = 2
+
+    override fun createFragment(position: Int): Fragment =
```
rmeissner, line 82: We don't have a collectible fragment yet (this is part of phase 2). I would only show one fragment (we want to make a public release before the collectibles feature)

app/src/main/java/io/gnosis/safe/ui/safe/balances/coins/CoinsFragment.kt
```diff
@@ -0,0 +1,70 @@
...
+
+                }
+            }
+        })
+        viewModel.load()
+    }
+
+    private fun handleError(throwable: Throwable) {
+        snackbar(requireView(), R.string.error_loading_balances)
+        throwable.printStackTrace()
```
rmeissner, line 62: This should not be here (should be part of the timber tree)

data/src/main/java/io/gnosis/data/backend/RelayServiceApi.kt
```diff
@@ -0,0 +1,18 @@
+package io.gnosis.data.backend
+
+import io.gnosis.data.BuildConfig
+import retrofit2.http.GET
+import retrofit2.http.Path
+import io.gnosis.data.backend.dto.ServiceTokenInfo
+
+@Deprecated("Use transaction service API if available")
+interface RelayServiceApi {
```
rmeissner, line 9: We should not need this api, right?

data/src/main/java/io/gnosis/data/repositories/TokenRepository.kt
```diff
@@ -0,0 +1,29 @@
...
+class TokenRepository(
+    private val erc20TokenDao: Erc20TokenDao,
+    private val transactionServiceApi: TransactionServiceApi
+) {
+
+    suspend fun loadBalanceOf(safe: Solidity.Address): List<Balance> =
+        transactionServiceApi.loadBalances(safe.asEthereumAddressChecksumString())
+            .map {
+                val token = it.tokenAsErc20Token()
+                erc20TokenDao.insertToken(token)
```
rmeissner, line 21: Do we load the token anywhere? Or is this just for future caching?

jpalvarezl: Future caching. But maybe it is pointless. It was basically to not let all the time I put into the migration tests and all that go to waste 😆 If I remove this, then I can't justify having an ERC20TokenDao and then the futility back propagates :D

app/src/main/java/io/gnosis/safe/ui/safe/balances/SafeBalancesFragment.kt
jpalvarezl: Hmm, that messes up a bit the whole navigation that @elgatovital prepared. If this is absolutely necessary, then I would create duplicate files with the suffix `tmp` otherwise this will really blow up in terms of complexity (because of the safe change updates and such)

rmeissner: Why? Just add a <fragment> for now to the layout

jpalvarezl: Or alternatively, could I just simple put an empty fragment under collectibles with a textview saying "Coming soon ... " or something like that? I realise that creating temp files will just make merging new feature into the graph almost impossible.

jpalvarezl: I was afraid of the callbacks not working, but never mind, you were right. I just made everything `gone` (so that I don't have to get rid of the code for the viewBinding) and added the fragment

app/build.gradle
```diff
@@ -48,12 +48,6 @@ android {
 
         ext.betaDistributionReleaseNotes = System.getenv("APP_RELEASE_NOTES")
 
-        javaCompileOptions {
```
elgatovital, line 51: this will be added later eventually. Can we just leave it here?

app/src/main/java/io/gnosis/safe/ui/safe/add/AddSafeNameFragment.kt
```diff
@@ -51,7 +54,14 @@ class AddSafeNameFragment : BaseFragment<FragmentAddSafeNameBinding>() {
                 is AddSafeNameState -> {
                     state.viewAction?.let { action ->
                         when (action) {
-                            is BaseStateViewModel.ViewAction.CloseScreen -> findNavController().navigateUp()
+                            is BaseStateViewModel.ViewAction.CloseScreen -> {
+                                (requireContext().getSystemService(Context.INPUT_METHOD_SERVICE) as? InputMethodManager?)
```
elgatovital, line 58: this could be extracted to a private method or placed in utils as extension function

app/src/main/java/io/gnosis/safe/ui/safe/balances/SafeBalancesFragment.kt
```diff
@@ -27,9 +35,24 @@ class SafeBalancesFragment : SafeOverviewBaseFragment<FragmentSafeBalancesBindin
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
+        with(binding) {
+            val adapter = BalancesPagerAdapter(this@SafeBalancesFragment)
+            balancesContent.adapter = adapter
+            TabLayoutMediator(balancesTabBar, balancesContent, true) { tab, position ->
+                if (position % 2 == 0) {
```
elgatovital, line 42: I suggest having enum Tab or Page { COINS, COLLECTABLES } and using this enum in mediator
when (Tab.values()[position]) {
   Tab.COINS ->
   Tab.COLLECTABLES ->
...
this will make it more readable

app/src/main/java/io/gnosis/safe/ui/safe/balances/coins/CoinsViewModel.kt
```diff
@@ -0,0 +1,44 @@
...
+
+    override fun initialState(): CoinsState = CoinsState(loading = false, refreshing = false, viewAction = null)
+
+    init {
+        safeLaunch {
+            safeRepository.activeSafeFlow().collect { load() }
+        }
+    }
+
+    fun load(isRefresh: Boolean = false) {
```
elgatovital, line 27: you use `isRefresh` here and `refreshing` in a state class. It would be refreshing to have refreshing in both places

app/src/main/java/io/gnosis/safe/utils/ImageUtils.kt
```diff
@@ -0,0 +1,31 @@
...
+        icon == "local::ethereum" -> {
+            setImageResource(R.drawable.ic_ethereum_logo)
+        }
+        icon?.startsWith("local::") == true -> {
+//            setImageResource(R.drawable.circle_background)
+        }
+        !icon.isNullOrBlank() ->
+            Picasso.get()
+                .load(icon)
+//                .placeholder(R.drawable.circle_background)
```
elgatovital, line 23: why do we have commented out code here?

data/src/main/java/io/gnosis/data/repositories/TokenRepository.kt
elgatovital: should we remove these db inserts for now? No reason to have them here after we agreed to always load fresh data when opening screens

app/build.gradle
rmeissner: Shouldn't that be on the data module and not here, or is it required also here?

app/src/main/java/io/gnosis/safe/ui/safe/add/AddSafeNameFragment.kt
rmeissner: Didn't we have utils in the old app for this?

data/src/main/java/io/gnosis/data/repositories/TokenRepository.kt
jpalvarezl: I would prefer still keeping it. Like I explained before, I would have to then remove the DAO, tests for it, migration tests, etc. I don't see any harm in caching this stuff, specially if eventually we are going to have to implement it anyway. I think the cost of re-writing this in the future far overweights the performance benefits for the user.

app/build.gradle
jpalvarezl: I moved it to the `data` module because that's where we have the room entities. If you leave it in `app/build.gradle` it won't work.

app/src/main/java/io/gnosis/safe/ui/safe/add/AddSafeNameFragment.kt
jpalvarezl: In svalinn in fact. I will replace it with that.

app/src/main/java/io/gnosis/safe/utils/ImageUtils.kt
jpalvarezl: There was a bunch of stuff I didn't paste from the actual method Utils that eventually we might need. I figured I would left it comment as a reminder that the functionality was copied partially. If you want I can carry over the assets and uncomment all this code. I suppose the token icon should behave the same as in old apps, right?

elgatovital: yes. It's either icon or some placeholder

app/src/main/java/io/gnosis/safe/ui/safe/balances/SafeBalancesFragment.kt
elgatovital, line 42: why are you not using your newly created Tabs.COINS and Tabs.COLLECTIBLES here instead of having some weird positional logic ;)

jpalvarezl: oops forgot that part xD it felt weird that I could simply make that private xD

PR #601, opened 2020-05-13
Added safe address DAO validation with error branch and tests

### LINKED ISSUE #592: Add Safe: Prevent duplicates
It should not be possible to add the same Safe twice. This is how iOS solved this currently. 
IMG_6923362853E4-1

jpalvarezl:
Closes #592 .

- Added UsedSafeAddress error branch
- Added tests
- SafeRepository.isSafeAddressUsed for DAO verification

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/safe/add/AddSafeViewModel.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/safe/add/AddSafeViewModel.kt
@@ -18,17 +18,16 @@ class AddSafeViewModel
         safeLaunch {
             runCatching {
                 updateState { AddSafeState(ViewAction.Loading(true)) }
-                val validSafe = safeRepository.isValidSafe(address.asEthereumAddress() ?: throw InvalidSafeAddress())
+                val newAddress = address.asEthereumAddress() ?: throw InvalidSafeAddress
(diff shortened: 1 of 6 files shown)


app/src/main/java/io/gnosis/safe/ui/safe/add/AddSafeViewModel.kt
```diff
@@ -18,17 +18,16 @@ class AddSafeViewModel
         safeLaunch {
             runCatching {
                 updateState { AddSafeState(ViewAction.Loading(true)) }
-                val validSafe = safeRepository.isValidSafe(address.asEthereumAddress() ?: throw InvalidSafeAddress())
+                val newAddress = address.asEthereumAddress() ?: throw InvalidSafeAddress
+                val validSafe = safeRepository.isValidSafe(newAddress)
+                val isAddressUsed = safeRepository.isSafeAddressUsed(newAddress)
```
elgatovital, line 23: I would reverse this check sequence. If safe is used it will be valid, so no need to check the validity afterward. Otherwise, we need also check validity.

jpalvarezl: Good call 👍  I wrote it in a way that we make the network call if it's absolutely necessary

Issue #576, opened 2020-04-27
Dagger impl revision


jpalvarezl:
**Describe the bug**
Dagger setup now uses a wrapper class for all the Repository classes that will exist

**To Reproduce**
Creating a new repository and add a provide method in the `ApplicationModule` does not compile

**Expected behavior**
We should be able to add new repositories without the need for a wrapper class

**Additional context**
https://github.com/gnosis/safe-android/pull/575#discussion_r413912840

PR #605, opened 2020-05-18
Cleanup Dagger implementation

jpalvarezl:
Closes # https://github.com/gnosis/safe-android/issues/576

- Removed `Repositories` 
- Added all viewModels' provider methods to `ViewModule` and type bindings `ViewComponent` 
- Corrected inject calls in base classes to match the documentation's best practices

https://developer.android.com/training/dependency-injection/dagger-android

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/base/BaseActivity.kt
```diff
@@ -12,8 +12,8 @@ import io.gnosis.safe.di.modules.ViewModule
 abstract class BaseActivity : AppCompatActivity() {
 
     override fun onCreate(savedInstanceState: Bundle?) {
-        super.onCreate(savedInstanceState)
         HeimdallApplication[this].inject(this)
+        super.onCreate(savedInstanceState)
```
rmeissner, line 16: Are you sure about this? :D 
I learned `super.onCreate(...)` should always be done first ... more or less :P

app/src/main/java/io/gnosis/safe/ui/base/BaseFragment.kt
```diff
@@ -17,9 +18,9 @@ abstract class BaseFragment<T> : Fragment()
     private var _binding: T? = null
     protected val binding get() = _binding!!
 
-    override fun onCreate(savedInstanceState: Bundle?) {
-        super.onCreate(savedInstanceState)
-        inject(buildViewComponent())
+    override fun onAttach(context: Context) {
```
rmeissner, line 21: This can be potentially called multiple times, is this an assue for dagger?

app/src/main/java/io/gnosis/safe/di/components/ApplicationComponent.kt
```diff
@@ -35,14 +36,18 @@ interface ApplicationComponent {
...
 
-    fun repositories(): Repositories
-
     fun appDispatchers(): AppDispatchers
 
     fun termsChecker(): TermsChecker
 
     fun tracker(): Tracker
 
+    fun safeRepository(): SafeRepository
```
rmeissner, line 45: Are these required here? I would assume that this is not required as they are only used via view models that are create via the `ViewModelFactory` ... if this is not the case I would guess that something is not set up correctly

app/src/main/java/io/gnosis/safe/ui/base/BaseActivity.kt
jpalvarezl: > When using activities, inject Dagger in the activity's onCreate() method before calling super.onCreate() to avoid issues with fragment restoration. During the restore phase in super.onCreate(), an activity attaches fragments that might want to access activity bindings.

https://developer.android.com/training/dependency-injection/dagger-android

jpalvarezl: I have no problem reverting that though, if you feel like it could be a source of problems

app/src/main/java/io/gnosis/safe/ui/base/BaseFragment.kt
jpalvarezl: Also recommended by the documentation: 

> When using fragments, inject Dagger in the fragment's onAttach() method. In this case, it can be done before or after calling super.onAttach().

Same link as in the previous comment.

app/src/main/java/io/gnosis/safe/di/components/ApplicationComponent.kt
jpalvarezl: I think it was left over from a previous approach. Thanks for noticing.

app/src/main/java/io/gnosis/safe/ui/base/BaseActivity.kt
rmeissner: if they recommend it should be ok ... right? ... they are google ... :P

jpalvarezl: I don't trust them fully :) in the case for activity I see no harm. For fragments though, I think you have a point. But maybe it also doesn't matter if the factories get called again.

Issue #585, opened 2020-05-05
Test coverage for #580


elgatovital:
Unit tests for view models that have been introduced with #580

PR #611, opened 2020-05-21
Add app settings

### LINKED ISSUE #566: App Settings
- User can see the app version.
- ~User can see and select reference fiat currency for the entire app~
  - This will be part of a separate epic
- User can see the network version (Mainnet, Rinkeby)
- User can view Terms (Link to https://gnosis-safe.io/terms/)
- User can view Privacy Policy (Link to https://gnosis-safe.io/privacy/)
- User can view used licenses (Link to https://gnosis-safe.io/licenses/)
- Terms, PP and licenses are not hardcoded, but links that open a browser window (exactly how we do it in the current app)
- User can get in touch with us (Email, Discord, Twitter)
  - As a company we decided to moved to Discord, so let's take this opportunity.
  - Help center: https://help.gnosis-safe.io
  - Email: safe@gnosis.io
  - Discord: https://discord.gg/FPMRAwK
  - Twitter: https://twitter.com/gnosisSafe
  - Feature suggestion: https://safe.cnflx.io/
- User can see other app settings:
  - ENS resolver address
  - RPC endpoint url
  - Backend services urls (safe-relay, safe-tranasction)
  - something else?

elgatovital:
Closes #566, #585, #616 

- Add settings fragments
- Add get in touch screen
- Use pager adapter for showing no safe fragment
- Use the right scope for dependency injection of view models
- Fix navigation

@gnosis/mobile-devs


elgatovital:
Remaining issues regarding app settings: 
Appearance (Dark mode): #612 
Fiat currency: #613

biafra23:
Is it intended that the settings tab is disabled when no safe is configured? At least the app settings make sense without a safe.

elgatovital:
> Is it intended that the settings tab is disabled when no safe is configured? At least the app settings make sense without a safe.

You are right. We should reuse NoSafesFragment in case active safe is not set. App settings should be always accessible. Making these adjustments now

app/src/main/java/io/gnosis/safe/ui/safe/settings/safe/SafeSettingsFragment.kt
```diff
@@ -0,0 +1,40 @@
...
+    companion object {
+
+        fun newInstance(): SafeSettingsFragment {
+            return SafeSettingsFragment()
+        }
+    }
+}
+
+
+
```
jpalvarezl, line 40: whitespace 😁

app/src/main/java/io/gnosis/safe/ui/safe/settings/safe/SafeSettingsViewModel.kt
```diff
@@ -0,0 +1,50 @@
...
+        }
+    }
+
+    fun removeSafe() {
+        safeLaunch {
+            runCatching {
+                val safe = safeRepository.getActiveSafe()
+                safe?.let {
+                    safeRepository.removeSafe(safe)
+                    safeRepository.setActiveSafe(null)
```
jpalvarezl, line 34: maybe this line should be part of the `removeSafe` method? We should encapsulate the whole logic of removing safe within the repository, in case in the future we should be able to remove safes from somewhere else in the app.

app/src/main/java/io/gnosis/safe/ui/safe/settings/view/EndpointItem.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    var value: String? = null
+        set(value) {
+            binding.value.text = value
+            field = value
+        }
+
+    init {
+        readAttributesAndSetupFields(context, attrs)
+    }
```
jpalvarezl, line 31: Just for styling, I would say that this `init` block should go on top with the assignment of `binding` as it is part of the constructor of this class

app/src/main/java/io/gnosis/safe/ui/safe/settings/view/EndpointItem.kt
```diff
@@ -0,0 +1,52 @@
...
+            attrs,
+            R.styleable.EndpointItem,
+            0, 0
+        )
+        try {
+            applyAttributes(context, a)
+
+        } finally {
+            a.recycle()
+        }
```
jpalvarezl, line 44: This would be a more idiomatic way of writing this:

```suggestion
        context.theme.obtainStyledAttributes(
            attrs,
            R.styleable.EndpointItem,
            0, 0
        ).also{
            runCatching { applyAttributes(context, it) }
                .onFailure { Timber.e(it) }
            it.recycle()
        }
```

app/src/main/java/io/gnosis/safe/ui/safe/settings/view/SettingItem.kt
```diff
@@ -0,0 +1,65 @@
...
+            attrs,
+            R.styleable.SettingItem,
+            0, 0
+        )
+        try {
+            applyAttributes(context, a)
+
+        } finally {
+            a.recycle()
+        }
```
jpalvarezl, line 51: Same suggestion as for `EndpointItem`

app/src/main/res/values/dimens.xml
```diff
@@ -10,4 +10,8 @@
     <dimen name="item_tx_s_height">72dp</dimen>
     <dimen name="item_tx_m_height">96dp</dimen>
     <dimen name="item_tx_l_height">112dp</dimen>
+    <dimen name="item_setting_openable_height">60dp</dimen>
+    <dimen name="item_setting_non_openable_height">44dp</dimen>
```
jpalvarezl, line 14: Not a multiple of 8

app/src/main/res/values/strings.xml
```diff
@@ -43,4 +43,32 @@
     <string name="coming_soon">Coming soon</string>
     <string name="get_started">Get started</string>
     <string name="change_mastercopy">Change mastercopy</string>
+
+    <string name="settings_tab_safe">SAFE SETTINGS</string>
+    <string name="settings_tab_app">APP SETTINGS</string>
```
jpalvarezl, line 48: These shouldn't be all uppercase here. I think that tab names are upper case by default. Otherwise, I would call `String.toUpperCase()` rather than having an assert fully in upper case.

app/src/main/res/values/strings.xml
```diff
@@ -43,4 +43,32 @@
...
+    <string name="settings_app_terms_of_use">Terms of use</string>
+    <string name="settings_app_privacy_policy">Privacy policy</string>
+    <string name="settings_app_licenses">Licenses</string>
+    <string name="settings_app_get_in_touch">Get in touch</string>
+    <string name="settings_app_version">App version</string>
+    <string name="settings_app_network">Network</string>
+    <string name="settings_app_advanced">Advanced</string>
+
+    <string name="ens_resolver_address">ENS RESOLVER ADDRESS</string>
+    <string name="endpoints">ENDPOINTS</string>
```
jpalvarezl, line 60: `String.toUpperCase()` is probably a better idea.

app/src/test/java/io/gnosis/safe/ui/safe/empty/NoSafeViewModelTest.kt
```diff
@@ -0,0 +1,120 @@
...
+        stateObserver.assertValueAt(0) {
+            it.activeSafe == SAFE_1 &&
+                    it.viewAction is BaseStateViewModel.ViewAction.NavigateTo
+        }
+
+        safeRepository.removeSafe(SAFE_1)
+
+        stateObserver.assertValueAt(0) {
+            it.activeSafe == null &&
+                    it.viewAction is BaseStateViewModel.ViewAction.None
```
jpalvarezl, line 85: missing `coVerifySequence` of methods.

app/src/test/java/io/gnosis/safe/ui/safe/empty/NoSafeViewModelTest.kt
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        safeRepository.setActiveSafe(SAFE_2)
+
+        viewModel.state.test().assertValueAt(0) {
+            it.activeSafe == SAFE_2 &&
+                    it.viewAction is BaseStateViewModel.ViewAction.NavigateTo
+
+        }
```
jpalvarezl, line 112: Some `coVerify` methods would be nice.

app/src/test/java/io/gnosis/safe/ui/safe/settings/SafeSettingsViewModelTest.kt
```diff
@@ -26,36 +28,41 @@ class SafeSettingsViewModelTest {
...
 
     @Test
     fun `removeSafe - should remove safe`() = runBlockingTest {
 
         coEvery { safeRepository.getActiveSafe() } returnsMany listOf(SAFE_1, null)
+        coEvery { safeRepository.activeSafeFlow() } returns flow {
+            emit(SAFE_1)
+            emit(null)
+        }
+            .conflate()
```
jpalvarezl, line 40: I am not sure you should be calling `conflate` there. The conflation of items should happen in the viewModel when it's ready to receive the flow

app/src/test/java/io/gnosis/safe/ui/safe/settings/SafeSettingsViewModelTest.kt
```diff
@@ -26,36 +28,41 @@ class SafeSettingsViewModelTest {
...
-        }
-
-        safeSettingsViewModel.state.test().assertValueAt(0) {
-            it is SafeSettingsState.SafeSettings &&
-                    it.safe == null
+        with(stateObserver.values()[0]) {
+            val viewAction = this.viewAction
+            assert (
+                viewAction is BaseStateViewModel.ViewAction.UpdateActiveSafe &&
+                        viewAction.newSafe == null
```
jpalvarezl, line 59: So we did have a `SafeSettingsState.SafeRemoved`? I would return to that pattern where we have a viewAction that represents `SafeRemoved`

data/src/main/java/io/gnosis/data/repositories/SafeRepository.kt
```diff
@@ -49,9 +49,9 @@ class SafeRepository(
             isSupported(request.checkedResult("Valid safe check failed").asEthereumAddress())
         }
 
-    suspend fun setActiveSafe(safe: Safe) {
+    suspend fun setActiveSafe(safe: Safe?) {
         preferencesManager.prefs.edit {
-            putString(ACTIVE_SAFE, safe.address.asEthereumAddressString())
+            putString(ACTIVE_SAFE, safe?.address?.asEthereumAddressString())
```
jpalvarezl, line 54: I think this is the root of the change of the patter. I think it would be mucho better to have an additional method `removeSafe()` that sets the active safe to null, than making this nullable. We lose so much null safety just by this design decision.

jpalvarezl: What about tests for this in the `SafeRepositoryTest.kt`? I would sincerely ask you to not make this nullable. The null semantics are not always obvious, it's much more preferable to set and explicit `remove` method for this.

app/build.gradle
```diff
@@ -104,6 +105,7 @@ android {
             minifyEnabled true
             multiDexEnabled true
             signingConfig signingConfigs.release
+            buildConfigField javaTypes.STRING, "BLOCKCHAIN_NAME", asString(getKey("BLOCKCHAIN_NAME", "Mainnet"))
```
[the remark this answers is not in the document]
elgatovital: let's talk about it when we need it. Currently, blockchain name pertains only to app

app/src/main/java/io/gnosis/safe/ui/base/BaseStateViewModel.kt
```diff
@@ -39,6 +40,7 @@ abstract class BaseStateViewModel<T>(private val dispatchers: AppDispatchers) :
     interface ViewAction {
         data class Loading(val isLoading: Boolean) : ViewAction
         data class ShowError(val error: Throwable) : ViewAction
+        data class UpdateActiveSafe(val newSafe: Safe?): ViewAction
```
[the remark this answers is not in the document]
elgatovital: yes, active safe can be null. When user starts app for the first time, active safe is null. When user removes safe -> active safe becomes null. When app is closed without active safe being selected, the next time user opens it, active safe will be null. Remove event would be triggered only once and it is not suitable for our use case.

app/src/test/java/io/gnosis/safe/ui/safe/settings/SafeSettingsViewModelTest.kt
elgatovital: this will change nothing in this test, as there is only one subscriber to the state

app/src/main/java/io/gnosis/safe/ui/base/BaseStateViewModel.kt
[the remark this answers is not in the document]
jpalvarezl: This is an event that happens only if the safe is updated. So on start up of the app the you wouldn't receive this event with null ever. This event should only be emitted if the safe is updated. It is **not** used for consulting the **current** state of the active safe.

data/src/main/java/io/gnosis/data/repositories/SafeRepository.kt
elgatovital: activeSafe == null is a valid case. Having additional removeSafe just adds complexity.

app/src/test/java/io/gnosis/safe/ui/safe/settings/SafeSettingsViewModelTest.kt
jpalvarezl: https://kotlinlang.org/docs/reference/coroutines/flow.html

The part about conflations says: 

Conflation
When a flow represents partial results of the operation or operation status updates, it may not be necessary to process each value, but instead, only most recent ones. In this case, the conflate operator can be used to skip intermediate values when a collector is too slow to process them. Building on the previous example:

jpalvarezl: That may be the reason you are losing some events in between.

app/src/main/java/io/gnosis/safe/ui/base/BaseStateViewModel.kt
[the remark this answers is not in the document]
elgatovital: I'll rename it in ActiveSafe event then, because that is exactly that it was used for

app/src/main/java/io/gnosis/safe/ui/safe/settings/app/AppSettingsFragment.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    override fun inflateBinding(inflater: LayoutInflater, container: ViewGroup?): FragmentSettingsAppBinding =
+        FragmentSettingsAppBinding.inflate(inflater, container, false)
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+        with(binding) {
+            terms.setOnClickListener {
+                requireContext().openUrl(getString(R.string.link_terms_of_use))
```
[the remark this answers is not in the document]
elgatovital: but this is actually the same. And having it.context doesn't look prettier imo :)

app/src/main/java/io/gnosis/safe/ui/safe/settings/safe/SafeSettingsViewModel.kt
elgatovital: I would like to have these operations as atomic. This is the cleaner approach I think. How they are combined should be decided by our business logic

app/src/main/res/values/strings.xml
elgatovital: We should try to use strings as is. This will allow us to replace our strings with resources downloaded from localized, which we will use at the latter point. Otherwise there would be too much ambiguity regarding lower/upper case, capital letters, etc.

app/src/test/java/io/gnosis/safe/ui/safe/settings/SafeSettingsViewModelTest.kt
elgatovital: this was my intention behind using it.

app/src/main/java/io/gnosis/safe/ui/base/BaseStateViewModel.kt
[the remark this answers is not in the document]
jpalvarezl: Then that is not the correct usage. The active safe should only be available through `safeRepository.getActiveSafe()` We had already agreed that this was the single source of truth for that. It doesn't make sense to have an event attached to it, when you have the app launched. If you need the activeSafe somewhere in the app, then create your own local event and emit there, but using this as ViewAction completely misuses the infrastructure we've defined for this.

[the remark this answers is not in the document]
elgatovital: agree. ActiveSafe is now local in SafeSettignsViewModel

app/src/main/java/io/gnosis/safe/ui/safe/settings/app/GetInTouchFragment.kt
```diff
@@ -0,0 +1,119 @@
...
+
+    private fun openHelpCenter() {
+        requireContext().openUrl(getString(R.string.link_help_center))
+    }
+
+    private fun openFeatureSuggestionPage() {
+        requireContext().openUrl(getString(R.string.link_feature_suggestion))
+    }
+
+    // Do we need telegram channel or is discord sufficient?
```
jpalvarezl, line 97: This question should be asked in the issue of the epic. If we leave in code we just forget about it and probably won't address.

app/src/main/java/io/gnosis/safe/ui/safe/settings/app/GetInTouchFragment.kt
```diff
@@ -0,0 +1,119 @@
...
+    private fun openTelegramChannel() {
+        var intent = try {
+            Intent(Intent.ACTION_VIEW, Uri.parse("tg://resolve?domain=${getString(R.string.id_telegram)}"))
+        } catch (e: Exception) { //App not found open in browser
+            Intent(Intent.ACTION_VIEW, Uri.parse("http://www.telegram.me/${getString(R.string.id_telegram)}"))
+        }
```
jpalvarezl, line 103: Same as above, maybe let's use `runCatching` and Timber the exception. Also you could just have an extension function for context like so: ` private fun Context.launchExternal(appPackage: String) { /* ...  */ }` getting rid of a lot of code duplication here.

app/src/main/java/io/gnosis/safe/ui/safe/settings/app/GetInTouchFragment.kt
```diff
@@ -0,0 +1,117 @@
...
+
+    private fun sendEmail() {
+        val intent = Intent(Intent.ACTION_SENDTO)
+        intent.type = "text/html"
+        intent.data = Uri.parse("mailto:${getString(R.string.email_feedback)}")
+        intent.putExtra(Intent.EXTRA_SUBJECT, getString(R.string.feedback_subject, getString(R.string.app_name)))
+        try {
+            startActivity(intent)
+        } catch (ex: ActivityNotFoundException) {
+            snackbar(binding.root, getString(R.string.email_chooser_error), Snackbar.LENGTH_SHORT)
```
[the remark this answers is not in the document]
jpalvarezl: `runCatching` and maybe using `.apply{ }` for the intent?

app/src/main/res/navigation/main_nav.xml
```diff
@@ -6,67 +6,58 @@
...
-            android:id="@+id/action_safeSettingsFragment_to_noSafeFragment"
-            app:destination="@id/noSafeFragment" />
+    <fragment
+        android:id="@+id/advancedAppSettingsFragment"
+        android:name="io.gnosis.safe.ui.safe.settings.app.AdvancedAppSettingsFragment"
+        android:label="AdvancedAppSettingsFragment"
+        tools:layout="@layout/fragment_settings_app_advanced"/>
+
+    <dialog
+        android:id="@+id/safeSelectionDialog"
```
jpalvarezl, line 52: Huh interesting, I didn't know you could have dialogs to a nav graph.

app/src/test/java/io/gnosis/safe/ui/safe/settings/SafeSettingsViewModelTest.kt
```diff
@@ -26,36 +29,41 @@ class SafeSettingsViewModelTest {
...
+            val viewAction = this.viewAction
+            assert (
+                viewAction is ActiveSafe &&
+                        viewAction.safe == null
+            )
         }
 
         coVerify(exactly = 1) { safeRepository.getActiveSafe() }
         coVerify(exactly = 1) { safeRepository.removeSafe(SAFE_1) }
+        coVerify(exactly = 1) { safeRepository.clearActiveSafe() }
```
jpalvarezl, line 66: If you put all the methods within the same `coVerifySequence` block, it would also verify that the methods were called the way you sorted them.

data/src/main/java/io/gnosis/data/repositories/EnsRepository.kt
```diff
@@ -50,7 +50,8 @@ class EnsRepository(
 
 
     companion object {
-        private val ENS_ADDRESS = BuildConfig.ENS_REGISTRY.asEthereumAddress()!!
+
+        val ENS_ADDRESS = BuildConfig.ENS_REGISTRY.asEthereumAddress()!!
```
jpalvarezl, line 54: You still seem to have a this constant defined in the SettingsFragment or somewhere in the UI level if I remember correctly. Is that really necessary?

app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
```diff
@@ -49,8 +48,8 @@ class StartActivity : BaseActivity(), SafeOverviewNavigationHandler {
     private fun isFullscreen(id: Int?): Boolean =
         id != R.id.safeBalancesFragment &&
                 id != R.id.safeTransactionsFragment &&
-                id != R.id.safeSettingsFragment &&
-                id != R.id.noSafeFragment
+                id != R.id.settingsFragment &&
+                id != R.id.safeSelectionDialog
```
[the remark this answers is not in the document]
jpalvarezl: I checked lower after this realising that this is in fact ok it seems. Too bad we have to check for this for the full screen stuff. I think we should probably come back to your initial approach with 2 nav host fragment, one that replaces everything and one for tabbed navigation.

data/src/main/java/io/gnosis/data/repositories/EnsRepository.kt
elgatovital: Just a reference constant AdvancedAppSettingFragment for readability sake:
 val ENS_REGISTRY = EnsRepository.ENS_ADDRESS

app/src/test/java/io/gnosis/safe/ui/safe/settings/SafeSettingsViewModelTest.kt
elgatovital: how do I account for activeSafeFlow() calls? They will make coVerifySequence hard to read

app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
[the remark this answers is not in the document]
elgatovital: safe selection is not modal. You can close it, click back and navigate away to the previous screen. This is actually intended way of using bottom sheets in nav graph

app/src/main/res/values/styles.xml
```diff
@@ -108,6 +108,20 @@
...
+        <item name="android:layout_height">@dimen/header_height</item>
+        <item name="android:background">@color/background</item>
+        <item name="android:gravity">center_vertical</item>
+        <item name="android:paddingLeft">16dp</item>
+        <item name="android:textSize">10sp</item>
+        <item name="android:fontFamily">sans-serif</item>
+        <item name="android:textStyle">bold</item>
+        <item name="android:textColor">@color/medium_grey</item>
+        <item name="android:letterSpacing">0.15</item>
+        <item name="android:lineSpacingExtra">2sp</item>
```
biafra23, line 122: This is probably what we discussed this morning. But why not use         
`<item name="android:textAllCaps">true</item>`here in the style and then have the Strings not necessarily all uppercase?

app/src/main/res/layout/fragment_settings_safe.xml
```diff
@@ -20,9 +21,9 @@
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
         android:layout_marginBottom="16dp"
-        android:text="TextView"
         app:layout_constraintBottom_toTopOf="@+id/button"
         app:layout_constraintEnd_toEndOf="parent"
-        app:layout_constraintStart_toStartOf="parent" />
+        app:layout_constraintStart_toStartOf="parent"
+        tools:text="Safe name" />
```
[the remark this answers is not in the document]
elgatovital: it's just a placeholder layout. Safe settings implementation is about to come

app/src/main/res/drawable/ic_settings_arrow.xml
```diff
@@ -0,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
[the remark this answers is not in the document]
elgatovital: I was not sure. Imo its fine to have small duplication for increased flexibility

app/src/main/res/values/styles.xml
elgatovital: I don't know if all headers are uppercase. Let's move this discussion out of this pr

app/src/main/java/io/gnosis/safe/ui/safe/settings/app/GetInTouchFragment.kt
```diff
@@ -0,0 +1,119 @@
...
+    private fun openDiscord() {
+        startActivity(Intent(Intent.ACTION_VIEW, Uri.parse("https://discord.gg/${getString(R.string.id_discord)}")))
+    }
+
+    private fun openTwitter() {
+        val intent = try {
+            // get the Twitter app if possible
+            requireContext().packageManager.getPackageInfo("com.twitter.android", 0)
+            Intent(Intent.ACTION_VIEW, Uri.parse("twitter://user?screen_name=${getString(R.string.id_twitter)}")).apply {
+                addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
```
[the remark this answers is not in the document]
elgatovital: good point

app/src/main/java/io/gnosis/safe/ui/safe/balances/SafeBalancesFragment.kt
```diff
@@ -80,11 +78,36 @@ class BalancesPagerAdapter(fragment: Fragment) : FragmentStateAdapter(fragment)
...
+            Tabs.COINS -> if (noActiveSafe) 3L else 1L
+            Tabs.COLLECTIBLES -> 2L
+        }
+    }
+
+    override fun containsItem(itemId: Long): Boolean {
+        return when {
+            noActiveSafe && itemId == 3L -> true
+            noActiveSafe && itemId != 1L -> false
+            else -> itemId == 1L || itemId == 2L
```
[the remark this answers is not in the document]
elgatovital: simplified containsItem, but I would keep getItemId as is (I think it has good readability)

app/src/main/java/io/gnosis/safe/ui/safe/balances/SafeBalancesFragment.kt
```diff
@@ -104,10 +100,10 @@ class BalancesPagerAdapter(fragment: Fragment) : FragmentStateAdapter(fragment)
...
 
     override fun containsItem(itemId: Long): Boolean {
-        return when {
-            noActiveSafe && itemId == 3L -> true
-            noActiveSafe && itemId != 1L -> false
-            else -> itemId == 1L || itemId == 2L
+        return when (itemId) {
+            3L -> noActiveSafe
+            1L -> !noActiveSafe
+            else -> true
```
jpalvarezl, line 106: Man this looks confusing :O I have no idea what is going on here at a single glance. Do you think you could use the enum here somehow? I feel like maybe the enum is getting in the way. There is too many magic numbers and magic booleans flying around. I think we either keep the magic numbers and get rid of the enum, or we stick to the enum, but this code looks really messy to me.

app/src/main/java/io/gnosis/safe/ui/safe/settings/app/GetInTouchFragment.kt
elgatovital: asked & removed comment. But I would leave code for opening telegram app in case we need it later

app/src/main/java/io/gnosis/safe/ui/safe/balances/SafeBalancesFragment.kt
elgatovital: enum it is

app/src/main/java/io/gnosis/safe/ui/safe/settings/app/GetInTouchFragment.kt
```diff
@@ -0,0 +1,112 @@
...
+        requireContext().openUrl(getString(R.string.link_help_center))
+    }
+
+    private fun openFeatureSuggestionPage() {
+        requireContext().openUrl(getString(R.string.link_feature_suggestion))
+    }
+
+    private fun openTelegramChannel() {
+        kotlin.runCatching {
+            val intent = Intent(Intent.ACTION_VIEW, Uri.parse("tg://resolve?domain=${getString(R.string.id_telegram)}"))
```
biafra23, line 88: Why are we handling Telegram different compared to Twitter and Discord?

PR #626, opened 2020-05-26
Transaction- and Relay-Service are now configured in data module

biafra23:
For the internal build we do not want to use the production services. We want to be able to use new features, when they are deployed to staging.

Closes # .

- Move config of TRANSACTION_SERVICE_URL & RELAY_SERVICE_URL to data module
- Align debug/internal/rinkeby/release build types wit respective services
- 

@gnosis/mobile-devs

### DIFF
--- a/buildsystem/android_library.gradle
+++ b/buildsystem/android_library.gradle
@@ -17,26 +17,16 @@ android {
 
 
     buildTypes {
-
         debug {
-            buildConfigField javaTypes.STRING, "TRANSACTION_SERVICE_URL", asString(getKey("TRANSACTION_SERVICE_URL", " https://safe-transaction.staging.gnosisdev.com/api/"))
-            buildConfigField javaTypes.STRING, "RELAY_SERVICE_URL", asString(getKey("RELAY_SERVICE_URL", "https:// https://safe-relay.staging.gnosisdev.com/api/"))
         }
 
         internal {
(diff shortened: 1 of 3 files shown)


buildsystem/android_library.gradle
```diff
@@ -19,15 +19,24 @@ android {
     buildTypes {
 
         debug {
```
elgatovital, line 21: these definitions should be directly in data's build.gradle. android_library.gradle is used by all android modules

data/build.gradle
```diff
@@ -10,23 +10,12 @@ android {
     defaultConfig {
         javaCompileOptions {
```
elgatovital, line 12: we can move buildConfigField definitions outside of javaCompileOptions block

data/build.gradle
```diff
@@ -10,23 +10,12 @@ android {
...
-            buildConfigField javaTypes.STRING, "TRANSACTION_SERVICE_URL", asString(getKey("TRANSACTION_SERVICE_URL", "https://safe-transaction.rinkeby.gnosis.io/api/"))
-            buildConfigField javaTypes.STRING, "RELAY_SERVICE_URL", asString(getKey("RELAY_SERVICE_URL", "https://safe-relay.rinkeby.gnosis.pm/api/"))
-
             annotationProcessorOptions {
                 arguments = ["room.schemaLocation": "$projectDir/schemas".toString()]
             }
         }
     }
 
-    buildTypes {
```
elgatovital, line 22: just add other flavors here explicitly and place buildConfigFields with endpoint addresses there

PR #627, opened 2020-05-27
Track screen names

### LINKED ISSUE #599: Tracking: manually track screens
Using nav component disables automatic firebase screen tracking mechanism. 
It is necessary to manually set current screen. See https://firebase.google.com/docs/analytics/screenviews#manually_track_screens

elgatovital:
Closes #599 

- Manually track screens with firebase analytics
- Split BaseFragment into BaseFragment and BaseViewBindingFragment
- Use BaseFragment to track screen id for screen fragments
- Use BaseBottomSheetDialogFragment to track screen id of bottom sheet dialogs
- Use BaseActivity to track screen id for screens represented by activities outside of the main nav graph

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/Tracker.kt
+++ b/app/src/main/java/io/gnosis/safe/Tracker.kt
@@ -1,5 +1,6 @@
 package io.gnosis.safe
 
+import android.app.Activity
 import android.content.Context
 import android.os.Bundle
 import com.google.firebase.analytics.FirebaseAnalytics
@@ -9,9 +10,13 @@ import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
-class Tracker @Inject constructor(@ApplicationContext context: Context) {
+class Tracker private constructor(context: Context) {
 
     private val firebaseAnalytics: FirebaseAnalytics = FirebaseAnalytics.getInstance(context)
+
+    fun setCurrentScreenId(activity: Activity, screenId: ScreenId) {
+        firebaseAnalytics.setCurrentScreen(activity, screenId.value, null)
+    }
 
     fun setNumSafes(numSafes: Int) {
         firebaseAnalytics.setUserProperty(Param.NUM_SAFES, numSafes.toString())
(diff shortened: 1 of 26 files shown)


app/src/main/java/io/gnosis/safe/Tracker.kt
```diff
@@ -57,4 +60,43 @@ class Tracker @Inject constructor(@ApplicationContext context: Context) {
...
+        fun getInstance(context: Context): Tracker =
+            INSTANCE ?: synchronized(this) {
+                INSTANCE ?: Tracker(context).also { INSTANCE = it }
+            }
+    }
+}
+
+enum class ScreenId(val value: String) {
+    SPLASH("splash"),
+    LAUNCH("launch"),
```
biafra23, line 78: @tschubotz changed https://github.com/gnosis/safe/wiki/Overview-screen-tracking (per my request) and we can get rid of splash now and track launch on  the splash screen. If you don't want to do that, I can do it in my PR regarding the SplashScreen.

app/src/main/java/io/gnosis/safe/ui/base/BaseFragment.kt
```diff
@@ -1,51 +1,41 @@
...
 
-    private var _binding: T? = null
-    protected val binding get() = _binding!!
+    @Inject
+    lateinit var tracker: Tracker
 
     override fun onAttach(context: Context) {
         super.onAttach(context)
-        inject(buildViewComponent(context))
+        HeimdallApplication[requireContext()].inject(this)
```
jpalvarezl, line 20: This returns the `appComponent` not the `viewComponent`. I think the `appComponent` is passed to the `viewComponent` when you are building it. I am not sure this necessary to call here.

app/src/main/java/io/gnosis/safe/ui/base/BaseViewBindingFragment.kt
```diff
@@ -0,0 +1,37 @@
...
+abstract class BaseViewBindingFragment<T> : BaseFragment()
+        where T : ViewBinding {
+
+    private var _binding: T? = null
+
+    protected val binding get() = _binding!!
+
+    override fun onAttach(context: Context) {
+        super.onAttach(context)
+        inject(buildViewComponent(context))
```
jpalvarezl, line 19: Injection is not totally decoupled like you mentioned. It seems like you are injecting the fragment still here.

app/src/main/java/io/gnosis/safe/ui/safe/add/AddSafeFragment.kt
```diff
@@ -8,27 +8,30 @@ import android.view.ViewGroup
...
 
-class AddSafeFragment : BaseFragment<FragmentAddSafeBinding>() {
+class AddSafeFragment : BaseViewBindingFragment<FragmentAddSafeBinding>() {
 
     @Inject
     lateinit var viewModel: AddSafeViewModel
 
     private val addressInputHelper by lazy {
-        AddressInputHelper(this, ::updateAddress, allowAddressBook = false)
+        AddressInputHelper(this, tracker, ::updateAddress, allowAddressBook = false)
```
jpalvarezl, line 30: I suppose this shouldn't be necessary if we are going to use the singleton patterns with `Tracker`, you are still using DI here.

app/src/main/java/io/gnosis/safe/ui/base/BaseFragment.kt
elgatovital: tracker is injected from app component

app/src/main/java/io/gnosis/safe/ui/safe/add/AddSafeFragment.kt
elgatovital: Yes, injection is intended. Tracker is singleton, but we still use injection. This way it more convenient to unit test user property tracking

app/src/main/java/io/gnosis/safe/ui/base/BaseBottomSheetDialogFragment.kt
```diff
@@ -39,21 +45,30 @@ abstract class BaseBottomSheetDialogFragment<T : ViewBinding> : BottomSheetDialo
     override fun onCreateDialog(savedInstanceState: Bundle?): Dialog =
         BottomSheetDialog(requireContext(), theme)
 
-    override fun getTheme(): Int {
-        return R.style.BottomSheetDialogTheme
+    override fun onStart() {
+        super.onStart()
+        screenId()?.let {
+            tracker.setCurrentScreenId(requireActivity(), it)
+        }
```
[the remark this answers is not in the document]
elgatovital: no, I decoupled base injection(like tracker that has to be injectied into all sub fragments)

app/src/main/java/io/gnosis/safe/Tracker.kt
```diff
@@ -57,4 +60,42 @@ class Tracker @Inject constructor(@ApplicationContext context: Context) {
...
+
+    companion object {
+
+        @Volatile
+        private var INSTANCE: Tracker? = null
+
+        fun getInstance(context: Context): Tracker =
+            INSTANCE ?: synchronized(this) {
+                INSTANCE ?: Tracker(context).also { INSTANCE = it }
+            }
```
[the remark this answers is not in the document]
elgatovital: it is still being injected by dagger using application context

app/src/main/java/io/gnosis/safe/ui/base/BaseFragment.kt
jpalvarezl: yes, but the fragment receives its dependencies from the `viewComponent` which itself should provide any dependencies from the appComponent automatically as far as I understand. You shouldn't have to call `inject` twice in different components

app/src/main/java/io/gnosis/safe/ui/base/BaseViewBindingFragment.kt
elgatovital: why should it be totally decoupled? I have not mentioned that

elgatovital:
> Looks like DI changed for fragments. I don't think you should be calling `inject` in 2 different components for fragments.

Why not? How do you want to inject tracker to all viewbinding-fragment subclasses?

app/src/main/java/io/gnosis/safe/ui/base/BaseFragment.kt
elgatovital: this is also exactly how we do it for our activities now. Same approach we allready have in our codebase. I don't think this refactoring is in scope of this pr

elgatovital: Also injection should happen exactly where @Inject annotation is specified. So current approach might be the only way. You can't inject specific view models into base class

PR #629, opened 2020-05-27
Splashscreen

### LINKED ISSUE #619: Splashscreen
Screens: 
https://zpl.io/2GJ0163
https://zpl.io/2G0Nnzm

biafra23:
**Draft because of missing tests**

Closes #619 .

- The Button is only shown if the terms are not agreed yet
- The StartActivity is automatically launched if the terms have been agreed

splash-fading

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/splash/SplashActivity.kt
```diff
@@ -35,11 +38,19 @@ class SplashActivity : BaseActivity() {
                         }
                     }.show(supportFragmentManager, TermsBottomSheetDialog::class.simpleName)
                 }
+                is SplashViewModel.ShowButton -> {
+                    binding.continueButton.visibility = View.VISIBLE
```
elgatovital, line 42: you could use visible extension from svalinn: binding.continueButton.visible(true)

app/src/main/res/layout/activity_splash.xml
```diff
@@ -1,22 +1,45 @@
...
+        app:layout_constraintHorizontal_bias="0.0"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@+id/safe_multisig_logo" />
 
     <TextView
         android:id="@+id/continue_button"
         style="@style/PrimaryButton"
         android:layout_width="0dp"
-        android:layout_marginEnd="@dimen/default_horizontal_margin"
+        android:layout_height="56dp"
```
elgatovital, line 35: PrimaryButton style already has layout_height defined and it has same value - 56dp. I think it might be better to move this to a dimension and use it in both places. Otherwise, it gives impression like you are overriding defaults of the primary button style

app/src/main/res/values/styles.xml
```diff
@@ -11,18 +11,12 @@
...
-    <style name="MountainTheme" parent="MountainThemeBase" />
-
     <style name="WhiteThemeBase" parent="AppTheme" />
 
     <style name="WhiteTheme" parent="WhiteThemeBase" />
 
     <style name="PrimaryButton">
-        <item name="android:layout_height">56dp</item>
+        <item name="android:layout_height">@dimen/primary_button_height</item>
         <item name="android:background">@drawable/primary_button</item>
```
elgatovital, line 20: background is already defined in style for PrimaryButton

biafra23: ? 
This is the style for primary button.

elgatovital: my bad, sorry

PR #630, opened 2020-05-28
Clean up whitespaces and imports according to default ktlint settings

biafra23:
Closes # .

- I ran ktlint -F and this is the result
- 
- 

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/base/Adapter.kt
```diff
@@ -59,4 +59,3 @@ abstract class Adapter<T, VH : Adapter.ViewHolder<T>> : RecyclerView.Adapter<VH>
 }
 
 class UnsupportedViewType(message: String? = null) : Throwable(message)
-
```
biafra23, line 62: Don't worry, there is still a new line at the end of the file. This removes an additional newline (there were two).

PR #650, opened 2020-06-11
TX repository test for polymorphic adapter

biafra23:
This merges back to feature/tx_repository

- Add tests for PolymorphicJsonAdapterFactory configuration
- move PolymorphicJsonAdapterFactory creation to data module


data/src/test/java/io/gnosis/data/models/TransactionDtoTest.kt
```diff
@@ -0,0 +1,76 @@
...
+import java.util.stream.Collectors
+
+class TransactionDtoTest {
+
+    private var moshi = MoshiBuilderFactory.makeMoshiBuilder()
+        .add(BigDecimalNumberAdapter())
+        .add(OperationEnumAdapter())
+        .add(transactionDtoJsonAdapterFactory)
+        .add(KotlinJsonAdapterFactory())
+        .build()
```
jpalvarezl, line 22: This will definitely fall out of sync with the one we use in the `ApplicationModule` I would say we push down the provides method for dagger into `:data` so we use the same object as we do in runtime.

PR #652, opened 2020-06-15
Enter safe screen design compliant

jpalvarezl:
- "NEXT" button is disabled upon entering the screen
- Added color state list for text buttons and made them all caps
- Adjusted font styles and margins to match that of the designs

| Before | After |
|-------|-------|
|image|image|

@gnosis/mobile-devs


biafra23:
What about the next screen (Enter_name https://zpl.io/bPJ516p)? In the designs it seems the naming screen also should have the next button grey and disabled.

jpalvarezl:
@biafra23 There is a different issue for that one. I have assigned myself to that one too :)

PR #654, opened 2020-06-15
Safe name screen fixes to match design

jpalvarezl:
- Restored button for terms (casing was not all caps and style was used) 
- Toolbar to look as in the mocks
- NEXT button behaviour implemented

@gnosis/mobile-devs


biafra23:
The boldness in the address formatting seems bolder in the Zeplin layout. I guess we're using this everywhere?

image

image

jpalvarezl:
Yeah, I am using `TextDark.Bold` like everywhere else.

jpalvarezl:
You can see that change in line 97 in this file: `app/src/main/res/layout/fragment_add_safe_name.xml`

biafra23:
> You can see that change in line 97 in this file: `app/src/main/res/layout/fragment_add_safe_name.xml`

I meant the first byte and the last two bytes of the Ethereum address. Not the text below it.

jpalvarezl:
Nah man you were absolutely right :) the span we used was just switching the colour to black without touching the typeface at all. Now I am also bolding the text :) well spotted.

jpalvarezl:
Looks like this now:
Screenshot_1592224423

Issue #645, opened 2020-06-10
[Android]Address field validation without pressing "Next" 


liliya-soroka:
**Is your feature request related to a problem? Please describe.**
To validate entered ( via QR code, ENS name, or from clipboard )safe address the user has to click Next. From my point of view, it's not very user-friendly as an additional step required.

**Describe the solution you'd like**
The entered address should be validated "on the fly" as soon as it's added to the "Enter safe address" field without an additional click on Next.
Next link is enabled when the address is not valid.
The same flow as in the ios version - https://drive.google.com/open?id=1jAxSwn3taQ20DOttL9C0C3xDdUB_dPJp

liliya-soroka:
Verified
app version 1.12.1-35

Issue #643, opened 2020-06-10
[Android]Enter Safe Address menu view 


liliya-soroka:
Questions:
1. Is the title on design is correct for the menu? Could we leave implemented title?
2.  The order of the items in the menu should be changed to the order of the items on the design, right?

### Expected behavior
https://app.zeplin.io/project/5e9efc30726e106958cc43de/screen/5ea803ae22315c2586ffe2be
1. Keep title as implemented 
2. Order should be according to the designs, from most used to less used

### Actual behavior
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/7bd08021-55aa-4d0c-9872-07208f8ec525)

### Steps to reproduce the behavior
1. Open Enter address dialog

### Additional information

- Device: Huawei p30
- Android OS version: 10
- Safe app version: 1.12.1-27

posthnikova:
@liliya-soroka
1. The title 'Choose input method' sounds better, let's keep it
2. Order should be according to the designs, from most used to less used

liliya-soroka:
verified.
app version 1.12.1-35

Issue #642, opened 2020-06-10
[Android]Bottom bar: rename Balance to Assets


liliya-soroka:
**Describe the bug**
The wrong name of the Assets tab is used in the bottom bar -https://app.zeplin.io/project/5e9efc30726e106958cc43de/screen/5ea96964d41f4f24bbc87dbd

**To Reproduce**
1. Open app
2. look at the bottom bar

**Expected behavior**
Assets  name should be used instead of Balance in the bottom bar

**Screenshots**
design -https://app.zeplin.io/project/5e9efc30726e106958cc43de/screen/5ea96964d41f4f24bbc87dbd

**Smartphone (please complete the following information):**
 - Device: Huawei
 - Android Version: 10
 - App Version 1.12.1-27

liliya-soroka:
verified
1.12.1 - 35

Issue #609, opened 2020-05-20
Localised numbers


jpalvarezl:
Numbers should be localised according to the devices locale.

For now only the balances screen

liliya-soroka:
verified
app version 1.12.1-35

PR #667, opened 2020-06-17
Various fixes to Balances and add safe flow

jpalvarezl:
References:
https://github.com/gnosis/safe-android/issues/645
https://github.com/gnosis/safe-android/issues/643
https://github.com/gnosis/safe-android/issues/670
https://github.com/gnosis/safe-android/issues/609
https://github.com/gnosis/safe-android/issues/642
https://github.com/gnosis/safe-android/issues/669

- Rearranged menu items for address input
- Added on input validation of address
- Having 2 decimals for fiat currency representation of balance
- Locale aware decimal punctuation
- Renamed `Balances` to `Assets` (in class names too) 
- Added the correct placeholder for coins

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/di/components/ViewComponent.kt
+++ b/app/src/main/java/io/gnosis/safe/di/components/ViewComponent.kt
@@ -6,9 +6,9 @@ import io.gnosis.safe.di.modules.ViewModule
 import io.gnosis.safe.ui.dialogs.EnsInputDialog
 import io.gnosis.safe.ui.safe.add.AddSafeFragment
 import io.gnosis.safe.ui.safe.add.AddSafeNameFragment
-import io.gnosis.safe.ui.safe.balances.SafeBalancesFragment
-import io.gnosis.safe.ui.safe.balances.coins.CoinsFragment
-import io.gnosis.safe.ui.safe.balances.collectibles.CollectiblesFragment
+import io.gnosis.safe.ui.safe.assets.AssetssFragment
+import io.gnosis.safe.ui.safe.assets.coins.CoinsFragment
+import io.gnosis.safe.ui.safe.assets.collectibles.CollectiblesFragment
 import io.gnosis.safe.ui.safe.selection.SafeSelectionDialog
 import io.gnosis.safe.ui.safe.settings.SettingsFragment
 import io.gnosis.safe.ui.safe.settings.app.AdvancedAppSettingsFragment
@@ -48,7 +48,7 @@ interface ViewComponent {
 
     fun inject(fragment: SafeSettingsFragment)
 
-    fun inject(fragment: SafeBalancesFragment)
+    fun inject(fragment: AssetssFragment)
 
     fun inject(fragment: CoinsFragment)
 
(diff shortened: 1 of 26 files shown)


app/src/main/java/io/gnosis/safe/ui/safe/balances/coins/CoinBalanceAdapter.kt
```diff
@@ -33,9 +35,13 @@ class CoinBalanceViewHolder(private val viewBinding: ItemCoinBalanceBinding) : R
     fun bind(balanceItem: Balance) {
         with(viewBinding) {
             logoImage.loadTokenLogo(balanceItem.token.logoUrl)
+            val formatter = DecimalFormat.getInstance()
             symbol.text = balanceItem.token.symbol
-            balance.text = balanceItem.balance.shiftedString(balanceItem.token.decimals, 5)
-            balanceUsd.text = viewBinding.root.context.getString(R.string.usd_balance, balanceItem.balanceUsd)
+            balance.text = formatter.format(balanceItem.balance.shifted(balanceItem.token.decimals))
```
elgatovital, line 40: this could be extracted to utility method

app/src/main/java/io/gnosis/safe/di/components/ViewComponent.kt
```diff
@@ -48,7 +48,7 @@ interface ViewComponent {
 
     fun inject(fragment: SafeSettingsFragment)
 
-    fun inject(fragment: SafeBalancesFragment)
+    fun inject(fragment: AssetssFragment)
```
elgatovital, line 51: AssetsFragment

app/src/main/java/io/gnosis/safe/ui/safe/balances/coins/CoinBalanceAdapter.kt
jpalvarezl: Not exactly sure how do you mean that. `shifted` is already a utility method. Just passing its output to `format`

elgatovital: the formatting part

jpalvarezl: Can you provide a concrete suggestion? I don't see how I can meaningfully improve this by extracting a method. Then is more lines of code and passing the formatter around, if that is what you are suggesting.

elgatovital: I suggest having formatting logic in one place. We could have formatting helper and inject it into adapter
```class BalanceFormatter(
val context: Context,
val formatter: DecimalFormatter
) {
	fun formatCurrency(balance: BigInteger, decimals: Int): String {
		return formatter.format(balance.shifted(decimals)
	}

	fun formatUsd(balanceUsd: BigDecimal): String {
	        return context.getString(R.string.usd_balance, formatter.fromat(balanceUsd.setScale(2, RoundingMode.DOWN))
	}
}```
This way currency formatting is encapsulated in one place and can be easily adjusted.

elgatovital: can be done later though. Currently, balances are formatted only here

jpalvarezl: That would definitely be the correct approach once we have to support various fiat currencies.

Issue #622, opened 2020-05-25
Tx List: Create Tx Repository


elgatovital:
Transaction network access needs to be defined according to https://github.com/gnosis/safe/issues/324

Added network access to the `all-transactions` endpoint.

Define the types like in the issue mentioned above.

Define unit tests for deserialisers using polymorphism. Unknown new types should be filtered the item out instead of throwing and crashing.

Map the 3 backend types (to this date) Multisig, Module, Ethereum transactions into Transfer, Settings change and Custom transactions. Provide unit tests asserting which fields from the DTOs map to the domain classes.

jpalvarezl:
Transaction type heuristic:

- Settings change -> transaction.to  and safe addresses are identical
- Transfer -> transaction.data  starts with 0xa9059cbb
- Custom transaction -> The rest

`Transfers` within a `Transaction` will contain the following class instead of `tokenAddress`

https://github.com/gnosis/safe-transaction-service/issues/96

`/safes/{address}/all-transactions` returns only `history` section of the whole list ( `nonce < currentSafeNonce`)

`/safes/{address}/transactions` returns only `queued` section of the whole list (`nonce > currentSafeNonce`)

Tx Sorting: https://github.com/gnosis/safe/wiki/How-should-transactions-be-sorted

Tx States: https://github.com/gnosis/safe/wiki/Safe-transaction-states

jpalvarezl:
https://github.com/gnosis/safe-transaction-service/issues/97

biafra23:
Please see: https://github.com/gnosis/safe/wiki/How-do-we-determine-the-tx-type for help on how to distinguish between the tx types.

jpalvarezl:
Looking good! Nice foundation!

Issue #663, opened 2020-06-16
Binding domain Transaction objects to their ViewHolders


jpalvarezl:
**Describe the solution you'd like**
We have already in place the layouts and the logic for allowing a multiple `ViewHolder` `RecyclerView.Adapter`. This task assumes that the direction of the transaction `Incoming/Outgoing` and its status `{Pending, AwaitingConfirmation, AwaitingExecution, Cancelled, Successful, Failed}` is known. Also, the correct domain `Transaction` type is supplied to the bind method `Transfer, Custom, SettingsChange`.

Bind the fields of the classes to their respective view.

`TransactionsViewModel` should map from all the subclasses of `Transaction` to all the subclasses of `TransactionView`

liliya-soroka:
Verified
app version 1.12.1 - 54

PR #668, opened 2020-06-17
TX Repository

biafra23:
Regarding #622 :  .

- Create a repository for several types of transfers by fetching a list from all-transactions endpoint
- Decide what kind of tx and create Domain objects of : `Transaction` (`Custom`, `SettingsChange` and `Transfer`) 
- Tests have been addded

Missing: 
- Proper extraction of to/from/value for ERC721 tokens as described here: https://github.com/gnosis/safe/issues/324
- Fallback if contractInfo does not become available soon (is also missing)

@gnosis/mobile-devs


data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
```diff
@@ -0,0 +1,180 @@
...
+        return Transaction.Transfer(
+            transferDto.to,
+            transferDto.from,
+            transferDto.value,
+            transferDto.executionDate?.formatBackendDate(),
+            transferDto.type.let {
+                when (it) {
+                    TransferType.ERC20_TRANSFER -> TokenRepository.FAKE_ERC20_TOKEN_INFO
+                    TransferType.ERC721_TRANSFER -> TokenRepository.FAKE_ERC721_TOKEN_INFO
+                    else -> TokenRepository.ETH_SERVICE_TOKEN_INFO
```
jpalvarezl, line 73: ah, ok. I see why you had to leave those mock values outside of tests. If you don't mind, I would prefer moving those over to this class, so we have all the stuff that needs removing or mocking in the same place, that is the `TransactionRepository` and then when the backend is ready we just check one place.

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
```diff
@@ -0,0 +1,180 @@
...
+    this?.map {
+        if (it.name == name) {
+            return it.value
+        }
+    }
+    return null
+}
+
+fun String.dataSizeBytes(): Long = removeHexPrefix().hexToByteArray().size.toLong()
+fun String?.hexStringNullOrEmpty(): Boolean = this?.dataSizeBytes() ?: 0L == 0L
```
jpalvarezl, line 180: I think that we must have a method for this in svalinn. Also, this check could be simplified to `this?.size > 2 == true`, where two accounts for `0x` but definitely have a look `:utils` in svalinn.

data/src/test/java/io/gnosis/data/repositories/TransactionRepositoryTest.kt
```diff
@@ -0,0 +1,391 @@
...
+    fun `getTransactions (module transaction) should return custom`() = runBlockingTest {
+        val transactionDto = buildModuleTransactionDto(module = Solidity.Address(BigInteger.ONE))
+        val pagedResult = listOf(transactionDto)
+        coEvery { transactionServiceApi.loadTransactions(any()) } returns Page(1, null, null, pagedResult)
+
+        val actual = transactionRepository.getTransactions(defaultSafeAddress)
+
+        with(actual.results[0] as Transaction.Custom) {
+            assertEquals(transactionDto.module, address)
+            assertEquals(transactionDto.data?.dataSizeBytes() ?: 0L, dataSize)
```
jpalvarezl, line 54: Why is it necessary to have a default value for these checks? I think that if something is null, it should be null also in the other side of the mapping.

jpalvarezl: Or hardcode the default value in the repository so we get more clarity what the default value is in the tests.

data/src/test/java/io/gnosis/data/repositories/TransactionRepositoryTest.kt
```diff
@@ -0,0 +1,391 @@
...
+    fun `getTransactions (UnknownTransaction with data) should return custom`() = runBlockingTest {
+        val transactionDto = UnknownTransactionDto
+        val pagedResult = listOf(transactionDto)
+        coEvery { transactionServiceApi.loadTransactions(any()) } returns Page(1, null, null, pagedResult)
+
+        val actual = transactionRepository.getTransactions(defaultSafeAddress)
+
+        assertEquals(1, actual.results.size)
+        with(actual.results[0] as Transaction.Custom) {
+            assertEquals(transactionDto.to, address) //TODO: Only to available in UnknownTransaction, but is it useful?
```
jpalvarezl, line 70: There should be enough fields to be able to render a `Custom` transaction.

data/src/test/java/io/gnosis/data/repositories/TransactionRepositoryTest.kt
```diff
@@ -0,0 +1,391 @@
...
+            assertEquals(transactionDto.data?.dataSizeBytes() ?: 0L, dataSize)
```
jpalvarezl, line 71: Let's either hard code the default value from `TransactionRepository` or get rid of the `?:`

data/src/test/java/io/gnosis/data/repositories/TransactionRepositoryTest.kt
```diff
@@ -0,0 +1,391 @@
...
+    private fun buildTransferDto(
+        type: TransferType = TransferType.ETHER_TRANSFER
+    ): TransferDto =
+        TransferDto(
+            to = defaultToAddress,
+            from = defaultFromAddress,
+            type = type,
+            value = BigInteger.ONE
+        )
+}
```
jpalvarezl, line 391: Generally, look really good, man. I am just missing some `coVerify { transactionApi.loadTransactions(address) }` calls. Although it's only one method under the hood for now but it could change in the future.

data/src/main/java/io/gnosis/data/adapters/OperationEnumAdapter.kt
```diff
@@ -0,0 +1,19 @@
+package io.gnosis.data.adapters
+
+import com.squareup.moshi.FromJson
+import com.squareup.moshi.JsonDataException
+import com.squareup.moshi.ToJson
+import io.gnosis.data.models.Operation
+
+class OperationEnumAdapter {
```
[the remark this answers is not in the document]
biafra23: The EnumJsonAdapter doesn't work in this case because it can't handle id values instead of enum names. The JSON contains 0,1 and not CALL/DELEGATE

data/src/main/java/io/gnosis/data/models/Transaction.kt
```diff
@@ -1,11 +1,157 @@
...
+        val date: String?,
+        val nonce: BigInteger
+    ) : Transaction()
+
+    data class Transfer(
+        val recipient: Solidity.Address,
+        val sender: Solidity.Address,
+        val value: BigInteger,
+        val date: String?,
+        val tokenInfo: ServiceTokenInfo
```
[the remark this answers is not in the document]
biafra23: Lets please have that in the next PR. Then we maybe also have the ContractInfo and know how to map it.

data/src/main/java/io/gnosis/data/models/Transaction.kt
```diff
@@ -1,11 +1,157 @@
...
+    data class Custom(
+        val nonce: BigInteger?,
+        val address: Solidity.Address,
+        val dataSize: Long,
+        val date: String?,
+        val value: BigInteger
+    ) : Transaction()
+
+    data class SettingsChange(
+        val dataDecoded: DataDecodedDto,
```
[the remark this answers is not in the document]
biafra23: For the List, yes we only need the name. But when we show details we also need the threshold and probably other data. Should we update this when we do the tx details?

data/src/main/java/io/gnosis/data/repositories/TokenRepository.kt
```diff
@@ -23,7 +25,12 @@ class TokenRepository(
...
 
     companion object {
-        val ETH_ADDRESS = Solidity.Address(BigInteger.ZERO)
+        private val defaultErc20Address = "0xc778417e063141139fce010982780140aa0cd5ab".asEthereumAddress()!!
+        private val defaultErc721Address = "0xB3775fB83F7D12A36E0475aBdD1FCA35c091efBe".asEthereumAddress()!!
+        private val ETH_ADDRESS = Solidity.Address(BigInteger.ZERO)
         val ETH_TOKEN_INFO = Erc20Token(ETH_ADDRESS, "Ether", "ETH", 18, "local::ethereum")
+        val ETH_SERVICE_TOKEN_INFO = ServiceTokenInfo(ETH_ADDRESS, 18, "ETH", "Ether", "local::ethereum")
+        val FAKE_ERC20_TOKEN_INFO = ServiceTokenInfo(defaultErc20Address, 18, "WETH", "Wrapped Ether", "local::ethereum")
+        val FAKE_ERC721_TOKEN_INFO = ServiceTokenInfo(defaultErc721Address, 18, "DRK", "Dirk", "local::ethereum")
```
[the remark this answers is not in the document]
biafra23: If they are only in the test, then they cannot be set in the Repository. We will replace them ASAP.

data/src/main/java/io/gnosis/data/models/Transaction.kt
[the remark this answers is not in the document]
jpalvarezl: Agreed, let's just not forget to create the issue for this.

data/src/main/java/io/gnosis/data/models/Transaction.kt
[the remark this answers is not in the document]
jpalvarezl: Sounds good, let's update when we strictly need 👍

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
biafra23: I did not find a function (String? -> Boolean) in svalin:util. Also ">" doesn't work because of the nullability. I don't think there is a simpler way to write this. There is however a more readable way: 
```
fun String?.hexStringNullOrEmpty(): Boolean = when {
    this == null -> true
    length > 2 -> false
    else -> true
}
```

data/src/test/java/io/gnosis/data/repositories/TransactionRepositoryTest.kt
biafra23: Custom.dataSize() currently can't be null. Because it is of type Long. Also it is computed and not in the dto. I would prefer *not* to make dataSize nullable. 

The default is necessary because data can be null. But IMHO dataSize should be 0 then, and not null.

data/src/test/java/io/gnosis/data/repositories/TransactionRepositoryTest.kt
biafra23: Making data "0x" in all subclasses of Transaction doesn't reflect reality. The Backend can and will still return `"data": null`. I think the tests should be aware of that.

data/src/test/java/io/gnosis/data/repositories/TransactionRepositoryTest.kt
biafra23: I removed the comment

data/src/test/java/io/gnosis/data/repositories/TransactionRepositoryTest.kt
biafra23: I added those

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionViewHolderFactory.kt
```diff
@@ -38,64 +41,81 @@ class TransactionViewHolderFactory : BaseFactory<BaseTransactionViewHolder<Trans
...
-
-    override fun bind(data: Transaction.Transfer, payloads: List<Any>) {
-        TODO("Not yet implemented")
+class TransferViewHolder(private val viewBinding: ItemTxTransferBinding) :
+    BaseTransactionViewHolder<TransactionView.Transfer>(viewBinding) {
+
+    override fun bind(data: TransactionView.Transfer, payloads: List<Any>) {
+        with(viewBinding) {
+            val value = data.transaction.value
+            amount.text = "${value.shiftedString(18)} ${data.transaction.tokenInfo?.symbol}"
```
elgatovital, line 84: we should extract this 18 to some constant with descriptive name

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionViewHolderFactory.kt
```diff
@@ -38,64 +41,81 @@ class TransactionViewHolderFactory : BaseFactory<BaseTransactionViewHolder<Trans
...
+            dateTime.text = data.transaction.date
+            if (data.isIncoming) {
+                txTypeIcon.setImageResource(R.drawable.ic_arrow_green_16dp)
+                blockies.setAddress(data.transaction.sender)
+                ellipsizedAddress.text = data.transaction.sender.asEthereumAddressString().asMiddleEllipsized(4)
+            } else {
+                txTypeIcon.setImageResource(R.drawable.ic_arrow_red_10dp)
+                blockies.setAddress(data.transaction.recipient)
+                ellipsizedAddress.text = data.transaction.recipient.asEthereumAddressString().asMiddleEllipsized(4)
```
elgatovital, line 93: maybe it is a good idea to extract this ethereum address formatting to a utility method, so that formatting can be edited in one place

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
```diff
@@ -1,14 +1,49 @@
...
+@Inject constructor(
+    private val transactionRepository: TransactionRepository,
+    private val safeRepository: SafeRepository,
+    appDispatchers: AppDispatchers
+) : BaseStateViewModel<TransactionsViewState>(appDispatchers) {
 
-    override fun initialState(): TransactionsViewState = TransactionsViewState(null, false)
+    override fun initialState(): TransactionsViewState = TransactionsViewState(null, true)
+
+    fun load() {//safeAddress: Solidity.Address) {
```
elgatovital, line 19: can we delete this comment?

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionViewHolderFactory.kt
biafra23: I extracted it as `formatForTxList()`. But where should I put it?

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
jpalvarezl: sure, you are right, nullability is not addressed. I would suggest then `his?.size ?: 0 > 2 == true` but this is really unimportant.

data/src/test/java/io/gnosis/data/repositories/TransactionRepositoryTest.kt
jpalvarezl: yeah but for the assertion I mean. You shouldn't have `?:` in the values you are asserting.

data/src/test/java/io/gnosis/data/repositories/TransactionRepositoryTest.kt
jpalvarezl: Once again, then your call to assert is not valid if you have an `?:` operator changing the values to an. arbitrary default in the test.

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionViewHolderFactory.kt
biafra23: We probably want a more comprehensive formatter here anyway. Should I mark this with a TODO? To fix this when we have the tokeninfo needed to decide how far to shift? Or is that the same for every token and Ether?

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionViewHolderFactory.kt
elgatovital: io.gnosis.safe.utils.StringUtils?

biafra23: Fixed in: d43eacfa613733d0eb13f09c533224ce0a05a8d9

data/src/main/java/io/gnosis/data/utils/DateUtils.kt
```diff
@@ -0,0 +1,3 @@
+package io.gnosis.data.utils
+
+fun String.formatBackendDate() = this.replace("Z", "").replace("T", ", ").split(".").first()
```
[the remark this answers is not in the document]
biafra23: I agree that this is hacky. But it wasn't introduced in this PR, I'd prefer to fix this in a separate PR. 
I expect some bigger discussions on how to solve this.

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
```diff
@@ -0,0 +1,191 @@
...
+import pm.gnosis.utils.decimalAsBigInteger
+import pm.gnosis.utils.hexToByteArray
+import pm.gnosis.utils.removeHexPrefix
+import java.math.BigInteger
+
+//TODO Remove all this, when we have ContractInfo from the backend
+private val defaultErc20Address = "0xc778417e063141139fce010982780140aa0cd5ab".asEthereumAddress()!!
+private val defaultErc721Address = "0xB3775fB83F7D12A36E0475aBdD1FCA35c091efBe".asEthereumAddress()!!
+private val ETH_ADDRESS = Solidity.Address(BigInteger.ZERO)
+val ETH_TOKEN_INFO = Erc20Token(ETH_ADDRESS, "Ether", "ETH", 18, "local::ethereum")
```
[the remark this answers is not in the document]
biafra23: ETH_ADDRESS and ETH_TOKEN_INFO have been moved (back) to TokenRepository

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
```diff
@@ -0,0 +1,191 @@
...
+val ETH_SERVICE_TOKEN_INFO = ServiceTokenInfo(ETH_ADDRESS, 18, "ETH", "Ether", "local::ethereum")
+val FAKE_ERC20_TOKEN_INFO = ServiceTokenInfo(defaultErc20Address, 18, "WETH", "Wrapped Ether", "local::ethereum")
+val FAKE_ERC721_TOKEN_INFO = ServiceTokenInfo(defaultErc721Address, 18, "DRK", "Dirk", "local::ethereum")
```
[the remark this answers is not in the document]
biafra23: To be able to check for it in a test it needs be here. The code tested does not depend on the test. So if this is only in the test, then it can't be used in the SUT.

data/src/main/java/io/gnosis/data/utils/DateUtils.kt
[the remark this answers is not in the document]
jpalvarezl: Yes, this is beyond the scope of this issue. The date formatting of the current backend is something that hasn't change since the prototyping days. I will try to push for this in the rust gateway service repo.

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
```diff
@@ -0,0 +1,191 @@
...
+        val from = transaction.dataDecoded?.parameters?.getValueByName("from")?.asEthereumAddress()
+
+        return Transaction.Transfer(
+            to,
+            from ?: transaction.safe,
+            value,
+            transaction.executionDate?.formatBackendDate()
+                ?: transaction.submissionDate?.formatBackendDate()
+                ?: transaction.modified?.formatBackendDate(),
+            FAKE_ERC20_TOKEN_INFO // TODO: find out correct token data source
```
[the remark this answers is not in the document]
biafra23: Do you want to remove the FAKE_ERC20_TOKEN_INFO and FAKE_ERC721_TOKEN_INFO completely (from tests and here)?

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionViewHolderFactory.kt
biafra23: So I can use `data.transaction.tokenInfo.decimals` instead of 18 here?

data/src/main/java/io/gnosis/data/models/Transaction.kt
```diff
@@ -1,11 +1,157 @@
...
+    ) : Transaction()
+}
+
+interface TransactionDto {
+    val txType: TransactionType?
+    val to: Solidity.Address
+    val data: String?
+}
+
+data class MultisigTransactionDto(
```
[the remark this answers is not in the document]
biafra23: So all Dtos go to the dto package. And where do the enums go?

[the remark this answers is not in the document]
elgatovital: depends whether those are dto or model enums. If they are used in dto -> then dto. Otherwise models. We should try to keep our dto's & models separate

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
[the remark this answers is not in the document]
elgatovital: I wanted to remove it only from the repository. I was unaware we are still missing token data

data/src/main/java/io/gnosis/data/utils/DateUtils.kt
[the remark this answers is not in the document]
elgatovital: please open an issue for this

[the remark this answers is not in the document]
biafra23: https://github.com/gnosis/safe-android/issues/673

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionViewHolderFactory.kt
jpalvarezl: This should come somehow from your lower layer of mapping. In your `TransactionView.Transfer`. I would also say that this is a bit outside scope as this should be addressed here: https://github.com/gnosis/safe-android/issues/663

jpalvarezl: Just let Liliya know that the UI she might see out of transactions right now is not valid and shouldn't be tested until #663 is implemented.

data/src/main/java/io/gnosis/data/models/Transaction.kt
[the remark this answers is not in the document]
biafra23: Done in a9accbb006d6fba824b8d4ebea9fd6272fb899ea

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
```diff
@@ -0,0 +1,191 @@
...
+import pm.gnosis.utils.removeHexPrefix
+import java.math.BigInteger
+
+//TODO Remove all this, when we have ContractInfo from the backend
+private val defaultErc20Address = "0xc778417e063141139fce010982780140aa0cd5ab".asEthereumAddress()!!
+private val defaultErc721Address = "0xB3775fB83F7D12A36E0475aBdD1FCA35c091efBe".asEthereumAddress()!!
+private val ETH_ADDRESS = Solidity.Address(BigInteger.ZERO)
+val ETH_TOKEN_INFO = Erc20Token(ETH_ADDRESS, "Ether", "ETH", 18, "local::ethereum")
+val ETH_SERVICE_TOKEN_INFO = ServiceTokenInfo(ETH_ADDRESS, 18, "ETH", "Ether", "local::ethereum")
+val FAKE_ERC20_TOKEN_INFO = ServiceTokenInfo(defaultErc20Address, 18, "WETH", "Wrapped Ether", "local::ethereum")
```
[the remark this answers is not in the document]
biafra23: Can we do that, when we get it from the backend?

PR #671, opened 2020-06-18
Force day mode

elgatovital:
Force day mode until dark mode is properly implemented

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -45,6 +45,7 @@
 
         <activity
             android:name=".ui.StartActivity"
+            android:launchMode="singleInstance"
             android:screenOrientation="portrait"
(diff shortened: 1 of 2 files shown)


jpalvarezl:
Wouldn't it make more sense to just change the parent style in `res/values/styles` to simply `Theme.MaterialComponents.NoActionBar`? Would that do the trick?

elgatovital:
> Theme.MaterialComponents.NoActionBar

this might have side-effects on layouts & styles

elgatovital:
> Can you provide an example of an unexpected side effect? I clicked through the app changing the parent theme to `Theme.MaterialComponents.Light.NoActionBar` and looks fine.

It might work now, but switching themes might affect ui components. With current fix we can be sure ui will remain unaffected

PR #672, opened 2020-06-22
Track screen events

elgatovital:
Track screens using firebase events

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/Tracker.kt
+++ b/app/src/main/java/io/gnosis/safe/Tracker.kt
@@ -12,16 +12,16 @@ class Tracker private constructor(context: Context) {
 
     private val firebaseAnalytics: FirebaseAnalytics = FirebaseAnalytics.getInstance(context)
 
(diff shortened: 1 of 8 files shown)


app/src/main/java/io/gnosis/safe/Tracker.kt
```diff
@@ -12,10 +12,6 @@ class Tracker private constructor(context: Context) {
 
     private val firebaseAnalytics: FirebaseAnalytics = FirebaseAnalytics.getInstance(context)
 
-    fun setCurrentScreenId(activity: Activity, screenId: ScreenId) {
-        firebaseAnalytics.setCurrentScreen(activity, screenId.value, null)
```
jpalvarezl, line 16: So we are treating screens as if they were any other event?

elgatovital: yes, because screen tracking doesn't work on iOS. We are restricted to 500 events (hopefully this will be enough)

PR #675, opened 2020-06-22
QR Scan fix

jpalvarezl:
Reference # 646 .

- Added retry logic if something that is not an Ethereum address is scanned
- Added alert dialog notifying the user of the error

@gnosis/mobile-devs

### DIFF
--- a/qrscanner/src/main/java/io/gnosis/safe/qrscanner/QRCodeScanActivity.kt
+++ b/qrscanner/src/main/java/io/gnosis/safe/qrscanner/QRCodeScanActivity.kt
@@ -55,9 +55,9 @@ class QRCodeScanActivity : AppCompatActivity() {
             finish()
         } else {
             AlertDialog.Builder(this)
-                .setTitle(R.string.error_title)
-                .setMessage(R.string.error_message)
-                .setNegativeButton(R.string.retry_button) { _, _ -> videographer.open(scan_view_finder) }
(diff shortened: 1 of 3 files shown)


qrscanner/src/main/res/values/string.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="error_title">Error</string>
```
elgatovital, line 3: better to prefix string names with qr_ to avoid collisions & overriding

qrscanner/src/main/java/io/gnosis/safe/qrscanner/QRCodeScanActivity.kt
```diff
@@ -46,11 +49,22 @@ class QRCodeScanActivity : AppCompatActivity() {
...
     private fun finishWithResult(value: String) {
-        val result = Intent().apply { putExtra(RESULT_EXTRA, value) }
-        setResult(Activity.RESULT_OK, result)
-        finish()
+        if (isEthereumAddress(value)) {
+            val result = Intent().apply { putExtra(RESULT_EXTRA, value) }
+            setResult(Activity.RESULT_OK, result)
+            finish()
+        } else {
+            AlertDialog.Builder(this)
```
elgatovital, line 57: can we just show snackbar and continue to capture screen automatically after short delay?

PR #676, opened 2020-06-22
tx repository handling erc721 properties

biafra23:
Relates to #622.

- Detect (workaround without contractInfo) and copy ERC721 token properties

@gnosis/mobile-devs


data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
```diff
@@ -121,14 +121,19 @@ class TransactionRepository(
         )
     }
 
-    private fun transferErc721(transaction: MultisigTransactionDto): Transaction.Transfer =
-        Transaction.Transfer(
-            transaction.to,
-            transaction.safe,
-            transaction.value,
+    private fun transferErc721(transaction: MultisigTransactionDto): Transaction.Transfer {
+        val from = transaction.dataDecoded?.parameters?.getValueByName("from")?.asEthereumAddress() ?: Solidity.Address(BigInteger.ZERO)
```
elgatovital, line 125: when is it possible for erc721 transfer to have `from` or `to` values set to zero?

data/src/test/java/io/gnosis/data/repositories/TransactionRepositoryTest.kt
```diff
@@ -155,7 +156,14 @@ class TransactionRepositoryTest {
     fun `getTransactions (multisig transaction for ERC721) should return transfer`() = runBlockingTest {
         val transactionDto = buildMultisigTransactionDto(
             contractInfoType = ContractInfoType.ERC721,
-            dataDecodedDto = DataDecodedDto("safeTransferFrom", null) // TODO: check also transferFrom (might have differrent adress copied
+            dataDecodedDto = DataDecodedDto(
+                "safeTransferFrom",
+                listOf(
+                    ParamsDto("from", "address", defaultFromAddress.asEthereumAddressChecksumString()),
+                    ParamsDto("to", "address", defaultToAddress.asEthereumAddressChecksumString()),
+                    ParamsDto("tokenId", "uint256", "23")
```
elgatovital, line 164: why is `tokenId` set to 23 here? erc721 transfer json has a different value

biafra23: "23" is just a random value for me. Like "foo" or "bar". Maybe "<random value>" is better? Or a `val defaultTokenId = "23"`?

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
biafra23: I will ask Uxio if that is possible.

data/src/test/java/io/gnosis/data/repositories/TransactionRepositoryTest.kt
biafra23: Extracted to defaultTokenId in b4c2b8a8e90f167fd41626ac300cc715b5b38476

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
biafra23: Uxio said: "No, but it could be possible that there's no data decoded due to this https://vessenes.com/the-erc20-short-address-attack-explained/#:~:text=What%20happens%20in%20this%20attack,given%20to%20the%20shortened%20address.&text=When%20the%20parser%20is%20getting,and%20calls%20it%20a%20day."

biafra23: If there is no data decoded section, then the app will treat it as a custom transaction

biafra23: So I guess we're good here?

Issue #679, opened 2020-06-23
Add TransactionView subclasses for headers


elgatovital:
- [x] bind string value (used as title)
- [x] add string values to strings.xml

**Notes for QA:**
This ticket covers the headers in the tx list only. The two headers QUEUE and HISTORY should be shown right above 
Note: the transactions are not yet rendered correctly

Example: 
![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/c4609a9f-872d-46c0-97a7-10cf7abf9026)

liliya-soroka:
Verified
app version 1.12.1-45

PR #682, opened 2020-06-24
Transaction view subclasses

biafra23:
Regarding #679 .

- Adds section header view holder types
- Adds them in hardcoded places. (Needs transfer status to compute correct position)

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
@@ -15,7 +15,7 @@ class TransactionsViewModel
 ) : BaseStateViewModel<TransactionsViewState>(appDispatchers) {
 
     override fun initialState(): TransactionsViewState = TransactionsViewState(null, true)
-
+    var fnord: Boolean = false
     fun load() {
         safeLaunch {
             val safeAddress = safeRepository.getActiveSafe()!!.address
@@ -24,10 +24,15 @@ class TransactionsViewModel
                 TransactionsViewState(
                     isLoading = false,
                     viewAction = LoadTransactions(transactions.results.mapNotNull { transaction ->
-                        when (transaction) {
-                            is Transaction.Transfer -> TransactionView.Transfer(transaction, transaction.recipient == safeAddress)
-                            is Transaction.SettingsChange -> TransactionView.SettingsChange(transaction)
-                            else -> null
+                        if (!fnord) {
+                            fnord = true
+                            TransactionView.SectionHeader(transaction, "QueuedOrHistory")
+                        } else {
+                            when (transaction) {
+                                is Transaction.Transfer -> TransactionView.Transfer(transaction, transaction.recipient == safeAddress)
+                                is Transaction.SettingsChange -> TransactionView.SettingsChange(transaction)
+                                else -> null
+                            }
                         }
                     })
                 )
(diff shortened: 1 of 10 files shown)


app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
```diff
@@ -15,29 +19,46 @@ class TransactionsViewModel
...
 
     override fun initialState(): TransactionsViewState = TransactionsViewState(null, true)
-
     fun load() {
         safeLaunch {
             val safeAddress = safeRepository.getActiveSafe()!!.address
             val transactions = transactionRepository.getTransactions(safeAddress)
             updateState {
+                val newTransactions = newTransactions(transactions, safeAddress)
+                val transactionsWithSectionHeaders = addSectionHeaders(newTransactions)
```
biafra23, line 28: There is probably a more idiomatic way in Kotlin...

elgatovital: for me naming of `newTransactions` is confusing. Maybe renaming it to mapTransactions or something similar will make it more clear

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
```diff
@@ -15,29 +19,46 @@ class TransactionsViewModel
...
+            }
+        }
+    }
+
+    private fun addSectionHeaders(newTransactions: List<TransactionView>): List<TransactionView> {
+        val mutableList = newTransactions.toMutableList()
+        //TODO: Find first QUEUED tx and insert SectionHeader before it
+        mutableList.add(0, TransactionView.SectionHeader(title = R.string.tx_list_queue))
+        //TODO: Find first HISTORY tx and insert SectionHeader before it
+        mutableList.add(3, TransactionView.SectionHeader(title = R.string.tx_list_history))
```
elgatovital, line 55: we can't assume it will be at position 3 all the time

biafra23: Of course. That is what the TODO is for. As soon as we have the status field in Transactions, we can find the right spot.

Maybe I'll just wait for Joses PR? And then amend this?

elgatovital: yes, let's review & merge Joses PR first. It is already open

rmeissner:
Should we add/ adjust tests here? I find it interesting that we make changes to a view model but no tests need to be adjusted ;)

biafra23:
> Should we add/ adjust tests here? I find it interesting that we make changes to a view model but no tests need to be adjusted ;)

The changes to the ViewModel were fake. Now with the status in place I can make the real changes and test them.

app/src/test/java/io/gnosis/safe/ui/transaction/TransactionsViewModelTest.kt
```diff
@@ -0,0 +1,147 @@
...
+    @get:Rule
+    val instantExecutorRule = TestLifecycleRule()
+
+    private val safeRepository = mockk<SafeRepository>()
+    private val transactionRepository = mockk<TransactionRepository>()
+    private lateinit var viewModel: TransactionsViewModel
+    private val defaultSafeName: String = "Default Name"
+    private val defaultSafeAddress: Solidity.Address = "0x1234".asEthereumAddress()!!
+    private val defaultAddress: Solidity.Address = Solidity.Address(BigInteger.ZERO)
+    private val defaultSafe = Safe(defaultSafeAddress, defaultSafeName)
```
elgatovital, line 40: we have put test constants to companion object in other test classes. This is not a blocker, but we should follow same structure for tests

biafra23: Sorry, I merged prematurely. I will keep that in mind.

Issue #665, opened 2020-06-16
Transaction status calculation


jpalvarezl:
**Describe the solution you'd like**
Transaction status should be calculated according to what is presented in this link:
https://github.com/gnosis/safe/wiki/Transaction-app-types-and-mapping-from-json-to-UI-layer#status-calculation

**Additional context**
These methods can be located in the `TransactionRepository` unless there is a more suitable place for them. 

Precondition: `Safe` details are available

PR #683, opened 2020-06-24
Transaction status

jpalvarezl:
Reference #665 .

- Passing `SafeInfo` to `getTransaction` method
- Calculation of status for each transaction, module and ethereum transactions are always assumed as `Success`
- Unit tests 

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
@@ -19,7 +19,7 @@ class TransactionsViewModel
     fun load() {
         safeLaunch {
             val safeAddress = safeRepository.getActiveSafe()!!.address
-            val safeInfo = safeRepository.getSafeInfo(safeAddress)!!
+            val safeInfo = safeRepository.getSafeInfo(safeAddress)
             val transactions = transactionRepository.getTransactions(safeAddress, safeInfo)
             updateState {
                 TransactionsViewState(
--- a/app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
@@ -19,7 +19,8 @@ class TransactionsViewModel
     fun load() {
         safeLaunch {
             val safeAddress = safeRepository.getActiveSafe()!!.address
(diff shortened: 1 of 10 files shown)


app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
```diff
@@ -19,7 +19,8 @@ class TransactionsViewModel
     fun load() {
         safeLaunch {
             val safeAddress = safeRepository.getActiveSafe()!!.address
-            val transactions = transactionRepository.getTransactions(safeAddress)
+            val safeInfo = safeRepository.getSafeInfo(safeAddress)!!
```
elgatovital, line 22: safeRepository should give back the non-nullable object. Every added safe on the device should be valid and have safe info.

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
```diff
@@ -19,7 +19,8 @@ class TransactionsViewModel
     fun load() {
         safeLaunch {
             val safeAddress = safeRepository.getActiveSafe()!!.address
```
elgatovital, line 21: this will crash if we remove last safe on the device. Active safe will be null. We should check for that and load only if we have active safe

jpalvarezl: This can be addressed in a different PR, I would say :) since I am not really working with transaction loading here.

elgatovital: sure, but let's have an issue open

elgatovital: https://github.com/gnosis/safe-android/issues/684

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
```diff
@@ -121,57 +127,79 @@ class TransactionRepository(
...
+            status,
             null,
             transaction.to,
             transaction.data?.dataSizeBytes() ?: 0L,
             null,
             BigInteger.ZERO
         )
+    }
+
+    // Pending not reachable yet
```
elgatovital, line 194: please add todo/fixme for returning pending status

Issue #662, opened 2020-06-16
Transaction List pagination


jpalvarezl:
**Describe the solution you'd like**
The `all-transactions` endpoint returns items in pages that look like:

```kotlin
data class Page<T>(
    val count: Int,
    val next: String?,
    val previous: String?,
    val results: List<T>
)
```

We need to implement pagination so that we use the `next` link, if available to load more items in the list. Do not attempt setting `offset` or `limit` despite what the URL for `next` and `previous` may look like.

On the first page of data we should also fetch the `Safe` from the `/safes/{address}/` endpoint.

This is required so that we have the safe `nonce` and we can thus calculate the corresponding status of any given transactions. 

**Additional context**
Be reminded that the `TransactionRepository` expands each page of data, depending on whether or not many `transfers` are part of the response object.

elgatovital:
PageKeyedDataSource -> PagedList<Transaction>

Adjust adapter to work with PagedList (instead of List)
Make data source part of the view model? Try not to polute repository with ui stuff.
PagingDataAdapter (RecyclerView subclass)

Testing (view model: loading next page, loading previous page, ui updates)

PR #685, opened 2020-06-25
Implement tx list pagination using androidx paging3 lib

elgatovital:
Handles #662

- Add android paging library
- Add endpoint for loading pages of tx data
- Add data source for providing pages of tx data
- Use paging data adapter to display tx views
- Refactor tx view model to work with paging

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/transaction/TransactionView.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transaction/TransactionView.kt
@@ -7,7 +7,7 @@ import androidx.annotation.StringRes
 import io.gnosis.data.models.TransactionStatus
 import pm.gnosis.model.Solidity
 
-sealed class TransactionView(open val status: TransactionStatus) {
+sealed class TransactionView(open val status: TransactionStatus?) {
     data class Transfer(
         override val status: TransactionStatus,
         @StringRes val statusText: Int,
@@ -120,4 +120,6 @@ sealed class TransactionView(open val status: TransactionStatus) {
     ) : TransactionView(status)
 
     data class SectionHeader(@StringRes val title: Int) : TransactionView(TransactionStatus.Pending)
+
+    object Unknown : TransactionView(null)
 }
--- a/app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
@@ -488,36 +488,6 @@ class TransactionsViewModel
         } ?: false
     }
 
-    private fun addSectionHeaders(transactions: List<TransactionView>): List<TransactionView> {
-        val mutableList = transactions.toMutableList()
-
-        val firstQueuedTransaction = mutableList.indexOfFirst { transactionView ->
-            when (transactionView.status) {
-                TransactionStatus.Pending,
-                TransactionStatus.AwaitingConfirmations,
-                TransactionStatus.AwaitingExecution -> true
-                else -> false
-            }
-        }
-        if (firstQueuedTransaction >= 0) {
-            mutableList.add(firstQueuedTransaction, TransactionView.SectionHeader(title = R.string.tx_list_queue))
-        }
-
-        val firstHistoricTransaction = mutableList.indexOfFirst { transactionView ->
-            when (transactionView.status) {
-                TransactionStatus.Cancelled,
-                TransactionStatus.Failed,
-                TransactionStatus.Success -> true
-                else -> false
-            }
-        }
-
-        if (firstHistoricTransaction >= 0) {
-            mutableList.add(firstHistoricTransaction, TransactionView.SectionHeader(title = R.string.tx_list_history))
-        }
-        return mutableList
-    }
-
     companion object {
         const val OPACITY_FULL = 1.0F
         const val OPACITY_HALF = 0.5F
--- a/app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
@@ -32,6 +32,7 @@ import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.flow.collectLatest
 import kotlinx.coroutines.flow.map
 import pm.gnosis.model.Solidity
+import pm.gnosis.utils.asEthereumAddress
 import java.math.BigInteger
 import javax.inject.Inject
 
--- a/app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
@@ -2,8 +2,15 @@ package io.gnosis.safe.ui.transaction
 
 import android.view.View
 import androidx.annotation.StringRes
-import io.gnosis.data.models.*
+import androidx.lifecycle.viewModelScope
+import androidx.paging.PagingData
+import androidx.paging.cachedIn
+import androidx.paging.insertSeparators
+import io.gnosis.data.models.Safe
+import io.gnosis.data.models.SafeInfo
+import io.gnosis.data.models.Transaction
 import io.gnosis.data.models.Transaction.*
+import io.gnosis.data.models.TransactionStatus
 import io.gnosis.data.repositories.SafeRepository
 import io.gnosis.data.repositories.SafeRepository.Companion.METHOD_CHANGE_MASTER_COPY
 import io.gnosis.data.repositories.SafeRepository.Companion.METHOD_DISABLE_MODULE
@@ -14,26 +21,29 @@ import io.gnosis.data.repositories.SafeRepository.Companion.SAFE_MASTER_COPY_0_1
 import io.gnosis.data.repositories.SafeRepository.Companion.SAFE_MASTER_COPY_1_0_0
 import io.gnosis.data.repositories.SafeRepository.Companion.SAFE_MASTER_COPY_1_1_1
 import io.gnosis.data.repositories.TokenRepository.Companion.ETH_SERVICE_TOKEN_INFO
-import io.gnosis.data.repositories.TransactionRepository
 import io.gnosis.data.repositories.getValueByName
 import io.gnosis.safe.R
 import io.gnosis.safe.ui.base.AppDispatchers
 import io.gnosis.safe.ui.base.BaseStateViewModel
-import io.gnosis.safe.ui.transaction.TransactionView.CustomTransaction
-import io.gnosis.safe.ui.transaction.TransactionView.CustomTransactionQueued
+import io.gnosis.safe.ui.transaction.list.TransactionPagingProvider
 import io.gnosis.safe.utils.shiftedString
+import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.collectLatest
+import kotlinx.coroutines.flow.map
 import pm.gnosis.model.Solidity
-import pm.gnosis.utils.asEthereumAddress
 import java.math.BigInteger
 import javax.inject.Inject
 
 class TransactionsViewModel
 @Inject constructor(
-    private val transactionRepository: TransactionRepository,
+    private val transactionsPager: TransactionPagingProvider,
     private val safeRepository: SafeRepository,
     appDispatchers: AppDispatchers
 ) : BaseStateViewModel<TransactionsViewState>(appDispatchers) {
+
+    private var currentSafeAddress: Solidity.Address? = null
+    private var currentSafeTxItems: Flow<PagingData<TransactionView>>? = null
 
     init {
         safeLaunch {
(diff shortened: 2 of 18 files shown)


app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
```diff
@@ -1,103 +1,122 @@
...
 class TransactionsViewModel
 @Inject constructor(
-    private val transactionRepository: TransactionRepository,
+    private val transactionsPagingRepository: TransactionPagingRepository,
     private val safeRepository: SafeRepository,
     appDispatchers: AppDispatchers
 ) : BaseStateViewModel<TransactionsViewState>(appDispatchers) {
 
+    private var currentSafeAddress: Solidity.Address? = null
+    private var currentSafeTxItems: Flow<PagingData<TransactionView>>? = null
```
jpalvarezl, line 31: Does this need to be a `var`? Couldn't we have a `flow` that starts empty and then emits as new pages of data become available?

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
```diff
@@ -1,103 +1,122 @@
...
-    private fun mapTransactions(
-        transactions: Page<Transaction>,
-        safe: Solidity.Address
-    ): List<TransactionView> {
-        return transactions.results.mapNotNull { transaction ->
-            when (transaction) {
-                is Transaction.Transfer -> TransactionView.Transfer(transaction, transaction.recipient == safe)
-                is Transaction.SettingsChange -> TransactionView.SettingsChange(transaction)
-                else -> null
+        currentSafeAddress = safe
```
jpalvarezl, line 45: Is this assignment necessary? You can only reach this point if, and only if, `safe` is equal to `currentSafeAddress` anyway.

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
```diff
@@ -1,103 +1,122 @@
...
+                    when (before.transaction?.status) {
+                        TransactionStatus.Pending,
+                        TransactionStatus.AwaitingConfirmation,
+                        TransactionStatus.AwaitingExecution -> when (after.transaction?.status) {
+                            TransactionStatus.Cancelled,
+                            TransactionStatus.Failed,
+                            TransactionStatus.Success -> TransactionView.SectionHeader(title = R.string.tx_list_history)
+                            else -> null  // no separator
+                        }
+                        else -> null  // no separator
```
jpalvarezl, line 90: This looks a bit bulky, do you think we could split up this lambda into a couple of methods? To know exactly what's going on. 

All the nested `when` blocks make it a bit hard to read.

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
```diff
@@ -1,103 +1,122 @@
...
 
 class TransactionsViewModel
 @Inject constructor(
-    private val transactionRepository: TransactionRepository,
+    private val transactionsPagingRepository: TransactionPagingRepository,
     private val safeRepository: SafeRepository,
     appDispatchers: AppDispatchers
 ) : BaseStateViewModel<TransactionsViewState>(appDispatchers) {
 
+    private var currentSafeAddress: Solidity.Address? = null
```
jpalvarezl, line 30: I could be wrong, but judging by where you use this `var` it might not be necessary at all to have it. I could have missed a place though.

app/src/main/java/io/gnosis/safe/ui/transaction/list/TransactionPagingSource.kt
```diff
@@ -0,0 +1,40 @@
...
+import java.io.IOException
+
+
+class TransactionPagingSource(
+    private val safe: Solidity.Address,
+    private val safeRepo: SafeRepository,
+    private val txRepo: TransactionRepository
+) : PagingSource<String, Transaction>() {
+
+    private lateinit var safeInfo: SafeInfo
```
jpalvarezl, line 19: Would it make sense to pass this as a constructor parameter? Then you can just require it upon construction of this object.

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
```diff
@@ -1,103 +1,122 @@
...
+                            else -> null  // no separator
+                        }
+                        else -> null  // no separator
+                    }
+                }
+            }
+            .cachedIn(viewModelScope)
+
+        currentSafeTxItems = safeTxItems
+        return safeTxItems
```
jpalvarezl, line 97: Where are you consuming this return value? Maybe there is a way to avoid having an inner private `var` keeping track of it?

app/src/main/java/io/gnosis/safe/ui/transaction/list/TransactionPagingRepository.kt
```diff
@@ -0,0 +1,26 @@
...
+import androidx.paging.Pager
+import androidx.paging.PagingConfig
+import androidx.paging.PagingData
+import io.gnosis.data.models.Transaction
+import io.gnosis.data.repositories.SafeRepository
+import io.gnosis.data.repositories.TransactionRepository
+import kotlinx.coroutines.flow.Flow
+import pm.gnosis.model.Solidity
+
+class TransactionPagingRepository(
```
jpalvarezl, line 12: `Repository` is a big word in our app :D Maybe calling it `TransactionPagingHelper` would be a bit more accurate?

app/src/main/java/io/gnosis/safe/ui/transaction/list/TransactionPagingSource.kt
```diff
@@ -0,0 +1,40 @@
...
+            LoadResult.Page(
+                data = response.results,
+                prevKey = response.previous,
+                nextKey = response.next
+            )
+        } catch (exception: IOException) {
+            return LoadResult.Error(exception)
+        } catch (exception: HttpException) {
+            return LoadResult.Error(exception)
+        }
```
jpalvarezl, line 40: Could this be written as a `runCatching` block? It's not a big deal anyway xD it just look a bit prettier :)

app/src/main/java/io/gnosis/safe/ui/transaction/list/TransactionPagingRepository.kt
```diff
@@ -0,0 +1,26 @@
...
+
+    fun getTransactionsStream(safe: Solidity.Address): Flow<PagingData<Transaction>> {
+        return Pager(
+            config = PagingConfig(pageSize = NETWORK_PAGE_SIZE, enablePlaceholders = false),
+            pagingSourceFactory = { TransactionPagingSource(safe, safeRepository, transactionRepository) }
+        ).flow
+    }
+
+    companion object {
+        private const val NETWORK_PAGE_SIZE = 50
```
jpalvarezl, line 24: Is this required by the framework? I think in the case of our backend it's actually 20.

app/src/main/java/io/gnosis/safe/ui/transaction/list/TransactionPagingRepository.kt
elgatovital: repository suits here better. This class provides pages of data, so it is a repository in the sense of a word. Helper is more misleading here as it usually indicates some utilitarian functionality

app/src/main/java/io/gnosis/safe/ui/transaction/list/TransactionPagingSource.kt
elgatovital: I would keep it this way. Right now loading of tx pages is completely encapsulated in the source. Having safeInfo as a parameter would mean we have to load it outside and pass it down to the pager. This increase view model code clutter

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
elgatovital: it is used for caching the paging results in the viewModel scope

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
elgatovital: it is used for caching - useful when navigating from and back to the fragment

app/src/main/java/io/gnosis/safe/ui/transaction/list/TransactionPagingRepository.kt
elgatovital: we need some default value. 20 is small in my opinion. You will be initiating lots of requests and ui experience might suffer.

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
elgatovital: don't see how. Imagine your active safe has changed -> load will be triggered -> current safe still points to the old value -> we don't use cached results, but request new pages and update currentSafeAddress

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
elgatovital: it is used for caching

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
elgatovital: here it is still clear what is going on. I think it is easier to read and understand now when everything is in one place. Should it become bulkier, more nested -> sure, I will dissect it into small pieces of easily digestible code

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
jpalvarezl: So, in that case using `safeRepository.getActiveSafe()` each time you call this method would be better?

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
jpalvarezl: I think that each `if` makes sense as a single method. Then you can describe with the method name what each `if` branch does.

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
jpalvarezl: Couldn't you use `safeRepository.getActiveSafe` every time you need the current safe instead of caching? That is technically the only source of truth you should trust when it comes to active safe.

app/src/main/java/io/gnosis/safe/ui/transaction/list/TransactionPagingSource.kt
jpalvarezl: Well, at the viewModel you have a coroutine scope from which you can launch anyway, then this class states in its constructor all the parameters it needs instead of requiring stuff that may not be initialised when it's needed.

app/src/main/java/io/gnosis/safe/ui/transaction/list/TransactionPagingRepository.kt
jpalvarezl: A repository, means that data could potentially be stored. This class, doesn't seem to be doing that. It orchestrates dependencies and maps them into a different wrapper type, which sounds to me like the sort of stuff a helper class does.

app/src/main/java/io/gnosis/safe/ui/transaction/list/TransactionPagingRepository.kt
jpalvarezl: Well, AFAIK the backend returns exactly 20 txs per page anyway, and we are not meant to fiddle with offsets and limit anyhow.

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
jpalvarezl: I understand that. But if you overwrite the `flow` every single time a new value is available, then you don't need a `flow` at all, I would say. Then you can just have a `PagingData<TransactionView>>?` var.

jpalvarezl: To me it seems like it all boils down to deciding whether this is a cache or a stream?

app/src/main/java/io/gnosis/safe/ui/transaction/list/TransactionPagingRepository.kt
elgatovital: there is no fiddling with offsets, but initial page size has to be known

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
elgatovital: please provide code snippet how you see it

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
elgatovital: new PagingData's are emitted when we load new pages. We overwrite the flow only if safe address changes

jpalvarezl: but does it matter that we have a new instance of `flow` each time? To me it seems similar to how we use `state`, in which case we can reuse the instance as it is more like a message bus rather than anything else.

app/src/main/java/io/gnosis/safe/ui/transaction/list/TransactionPagingRepository.kt
elgatovital: it can potentially store pages. We could use [RemoteMediator](https://developer.android.com/reference/kotlin/androidx/paging/RemoteMediator) and store pages in a database. So this page provider can be extended to store data if we decide to support offline tx browsing. Repository in the name is a good reminder of this potential

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
jpalvarezl: So generally when you have 2 levels of when blocks I would suggest adding methods with names that explain your code somewhat. Like you see for the 2nd method, I don't really know how and why there is a need to add the `history` header again in certain cases. 

```kotlin
                it.insertSeparators { before, after ->

                    if (after == null) {
                        // we're at the end of the list
                        return@insertSeparators null  // no separator
                    }

// start of fun "setListHeader"
                    if (before == null) {
                        // we're at the beginning of the list
                        return@insertSeparators when (after.transaction?.status) {
                            TransactionStatus.Pending,
                            TransactionStatus.AwaitingConfirmation,
                            TransactionStatus.AwaitingExecution -> TransactionView.SectionHeader(title = R.string.tx_list_queue)

                            TransactionStatus.Cancelled,
                            TransactionStatus.Failed,
                            TransactionStatus.Success -> TransactionView.SectionHeader(title = R.string.tx_list_history)

                            else -> null  // no separator
                        }
                    }
// end of fun "setListHeader"

// start of "iHaveNoIdeaWhatThisCodeDoes"
                    when (before.transaction?.status) {
                        TransactionStatus.Pending,
                        TransactionStatus.AwaitingConfirmation,
                        TransactionStatus.AwaitingExecution -> when (after.transaction?.status) {
                            TransactionStatus.Cancelled,
                            TransactionStatus.Failed,
                            TransactionStatus.Success -> TransactionView.SectionHeader(title = R.string.tx_list_history)
                            else -> null  // no separator
                        }
                        else -> null  // no separator
               }
// end of "iHaveNoIdeaWhatThisCodeDoes"
        }
}
```

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
jpalvarezl: Is this consumed in the fragment? I can't seem to find the place where you use the value returned.

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
elgatovital: we need to save last safe address. Otherwise we would not know if safe was changed when we select another safe from  safe selection dialog.

app/src/main/java/io/gnosis/safe/ui/transaction/list/TransactionPagingRepository.kt
jpalvarezl: I don't think it is a good practice to name classes after what they potentially could become. If you feel like this is an improvement we would need in the future, I would say that placing the link you provided in a comment within this class is a much better way to remind ourselves while at the same time naming the class for what it actually is at the time of the implementation.

app/src/main/java/io/gnosis/safe/ui/transaction/list/TransactionPagingRepository.kt
jpalvarezl: yeah, I am aware of that. But I was just letting you know that this value is known and it is `20`. If it doesn't need to match then it's all good :)

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
elgatovital: yes, because new instance of flow is created by a paging library

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
elgatovital: it is consumed in the view model's load() method

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
jpalvarezl: Don't we have a `flow` in the `SafeRepository` that we could use for this?

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionView.kt
```diff
@@ -7,7 +7,7 @@ import androidx.annotation.StringRes
 import io.gnosis.data.models.TransactionStatus
 import pm.gnosis.model.Solidity
 
-sealed class TransactionView(open val status: TransactionStatus) {
+sealed class TransactionView(open val status: TransactionStatus?) {
```
jpalvarezl, line 10: Maybe instead of making this nullable, we should leave it as is and add `TransactionStatus.Unknown`. Then we don't need to program so defensively. As far as I can see, this is null so you can pass `null` to `object Unknow : TransactionView(null)` ?

app/src/main/java/io/gnosis/safe/ui/transaction/list/TransactionViewListAdapter.kt
```diff
@@ -0,0 +1,47 @@
...
+    override fun onBindViewHolder(holder: BaseTransactionViewHolder<TransactionView>, position: Int) {
+        val uiModel = getItem(position)
+        holder.bind(uiModel!!, listOf())
+    }
+
+    override fun getItemViewType(position: Int): Int {
+        val uiModel = getItem(position)
+        return uiModel?.let {
+            factory.viewTypeFor(it).takeUnless { it < 0 } ?: throw UnsupportedViewType()
+        } ?: throw UnsupportedViewType()
```
jpalvarezl, line 31: Maybe we shouldn't crash here? Not sure though.

elgatovital: it should never occur in the first place

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionView.kt
elgatovital: it is overridden in the subclasses and it is not null there

elgatovital: Adding Unknown status would mean we have to extend all of when statements to account for it. It is just more boilerplate code

elgatovital:
> I would get @biafra23 to have a look at this before you merge. I wish there was a way to somehow preserve the work he did with tests. Maybe we could at least leave a comment with the test names removed, so that we re-implement them once we can decouple the list of TXs from paged data?

there were mainly mapping tests and they are still preserved. Only checks for header sections were removed

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionView.kt
jpalvarezl: Ok sounds good. Then let's not do it.

Issue #687, opened 2020-06-26
[Android] The header with safe info is not changed to the active after the new safe is added


liliya-soroka:
**Describe the bug**
The previous active safe info is displayed in the header after the new safe is added. At the same time the content of the opened page is changed to the new active safe

**To Reproduce**
Precondition: at least one safe should be loaded 
1. Open Transactions list for the active safe
2. Click switch safe icon and add new  safe
3. Look at the header

**Expected behavior**
The header and the page content should have data of the new active safe

**Video link**
https://drive.google.com/file/d/160WMdQGVtI40S6hAlY0Jel5YECSSGGtv/view?usp=sharing

**Smartphone (please complete the following information):**
 - Device: Huawei P30 and Google Pixel 4
 - Android Version: [10
 - App Version 1.12.1-45

**Additional context**

liliya-soroka:
Verified
app version 1.12.1-46

Issue #684, opened 2020-06-24
Transaction list empty state


elgatovital:
TransactionsViewModel.load:

check if active safe is null

jpalvarezl:
We should implement the empty state in accordance to the design:

https://zpl.io/a35ne9N

jpalvarezl:
@liliya-soroka Just FYI, the test cases with thought about were the following (not sure if we covered everything): 
- Fresh install of the app
- Add safe with TXs should refresh the list
- Removing safe, then no safe is selected, should show empty state
- Removing safe and switching to newly selected safe should show the new list

There is still no error handling in this screen, so sometimes you can run into the situation that something breaks underneath and you get displayed a blank screen. This should change after we've resolved the ticket for it. I am just waiting for feedback from Sasha or Tobi.

liliya-soroka:
according to design the screen transactions list (no safe) should be used  with Load safe button in the case if there is no safe in the wallet https://app.zeplin.io/project/5e9efc30726e106958cc43de/screen/5eb929c14c23207639d1607e

https://app.zeplin.io/project/5e9efc30726e106958cc43de/screen/5eb2b9fbcea0c078fd77277c  should be used if the safe is added but the safe doesn't contain any transaction

liliya-soroka:
Verified
app version 1.12.1-46

PR #691, opened 2020-06-30
Empty state to be shown only when a safe has no transactions

jpalvarezl:
Reference #684 and #687 .

- Empty state is only shown if the loaded safe has no transactions

Remaining issues: the tool bar is not updated on safe change because it is not possible to inherit from `SafeOverviewBaseFragment` without losing tracking information. Not sure what to do about that.

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsFragment.kt
@@ -28,6 +28,7 @@ class TransactionsFragment : SafeOverviewBaseFragment<FragmentTransactionsBindin
     lateinit var viewModel: TransactionsViewModel
 
     private val adapter by lazy { MultiViewHolderAdapter(TransactionViewHolderFactory()) }
+    private val noSafeFragment by lazy { NoSafeFragment.newInstance(NoSafeFragment.Position.TRANSACTIONS) }
 
     override fun inject(component: ViewComponent) {
         component.inject(this)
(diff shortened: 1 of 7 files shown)


app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsFragment.kt
```diff
@@ -46,20 +48,38 @@ class TransactionsFragment : BaseViewBindingFragment<FragmentTransactionsBinding
...
 
+    private fun loadNoSafeFragment() {
+        with(binding) {
+            transactions.visible(false)
+            progress.visible(false)
+            imageEmpty.visible(false)
+            labelEmpty.visible(false)
+            noSafe.apply {
+                childFragmentManager.beginTransaction()
+                    .add(noSafe.id, NoSafeFragment.newInstance(NoSafeFragment.Position.TRANSACTIONS))
```
elgatovital, line 69: Probably replace instead of add would be better. Also here we might add fragment multiple times here. E.g. we have no safes, start add flow from transactions tab and after we done delete the newly added safe. In this case another fragment will be created and added. We should either reuse existing fragment or manually pop childFragmentManager's backstack when safe is loaded.

jpalvarezl: good call. Should I just call `replace` instead of `add`?
```suggestion
                    .replace(noSafe.id, NoSafeFragment.newInstance(NoSafeFragment.Position.TRANSACTIONS))
```

elgatovital: This will solve redundancy, but it will still hang around after safe is loaded, only in an invisible state. I think better to pop it off the stack

PR #699, opened 2020-07-03
Binding view holders

biafra23:
Regarding #663  .

- populate tx properties to viewholders
- 
- 

Missing: 
- Proper DateTime formatter (future PR)
- Proper detection of queued ERC20/ER721 transfers (Reason: missing tokenInfo in un-mined transactions)
- Tests are missing
- There is a lot of duplication in the the ViewHolderFactory. I would like to reduce it, when I have more unit tests

@gnosis/mobile-devs


data/src/main/java/io/gnosis/data/backend/TransactionSerivceApi.kt
```diff
@@ -13,7 +13,7 @@ interface TransactionServiceApi {
     @GET("v1/safes/{address}/balances/usd/")
     suspend fun loadBalances(@Path("address") address: String): List<ServiceBalance>
 
-    @GET("v1/safes/{address}/all-transactions/")
+    @GET("v1/safes/{address}/all-transactions/?limit=100")
```
biafra23, line 16: Useful for testing while we don't have paging. Must be removed, when we have paging

data/src/main/java/io/gnosis/data/models/Transaction.kt
```diff
@@ -19,26 +29,47 @@ sealed class Transaction {
...
 
-enum class TransactionStatus {
-    AwaitingConfirmation,
-    AwaitingExecution,
-    Cancelled,
-    Failed,
-    Success,
-    Pending // Not supported yet as these correspond to the ones issued from the device
+enum class TransactionStatus(val displayString: String) {
+    AwaitingConfirmations("Awaiting confirmations"),
```
elgatovital, line 69: better not to mix models and ui related elements like display strings. We should be using localized string resources and the decision which string to show should be done at the view model level

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
```diff
@@ -137,48 +143,76 @@ class TransactionRepository(
...
+            transaction.bestAvailableDate(),
+            tokenInfo,
+            transaction.nonce
         )
     }
 
+    private fun serviceTokenInfo(transfer: TransferDto): ServiceTokenInfo =
+        transfer.tokenInfo
+            ?: transfer.tokenAddress?.let { address ->
+                Log.w("NFT", "transfer.tokenAddress: ${transfer.tokenAddress}")
```
elgatovital, line 163: Log -> Timber

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
```diff
@@ -137,48 +143,76 @@ class TransactionRepository(
...
+                    ENS_ERC721_TOKEN_INFO
+                } else {
+                    NFT_ERC721_TOKEN_INFO
+                }
+            } ?: ETH_SERVICE_TOKEN_INFO
+
+
+    private fun serviceTokenInfo(transaction: MultisigTransactionDto): ServiceTokenInfo {
+        return if (transaction.transfers != null && transaction.transfers.isNotEmpty()) {
+            return serviceTokenInfo(transaction.transfers[0])
```
elgatovital, line 174: no need for an extra return statement

data/src/test/java/io/gnosis/data/repositories/TransactionRepositoryTest.kt
```diff
@@ -154,22 +158,23 @@ class TransactionRepositoryTest {
...
-            assertEquals(FAKE_ERC20_TOKEN_INFO, tokenInfo)
+            assertEquals(ETH_SERVICE_TOKEN_INFO, tokenInfo)
         }
     }
 
     @Test
     fun `getTransactions (multisig transaction for ERC721) should return transfer`() = runBlockingTest {
         val transactionDto = buildMultisigTransactionDto(
-            contractInfoType = ContractInfoType.ERC721,
+//            contractInfoType = ContractInfoType.ERC721,
```
elgatovital, line 168: can this be deleted?

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
biafra23: Timber is not (yet) available in data. But also the log statement was only for debugging. So I deleted it.

data/src/main/java/io/gnosis/data/backend/TransactionSerivceApi.kt
jpalvarezl: Actually, I would say this should be removed before this PR is merged.

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionViewHolderFactory.kt
```diff
@@ -36,87 +56,223 @@ class TransactionViewHolderFactory : BaseFactory<BaseTransactionViewHolder<Trans
...
+            confirmationsIcon.setImageDrawable(resources.getDrawable(viewTransfer.confirmationsIcon))
+            confirmations.text = resources.getString(R.string.tx_list_confirmations, viewTransfer.confirmations, viewTransfer.threshold)
+            nonce.text = viewTransfer.nonce
+        }
     }
 }
 
 class SettingsChangeViewHolder(private val viewBinding: ItemTxSettingsChangeBinding) :
     BaseTransactionViewHolder<TransactionView.SettingsChange>(viewBinding) {
+    private val resources = viewBinding.root.context.resources
```
jpalvarezl, line 134: let's not keep  reference to `context` related objects in the `viewHolder`. If you need `resources` somewhere in your `bind` method, I would ask you to inline this `val` or move it within the `bind` method. You have access to `viewBinding` there anyhow.

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionViewHolderFactory.kt
```diff
@@ -36,87 +56,223 @@ class TransactionViewHolderFactory : BaseFactory<BaseTransactionViewHolder<Trans
...
+            settingName.alpha = viewTransfer.alpha
         }
     }
 }
 
-class TransferViewHolder(private val viewBinding: ItemTxTransferBinding) :
-    BaseTransactionViewHolder<TransactionView.Transfer>(viewBinding) {
+class SettingsChangeQueuedViewHolder(private val viewBinding: ItemTxQueuedSettingsChangeBinding) :
+    BaseTransactionViewHolder<TransactionView.SettingsChangeQueued>(viewBinding) {
+    private val resources = viewBinding.root.context.resources
```
jpalvarezl, line 152: Same, here. Keeping a reference to resource that can be accessed when it's required by `bind`

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionViewHolderFactory.kt
```diff
@@ -36,87 +56,223 @@ class TransactionViewHolderFactory : BaseFactory<BaseTransactionViewHolder<Trans
...
+            confirmations.text = resources.getString(R.string.tx_list_confirmations, viewTransfer.confirmations, viewTransfer.threshold)
         }
     }
 }
 
-class ChangeMastercopyQueuedViewHolder(viewBinding: ItemTxQueuedChangeMastercopyBinding) :
-    BaseTransactionViewHolder<TransactionView.SettingsChangeQueued>(viewBinding) {
+class ChangeMastercopyViewHolder(private val viewBinding: ItemTxChangeMastercopyBinding) :
+    BaseTransactionViewHolder<TransactionView.ChangeMastercopy>(viewBinding) {
+    private val resources = viewBinding.root.context.resources
```
jpalvarezl, line 168: Same as above.

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionViewHolderFactory.kt
```diff
@@ -36,87 +56,223 @@ class TransactionViewHolderFactory : BaseFactory<BaseTransactionViewHolder<Trans
...
+            moduleAddress.visibility = viewTransfer.visibilityModuleAddress
+        }
     }
 }
 
-class SettingsChangeQueuedViewHolder(viewBinding: ItemTxQueuedSettingsChangeBinding) :
-    BaseTransactionViewHolder<TransactionView.SettingsChangeQueued>(viewBinding) {
+class ChangeMastercopyQueuedViewHolder(private val viewBinding: ItemTxQueuedChangeMastercopyBinding) :
+    BaseTransactionViewHolder<TransactionView.ChangeMastercopyQueued>(viewBinding) {
+    private val resources = viewBinding.root.context.resources
```
jpalvarezl, line 197: Same as above.

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionViewHolderFactory.kt
```diff
@@ -36,87 +56,223 @@ class TransactionViewHolderFactory : BaseFactory<BaseTransactionViewHolder<Trans
...
+            moduleAddress.visibility = viewTransfer.visibilityModuleAddress
+        }
     }
 }
 
-class TransferQueuedViewHolder(viewBinding: ItemTxQueuedTransferBinding) :
-    BaseTransactionViewHolder<TransactionView.TransferQueued>(viewBinding) {
+class CustomTransactionQueuedViewHolder(private val viewBinding: ItemTxQueuedTransferBinding) :
+    BaseTransactionViewHolder<TransactionView.CustomTransactionQueued>(viewBinding) {
+    private val resources = viewBinding.root.context.resources
```
jpalvarezl, line 223: Same as above.

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionViewHolderFactory.kt
```diff
@@ -36,87 +56,223 @@ class TransactionViewHolderFactory : BaseFactory<BaseTransactionViewHolder<Trans
...
+
+            blockies.setAddress(viewTransfer.address)
+            ellipsizedAddress.text = viewTransfer.address.formatForTxList()
+        }
+    }
+}
+
+class CustomTransactionViewHolder(private val viewBinding: ItemTxTransferBinding) :
+    BaseTransactionViewHolder<TransactionView.CustomTransaction>(viewBinding) {
+    private val resources = viewBinding.root.context.resources
```
jpalvarezl, line 249: Same as above

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionViewHolderFactory.kt
```diff
@@ -129,3 +285,18 @@ class SectionHeaderViewHolder(private val viewBinding: ItemTxSectionHeaderBindin
...
+        TransactionStatus.Cancelled -> resources.getColor(R.color.dark_grey)
+        else -> resources.getColor(R.color.safe_failed_red)
+    }
+}
+
+private fun formatAmount(viewTransfer: Transaction.Transfer, incoming: Boolean): String {
+    val inOut = if (incoming) "+" else "-"
+    val symbol = viewTransfer.tokenInfo?.symbol
+    val value: String = viewTransfer.tokenInfo?.decimals?.let { viewTransfer.value.shiftedString(decimals = it) }.toString()
+    return "%s%s %s".format(inOut, value, symbol)
```
jpalvarezl, line 301: seems here that value could be the string "null" ? Maybe it makes more sense to return an empty string if `value` is null?

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
```diff
@@ -59,24 +75,361 @@ class TransactionsViewModel
...
+                transaction is SettingsChange && isHistoricSettingsChange(transaction) -> historicSettingsChange(transaction)
+                transaction is Custom && isQueuedCustomTransaction(transaction) -> queuedCustomTransaction(transaction, safeInfo.threshold)
+                transaction is Custom && isHistoricCustomTransaction(transaction) -> historicCustomTransaction(transaction)
                 else -> null
             }
         }
     }
 
+    private fun isHistoricSetFallbackHandler(settingsChange: SettingsChange): Boolean {
+        return settingsChange.isSetFallBackHandler() && settingsChange.isCompleted()
```
jpalvarezl, line 100: maybe the method that is part of the `SettingsChange` class at the moment, should be inlined here.

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
```diff
@@ -59,24 +75,361 @@ class TransactionsViewModel
...
+    }
+
+    private fun isQueuedSetFallbackHandler(settingsChange: SettingsChange): Boolean {
+        return settingsChange.isSetFallBackHandler() && !settingsChange.isCompleted()
```
jpalvarezl, line 104: Same here, and all the same below.

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
```diff
@@ -101,14 +454,119 @@ class TransactionsViewModel
...
+        val address: Solidity.Address,
+        val confirmations: Int,
+        val threshold: Int,
+        @ColorRes val confirmationsTextColor: Int,
+        @DrawableRes val confirmationsIcon: Int,
+        val dataSizeText: String,
+        val amountText: String,
+        @ColorRes val amountColor: Int,
+        val nonce: String
+    ) : TransactionView(status)
```
jpalvarezl, line 567: Oh man, this really blew up in terms of complexity. Why did you opt out from simply passing the `Transaction.SettingsChange` or `Transaction.Transfer` or `Transaction.Custom` upwards? These UI assets would make total sense to resolve in the `viewHolder`, actually you have them requested and used in the same place. Here in the `TransactionsViewModel` feels to me that it loses a bit of visibility to where all of this is used.

jpalvarezl: There is a lot of UI specific stuff that doesn't need to be defined by the viewModel. I think a lot of the logic that you defined here would be easier to digest if it were placed in the respective viewHolder, as a lot of it, is only applicable to the specific layout the viewHolder binds to.

jpalvarezl: I think it is ok to have a viewModel/viewHolder pair defined for each cell, but as it is it feels a bit like a monolithical solution provided uniquely by the `TransactionsViewModel` which is a bit detrimental to readability. I think maybe it would be a good idea to address this in the future. As this, I suppose, works already. Then we could plan for a bit of a refactor/cleanup, if you would agree?

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
```diff
@@ -137,66 +153,103 @@ class TransactionRepository(
...
-    private fun custom(transaction: TransactionDto, safeInfo: SafeInfo? = null): Transaction.Custom {
+    private fun custom(transaction: TransactionDto, safeInfo: SafeInfo): Transaction.Custom {
         val status =
-            if (transaction is MultisigTransactionDto && safeInfo != null) transaction.status(safeInfo)
+            if (transaction is MultisigTransactionDto) transaction.status(safeInfo)
             else TransactionStatus.Success
 
+        val confirmations =
+            if (transaction is MultisigTransactionDto) transaction.confirmations?.size ?: 0
+            else null
```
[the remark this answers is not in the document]
biafra23: 0 should be fine in any case. I will change this.

data/src/main/java/io/gnosis/data/models/Transaction.kt
```diff
@@ -19,23 +29,48 @@ sealed class Transaction {
...
+            return "changeMasterCopy" == dataDecoded.method
+        }
+
+        fun isSetFallBackHandler(): Boolean {
+            return "setFallbackHandler" == dataDecoded.method
+        }
+
+        fun isModuleChange(): Boolean {
+            return "enableModule" == dataDecoded.method || "disableModule" == dataDecoded.method
+        }
```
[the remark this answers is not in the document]
biafra23: It can't be in the viewHolder, because these checks need to be done to create the viewHolder. I inlined these methods and created isComplete(tx) in TransactionViewModel.

data/src/main/java/io/gnosis/data/models/Transaction.kt
```diff
@@ -19,23 +29,48 @@ sealed class Transaction {
...
+
+    fun isCompleted(): Boolean =
+        when (status) {
+            AwaitingConfirmations,
+            AwaitingExecution,
+            Pending -> false
+            Success,
+            Failed,
+            Cancelled -> true
+        }
```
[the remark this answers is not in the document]
biafra23: I created isComplete(tx) in TransactionViewModel.

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionViewHolderFactory.kt
biafra23: Moved to bind method

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
biafra23: Done

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionsViewModel.kt
biafra23: The viewHolder classes should not have any logic. The logic should be in a class that can be unit tested. That is why I decided to do the conversion in the viewModel. We can certainly moved it somewhere else. But the TransactionView classes should be dumb data containers.

biafra23: ViewHolders don't need to be data classes. I guess thats what you're referring to? Let's do a refactoring! As a separate PR. We could probably get rid of  a lot of duplication. Maybe we don't need independent classes for queued and history txs, for example.

jpalvarezl: Absolutely. I totally agree with you on what the viewHolders should do. My point was more towards the fact that the `TransactionView` classes became kind of a full view descriptor; holding info about variables that are too close to the view layer, like alpha, colours and string asset ids. That's a little too fine grained in my opinion.

Issue #649, opened 2020-06-10
[Android] The Load Safe process is closed if the app is re-launch from background


liliya-soroka:
**Describe the bug**
The user has to start the Load Safe process one more time if the app is relaunched ( the app was in background)

**To Reproduce**
1. Run the app
2. Open any step from Load Safe Multisig ( Enter ENs name for example)
3. Send the app to the background ( open any other application)
4. Click on the app icon

**Current result** The app is started from the splash screen and the Assets page is displayed. The whole process of the load safe is escaped and the user has to start the process again

**Expected behavior**
The app should start from the page it was sent to the background

**Smartphone (please complete the following information):**
 - Device: Hexus p30
 - Android Version: 10
 - App Version 1.12.1-28 - yellow

liliya-soroka:
Verified
app version 1.12.1-52

PR #711, opened 2020-07-08
Made the add safe flow part of the main graph

jpalvarezl:
Reference  #649  .

- The add safe load process is part of the main graph now

@gnosis/mobile-devs


app/src/main/res/navigation/main_nav.xml
```diff
@@ -6,11 +6,9 @@
     android:id="@+id/main_nav_graph.xml"
     app:startDestination="@id/safeBalancesFragment">
 
-    <include app:graph="@navigation/add_safe_nav" />
-
     <action
         android:id="@+id/action_to_add_safe_nav"
```
elgatovital, line 10: we don't need this global action now. You can navigate directly to addSafeFragment

app/src/main/res/navigation/main_nav.xml
```diff
@@ -6,11 +6,9 @@
     android:id="@+id/main_nav_graph.xml"
     app:startDestination="@id/safeBalancesFragment">
 
-    <include app:graph="@navigation/add_safe_nav" />
```
elgatovital, line 9: add_save_nav.xml can be deleted now

jpalvarezl: This has been deleted already, if you look at the other file change in this PR.

app/src/main/res/navigation/main_nav.xml
jpalvarezl: We need it for `NoSafeFragment` . If you checkout the branch and "look for usages" you will find it is still in use.

elgatovital: why can't you go from NoSafeFragment to AddSafeFragment?

PR #713, opened 2020-07-09
#663 binding view holders test

biafra23:
Regarding #663  .

- Adds tests for viewModel class
- Extracts several constants
- 

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/dialogs/EnsInputDialog.kt
```diff
@@ -16,6 +16,8 @@ import io.gnosis.safe.R
 import io.gnosis.safe.di.components.DaggerViewComponent
 import io.gnosis.safe.di.modules.ViewModule
 import io.gnosis.safe.helpers.AddressHelper
+import io.gnosis.safe.ui.transaction.TransactionsViewModel.Companion.OPACITY_FULL
```
elgatovital, line 19: better to extract those values and put them into dimens.xml

app/src/main/java/io/gnosis/safe/ui/transaction/TransactionView.kt
```diff
@@ -58,7 +58,7 @@ sealed class TransactionView(open val status: TransactionStatus) {
         val nonce: String
     ) : TransactionView(status)
 
-    data class ChangeMastercopy(
+    data class SettingsChangeVariant(
```
elgatovital, line 61: why such strange name?

biafra23: Because it is used for more than only change mastercopy. Do you have a better name. It is used for enable/disable module, change mastercopy and change fallback handler.

app/src/main/java/io/gnosis/safe/ui/dialogs/EnsInputDialog.kt
biafra23: What would be the advantage to have this in dimens.xml? It is not used in layout.

elgatovital: it is used in EnsInputDialog. Which is strange. Why should EnsInputDialog import something from TransactionsViewModel?

biafra23: How can put floating point values into diments.xml?

elgatovital: ```<item name="name" format="float" type="dimen">1.0</item>```

elgatovital:
```kotlin

biafra23: That is a lot of code. Maybe I just revert to 1F and 0.5F?

PR #724, opened 2020-07-10
Add deployments for mainnet and rinkeby

### LINKED ISSUE #648: Create firebase app distribution for production (mainnet & rinkeby)
### Expected behavior

We want to be able to distribute `io.gnosis.safe` and `io.gnosis.safe.rinkeby` via firebae app distribution. As we do currently for the `io.gnosis.safe.internal` app. 

- Please also check that the google-services.json files work for all 3 apps: internal, production & rinkeby
- we need to make sure that subscribers are notified about new releases. ATM only internal-beta users are notified about new internal releases. Those users who subscribed via the subscription link are not notified

biafra23:
Closes #648  .

- Added pipeline scripts to be used in buildkite
- Basically duplicates everything to make mainnet and rikeby app distributions
- api keys might need to be adjusted additionally on the ci server
- necessary (mainnet-release & rinkeby-release) groups have need created in firebase projects

@gnosis/mobile-devs


ci/upload_app_distribution_mainnet.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
elgatovital, line 1: upload script remains basically the same. We can leave just one upload_app_distribution.sh file and just pass arguments to this bash script

biafra23: changed in 0d9a4935498e880bf0da1640e17e5203deb0bb58

Issue #594, opened 2020-05-07
Safe Settings: Advanced screen


elgatovital:

elgatovital:
![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/0a3ea531-b27a-448e-8f16-46890ea33847)

jpalvarezl:
Sasha provided the mock for when a fallback handler is not set: https://zpl.io/aMJvzgK

liliya-soroka:
Verified
app version 1.12.1-59
Sprint 7 - Safe settings test run in TestRails

PR #725, opened 2020-07-10
Advanced Safe Settings

jpalvarezl:
Reference #594 .

- Added advanced safe settings
- Fallback handler can potentially be null, the default or unknown
- Module names are not resolved yet as it is not specified how

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/safe/settings/safe/AdvancedSafeSettingsFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/safe/settings/safe/AdvancedSafeSettingsFragment.kt
@@ -1,11 +1,14 @@
 package io.gnosis.safe.ui.safe.settings.safe
 
+import android.content.res.Resources
 import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.MenuItem
 import android.view.View
 import android.view.ViewGroup
+import android.widget.FrameLayout
 import android.widget.LinearLayout
+import androidx.annotation.StringRes
 import androidx.appcompat.app.AppCompatActivity
 import androidx.core.content.ContextCompat
 import androidx.lifecycle.Observer
@@ -18,10 +21,13 @@ import io.gnosis.safe.di.components.ViewComponent
 import io.gnosis.safe.ui.base.BaseStateViewModel
 import io.gnosis.safe.ui.base.BaseViewBindingFragment
 import io.gnosis.safe.ui.safe.settings.view.LabeledAddressItem
+import io.gnosis.safe.ui.safe.settings.view.SettingItem
+import io.gnosis.safe.utils.dpToPx
 import pm.gnosis.model.Solidity
 import pm.gnosis.svalinn.common.utils.snackbar
 import pm.gnosis.svalinn.common.utils.visible
 import timber.log.Timber
+import java.math.BigInteger
 import javax.inject.Inject
 
 class AdvancedSafeSettingsFragment : BaseViewBindingFragment<FragmentAdvancedSafeSettingsBinding>() {
@@ -60,18 +66,30 @@ class AdvancedSafeSettingsFragment : BaseViewBindingFragment<FragmentAdvancedSaf
 
     private fun setSafeInfo(safeInfo: SafeInfo) {
         with(binding) {
-            val fallbackHandlerLabel =
-                if (viewModel.isDefaultFallbackHandler(safeInfo.fallbackHandler)) R.string.safe_settings_default_fallback_handler
-                else R.string.safe_settings_unknown
             fallbackHandlerContainer.removeAllViews()
-            fallbackHandlerContainer.addView(labeledAddress(safeInfo.fallbackHandler, getString(fallbackHandlerLabel)))
+            fallbackHandlerContainer.addView(fallbackHandlerView(safeInfo.fallbackHandler))
             nonce.name = safeInfo.nonce.toString()
             modulesContainer.removeAllViews()
-            safeInfo.modules.forEachIndexed { index, module ->
+            safeInfo.modules.takeUnless { it.isEmpty() }?.forEachIndexed { index, module ->
                 modulesContainer.addView(labeledAddress(module, "Module $index"))
-            }
+            } ?: run { moduleLabel.visible(false) }
         }
     }
+
+    private fun fallbackHandlerView(fallbackHandler: Solidity.Address?): View =
+        when {
+            fallbackHandler == null || fallbackHandler.value == BigInteger.ZERO -> SettingItem(requireContext()).apply {
+                background = ContextCompat.getDrawable(requireContext(), R.drawable.background_selectable_white)
+                layoutParams = FrameLayout.LayoutParams(FrameLayout.LayoutParams.MATCH_PARENT, dpToPx(60))
+                openable = false
+                name = getString(R.string.safe_settings_not_set)
+            }
+            viewModel.isDefaultFallbackHandler(fallbackHandler) -> labeledAddress(fallbackHandler, R.string.safe_settings_default_fallback_handler)
+            else -> labeledAddress(fallbackHandler, R.string.safe_settings_unknown)
+        }
+
+    private fun labeledAddress(address: Solidity.Address, @StringRes labelId: Int): LabeledAddressItem =
+        labeledAddress(address, getString(labelId))
 
     private fun labeledAddress(address: Solidity.Address, label: String): LabeledAddressItem {
         return LabeledAddressItem(requireContext()).apply {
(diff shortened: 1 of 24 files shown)


app/src/main/java/io/gnosis/safe/ui/safe/settings/safe/AdvancedSafeSettingsFragment.kt
```diff
@@ -0,0 +1,130 @@
...
+    override fun inflateBinding(inflater: LayoutInflater, container: ViewGroup?): FragmentAdvancedSafeSettingsBinding =
+        FragmentAdvancedSafeSettingsBinding.inflate(inflater, container, false)
+
+    override fun inject(component: ViewComponent) {
+        component.inject(this)
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        setHasOptionsMenu(true)
```
elgatovital, line 48: google recommends calling setHasOptionsMenu in onCreate()

app/src/main/java/io/gnosis/safe/ui/safe/settings/safe/AdvancedSafeSettingsFragment.kt
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    private fun setUi(isLoading: Boolean, safeInfo: SafeInfo? = null) {
+        toggleLoading(isLoading)
+        safeInfo?.let { setSafeInfo(it) }
+    }
+
+    private fun setSafeInfo(safeInfo: SafeInfo) {
+        with(binding) {
+            fallbackHandlerContainer.removeAllViews()
```
elgatovital, line 68: there can be only one fallback handler or none at all for the older contract versions, right?

app/src/main/java/io/gnosis/safe/ui/safe/settings/safe/AdvancedSafeSettingsFragment.kt
```diff
@@ -0,0 +1,130 @@
...
+            viewModel.isDefaultFallbackHandler(fallbackHandler) -> labeledAddress(fallbackHandler, R.string.safe_settings_default_fallback_handler)
+            else -> labeledAddress(fallbackHandler, R.string.safe_settings_unknown)
+        }
+
+    private fun labeledAddress(address: Solidity.Address, @StringRes labelId: Int): LabeledAddressItem =
+        labeledAddress(address, getString(labelId))
+
+    private fun labeledAddress(address: Solidity.Address, label: String = ""): LabeledAddressItem {
+        return LabeledAddressItem(requireContext()).apply {
+            background = ContextCompat.getDrawable(requireContext(), R.drawable.background_selectable_white)
```
elgatovital, line 98: can we set background in xml?

app/src/main/res/layout/fragment_advanced_safe_settings.xml
```diff
@@ -0,0 +1,81 @@
...
+
+
+    <ProgressBar
+        android:id="@+id/progress"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_gravity="center"
+        android:visibility="gone"
+        tools:visibility="visible" />
+
```
elgatovital, line 30: it makes sense to have pull to refresh here and also on safe settings screen to reload safe details

app/src/main/res/layout/view_labeled_address_item.xml
```diff
@@ -0,0 +1,70 @@
+<?xml version="1.0" encoding="utf-8"?>
+<merge
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="68dp"
+    tools:parentTag="androidx.constraintlayout.widget.ConstraintLayout">
+
```
elgatovital, line 9: just some whitespaces

app/src/main/res/layout/view_labeled_address_item.xml
```diff
@@ -0,0 +1,70 @@
...
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        app:layout_constraintBottom_toTopOf="@id/address"
+        app:layout_constraintEnd_toStartOf="@id/link"
+        app:layout_constraintHorizontal_bias="0"
+        app:layout_constraintStart_toStartOf="@id/address"
+        app:layout_constraintTop_toTopOf="parent"
+        app:layout_constraintVertical_chainStyle="packed"
+        tools:text="DefaultCallbackHandler" />
+
```
elgatovital, line 23: wasting valuable space resource here :)

app/src/main/java/io/gnosis/safe/ui/safe/settings/safe/AdvancedSafeSettingsFragment.kt
jpalvarezl: This is needed because, if the Fallback handler is not set the view should be a `SettingsItem` which is 60dp high, instead of having the label/blockies/addres/link icon combined view. I am not too fond either of this :/

jpalvarezl: I think there could be a situation where the fallback handler could be set to an address that isn't the default. Safes with master copy version lower than 1.1.1  have the fallback handler set to null.

elgatovital: image
so its 68dp vs 60dp when it is not set. And it makes more effort to handle this case in the fallback handler view vs using settings view?

app/src/main/java/io/gnosis/safe/ui/safe/settings/safe/AdvancedSafeSettingsFragment.kt
```diff
@@ -0,0 +1,139 @@
...
+            mainContainer.visible(false)
+            progress.visible(false)
+        }
+        snackbar(requireView(), throwable.message ?: getString(R.string.error_invalid_safe))
+        Timber.e(throwable)
+    }
+
+    private fun updateLoading(showLoading: Boolean) {
+        with(binding) {
+            refresh.isRefreshing = false
```
elgatovital, line 124: not a blocker: when we have content and we do pull to refresh, showing progress bar is not necessary. Pull to refresh should be enough

app/src/main/java/io/gnosis/safe/ui/safe/settings/safe/AdvancedSafeSettingsViewModel.kt
```diff
@@ -0,0 +1,43 @@
...
+                    AdvancedSafeSettingsState(
+                        isLoading = false,
+                        viewAction = LoadSafeInfo(safeInfo)
+                    )
+                }
+            }
+        }
+    }
+
+    fun isDefaultFallbackHandler(fallbackHandle: Solidity.Address): Boolean =
```
elgatovital, line 32: this should be part of the safe repository

app/src/main/java/io/gnosis/safe/ui/safe/settings/safe/AdvancedSafeSettingsViewModel.kt
```diff
@@ -0,0 +1,43 @@
...
+class AdvancedSafeSettingsViewModel
+@Inject constructor(
+    private val safeRepository: SafeRepository,
+    appDispatchers: AppDispatchers
+) : BaseStateViewModel<AdvancedSafeSettingsState>(appDispatchers) {
+
+    override fun initialState(): AdvancedSafeSettingsState = AdvancedSafeSettingsState()
+
+    fun load() {
+        safeLaunch {
```
elgatovital, line 19: so initially there will be no progress indicator shown even though we are requesting recent safe info details from backend?

jpalvarezl: The default value is `true` for `loading` in the `AdvancedSafeSettingsState` . Check out line 40 in this file.

jpalvarezl: This is set to loading with the `override fun initialState()` method

app/src/main/java/io/gnosis/safe/ui/safe/settings/safe/AdvancedSafeSettingsViewModel.kt
jpalvarezl: I figured, because it is only an equality check, that putting this into the safeRepository would be a bit overkill. When there are more fallback handlers address we need to observe, kind of like with master copy, I would have something more generic, but for now I think this ok and it's also covered with tests which is nice :)

PR #729, opened 2020-07-14
Fix app distribution upload

elgatovital:
Make upload script executable again

@gnosis/mobile-devs


jpalvarezl:
Wouldn't it make more sense to just close this PR without merging? It doesn't seem to be changing anything 🤔

elgatovital:
> Wouldn't it make more sense to just close this PR without merging? It doesn't seem to be changing anything 🤔

It changes permissions for an upload script. chmod 755. Otherwise it can't be executed on buildkite

Issue #695, opened 2020-07-01
[Android]Receiver pop-up with copy address option


liliya-soroka:
### Expected behavior
1. User can view the Safe on the Receiver pop-up that is opened by click on the safe info in the header:
ENS name
Address
QR code
Identicon

2. It should be possible to copy address from the Receiver pop-up

3. Design:
https://app.zeplin.io/project/5e9efc30726e106958cc43de/screen/5ea803d790f5ffbfc82651b2
https://app.zeplin.io/project/5e9efc30726e106958cc43de/screen/5ea803d890f5ffbfc82655af

liliya-soroka:
Verified
app version 1.12.1.-57 
[https://app.zenhub.com/files/96440096/35b9e0ce-6c24-46dc-bba5-77875e15c40b/download](https://app.zenhub.com/files/96440096/35b9e0ce-6c24-46dc-bba5-77875e15c40b/download)
[https://app.zenhub.com/files/96440096/6904f1c2-406f-4121-a0c4-ea9b2197318e/download](https://app.zenhub.com/files/96440096/6904f1c2-406f-4121-a0c4-ea9b2197318e/download)

PR #734, opened 2020-07-14
Safe share pop-up dialog

jpalvarezl:
Reference #695 .

- Add pop-up share dialog
- QR code generation
- Address can be copied and checked in etherscan

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
```diff
@@ -70,6 +73,9 @@ class StartActivity : BaseActivity(), SafeOverviewNavigationHandler {
 
     private fun setSafe(safe: Safe) {
         with(toolbarBinding) {
+            safeImage.setOnClickListener {
+                Navigation.findNavController(this@StartActivity, R.id.nav_host).navigate(ShareSafeDialogDirections.actionGlobalShareSafeDialog())
```
elgatovital, line 77: It is possible to navigate using dialog id just like we do it with safeSelectionDialog:
```Navigation.findNavController(this@StartActivity, R.id.nav_host).navigate(R.id.shareSafeDialog)```
I think it is more readable this way. And we won't need another global action

File: app/src/main/java/io/gnosis/safe/ui/safe/share/ShareSafeDialog.kt [hunk 2]
```diff
@@ -0,0 +1,106 @@
...
+
+class ShareSafeDialog : BaseViewBindingDialogFragment<DialogShareSafeBinding>() {
+
+    @Inject
+    lateinit var viewModel: ShareSafeViewModel
+
+    override fun inflateBinding(inflater: LayoutInflater, container: ViewGroup?): DialogShareSafeBinding =
+        DialogShareSafeBinding.inflate(layoutInflater, container, false)
+
+    override fun screenId(): ScreenId? = null
```
elgatovital, line 33: we have to track this screen: ScreenId.SAFE_RECEIVE

app/src/main/java/io/gnosis/safe/ui/safe/share/ShareSafeDialog.kt
```diff
@@ -0,0 +1,106 @@
...
+                        )
+                    )
+                }
+            }
+            safeQrCode.setImageBitmap(safeDetails.qrCode)
+            safeEnsName.text = safeDetails.ensName ?: getString(R.string.safe_settings_not_set)
+        }
+    }
+
+    private fun toggleLoading(loading: Boolean) {
```
elgatovital, line 91: do you toggle loading for qr code generation or for ens name resolution as well?

app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
jpalvarezl: I was just using the `safeArgs`, but I agree in this case the `id` is shorter and looks nicer.

app/src/main/java/io/gnosis/safe/ui/safe/share/ShareSafeViewModel.kt
```diff
@@ -0,0 +1,65 @@
...
+    fun load() {
+        safeLaunch {
+            safeRepository.getActiveSafe()?.let { activeSafe ->
+                val ensName = runCatching { ensRepository.reverseResolve(activeSafe.address) }
+                    .onFailure { Timber.e(it) }
+                    .getOrNull()
+                val qrCode = runCatching {
+                    qrCodeGenerator.generateQrCodeSync(
+                        activeSafe.address.asEthereumAddressChecksumString(),
+                        512,
```
elgatovital, line 38: is there any reason to specify dimensions and background explicitly?

jpalvarezl: Those are the default values of the method for the rxJava implementation, which suited our purposes. If you check `svalinn` out. For some reason the bitmap gets generated with some padding already, which I found a bit annoying. Maybe this could be changed, but then we'd need a new release for `svalinn` I suppose. This forces us to pull in rx too :/

app/src/main/java/io/gnosis/safe/ui/safe/share/ShareSafeViewModel.kt
```diff
@@ -0,0 +1,65 @@
...
+                        512,
+                        Color.WHITE
+                    )
+                }.onFailure { Timber.e(it) }
```
elgatovital, line 42: should we propagate qr code generation error to user and show a snackbar? Or failing silently would be enough?

jpalvarezl: The legacy app failed silently, I think. We still log, but I think that the rest of the UI is still meaningful without a QR code.

app/src/main/res/values/styles.xml
```diff
@@ -188,6 +188,11 @@
         <item name="bottomSheetStyle">@style/BottomSheet</item>
     </style>
 
+    <style name="BaseSheetDialog" parent="Theme.AppCompat.Light.Dialog.Alert">
```
elgatovital, line 191: It's not a sheet dialog, but just a dialog -> BaseDialog

elgatovital, line 191: why SheetDialog and not just Dialog?

elgatovital:
first pr to go through new deployment steps 🚀

PR #735, opened 2020-07-15
Refactor package structure

elgatovital:
@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/safe/transactions/TransactionListViewHolderFactory.kt
```diff
@@ -95,6 +96,11 @@ class TransferViewHolder(private val viewBinding: ItemTxTransferBinding) :
             txTypeIcon.alpha = viewTransfer.alpha
             blockies.alpha = viewTransfer.alpha
             ellipsizedAddress.alpha = viewTransfer.alpha
+
+            root.setOnClickListener {
+                //TODO: pass tx data to details page
```
jpalvarezl, line 101: Just FYI, we should try using the `/transactions/{safe_tx_hash}/` afaik, but that only provides info for multisig transactions.

Issue #647, opened 2020-06-10
[Android] Load safe page: Address field should contain Identicon  and the whole entered address should be displayed


liliya-soroka:
**Describe the bug**
After the address is entered ( any method of entering) there is no address ldenticon and the address is not seen completely 
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/b7176a26-c1e9-4b43-b19f-7366b4fa6cdd)

**To Reproduce**
1. Go to Load Safe Multisig page
2. Enter the address ( any type of entering : QR code, paste or ENs name)

**Expected behavior**

- [x] show address Identicon
- [x] show the whole address in two lines
- [x] Hide text about new safe if the entered address is valid and the user can go to the next step

**Smartphone (please complete the following information):**
 - Device: Huawei p30
 - Android Version: 10
 - App Version 1.12.1.-28

**Additional context**

elgatovital:
![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/2510ca71-cb72-429d-834b-e6f348886116)

elgatovital:
![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/52e81f0a-0552-4c68-8957-9662173e4577)

liliya-soroka:
Reopened:
Current result: The address is not seen completely
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/a40d4d49-dd2f-4b10-abaf-d53400c3dcd1)
Expected result: The full address should be seen

app version 1.12.1-63
device : Huawei p30
os version 10

jpalvarezl:
@liliya-soroka I removed all the margin at the end of the address. This doesn't guarantee that the address won't ellipsize on very narrow screens. I made it so that the margins for the "..." icon are symmetrical.

liliya-soroka:
Verified
app version 1.12.1-64
staging, rinkeby

PR #741, opened 2020-07-15
UI improvements for Add Safe address screen

jpalvarezl:
Reference #647 .

- Fixed small issue with error propagation to safe settings. Adjusted tests
- Added `SafeInputView` complying with the designs

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/safe/add/SafeInputView.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/safe/add/SafeInputView.kt
@@ -1,0 +1,66 @@
+package io.gnosis.safe.ui.safe.add
+
+import android.content.Context
+import android.content.res.ColorStateList
+import android.util.AttributeSet
+import android.view.LayoutInflater
+import android.view.View
+import android.widget.LinearLayout
+import androidx.core.content.res.ResourcesCompat
+import io.gnosis.safe.R
+import io.gnosis.safe.databinding.ViewSafeInputBinding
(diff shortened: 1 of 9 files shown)


app/src/main/java/io/gnosis/safe/ui/safe/add/SafeInputView.kt
```diff
@@ -0,0 +1,66 @@
...
+
+class SafeInputView @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyleAttr: Int = 0
+) : LinearLayout(context, attrs, defStyleAttr) {
+
+    private val binding = ViewSafeInputBinding.inflate(LayoutInflater.from(context), this)
+
+    var address: Solidity.Address? = null
```
elgatovital, line 24: there is no need to use Solidity.Address here. address field could be just a String. There is anyway no validation here and error state is set to false regardless if address is a correct safe address. I think it is a good idea for this view to be dumb and just display stuff. All business logic & validation should be in view model

jpalvarezl: Well, I need it for the blockies as an Address. If the string is an invalid address, that process should happen outside this view. In the current state I would say this view is as dumb as it can be. It either displays an address or an error label.

elgatovital: having it as a string will alllow to potentially set values in xml. You can convert into solidity in the view itself.

Issue #661, opened 2020-06-16
Back link on the Enter_name screen doesn't lead on the previous step


liliya-soroka:
**Describe the bug**
Back link on the Enter_name screen doesn't lead on the previous step. Instead of it, Load safe flow is closed

**To Reproduce**
1. Go Load safe flow
2. Enter valid safe address
3. Click Next
4. Click Back link on the "Enter_name" screen

**Expected behavior**
Back link should lead on the previous step. The already entered on previous step data should be displayed

**Smartphone (please complete the following information):**
 - Device: Huawei p30
 - Android Version: 10
 - App Version 1.12.1-33

liliya-soroka:
Verified
app version 1.12.1-63

PR #743, opened 2020-07-16
Back navigation and state retention fixed add safe flow

jpalvarezl:
References #661 .

- Add safe address screen preserves state on back navigation and doesn't get inclusively popped.
- Added ripple effect for safe selection items, safe selection button (increased click area) and bottom nav bar
- 

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/safe/add/AddSafeViewModel.kt
```diff
@@ -13,45 +14,26 @@ class AddSafeViewModel
     appDispatchers: AppDispatchers
 ) : BaseStateViewModel<BaseStateViewModel.State>(appDispatchers) {
 
-    fun submitAddress(address: String) {
```
elgatovital, line 16: nice!

Issue #747, opened 2020-07-21
Wrong balances


tschubotz:
**Describe the bug**
The balances on the assets tab are wrong

**To Reproduce**
1. Clean install 
2. Add `0xA063Cb7CFd8E57c30c788A0572CBbf2129ae56B6`

Now I see:
- 1.383 ETH
- 650.608 dAI
- 295.148 SAI

**Expected behavior**
Balances like here: https://gnosis-safe.io/app/#/safes/0xA063Cb7CFd8E57c30c788A0572CBbf2129ae56B6/balances

**Screenshots**
I'm not allowed to take screenshots

**Smartphone (please complete the following information):**
 - Device: Pixel 3a
 - Android Version: 10
 - App Version: 1.12.1-65-gf499f6f1 (mainnet)

tschubotz:
The same is reproducible with the Rinkeby build and e.g. `0xaE3c91c89153DEaC332Ab7BBd167164978638c30` - balances are completely wrong

liliya-soroka:
Verified
Prod Rinkeby and Mainnet: app version 1.12.1-70

Issue #692, opened 2020-06-30
[Android]App settings: Email template is missed


liliya-soroka:
**Describe the bug**
Email template is not implemented. The empty a new email is opened ( only to field is predefined correctly)

**To Reproduce**
1. Go to App settings
2. Click get in touch
3. Select email 

**Expected behavior**
Email template
To: safe@gnosis.io
Subject: Feedback
Body:

<App name>
Safe address: <Safe address>
Feedback:
e.g. Safe Multisig v0.2.0 (1)
e.g. 0x8E77c8D47372Be160b3DC613436927FCc34E1ec0 (currently active Safe, None if no Safe added yet)

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/ad5082a7-fbba-4174-9c1d-0534965347f1)

**Smartphone (please complete the following information):**
 - Device: Google Pixel 4
 - Android Version: 10
 - App Version 1.12.1-45

**Additional context**

elgatovital:
Email template
To: safe@gnosis.io
Subject: Feedback
Body:

<App name>
Safe address: <Safe address>
Feedback:
e.g. Safe Multisig v0.2.0 (1)
e.g. 0x8E77c8D47372Be160b3DC613436927FCc34E1ec0 (currently active Safe, None if no Safe added yet)

liliya-soroka:
Reopened:
1. Could you start  the main data from the new lines:
<App name>
Safe address: <Safe address>
Feedback:
Current result: 
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/1513ffa6-cb96-4d56-8bf5-5115d40b7607)

2. Could we add the template for the email in Gmail app? Does any limitation exist?
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/75b79738-a91a-4368-bcc8-92174a0f37be)

app version 1.12.1-63
device: Pixel 4
os version: 10

liliya-soroka:
Reopened:
There is no template in the Gmail app on the Pixel 4 /Huawei P30
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/025f6664-d692-45b5-8675-4268e7459716)
app version 1.12.1. -66

liliya-soroka:
Verified:
It works for gmail too. App version 1.12.1-70
Could the eact record be started from the new line as in the example:
<App name>
[...]

PR #752, opened 2020-07-22
Show correct ERC20 token balances

elgatovital:
Handles #747 
Handles #692

- Pass data to gmail outside of intent extras due to gmail's ignorance!
- Parse balances correctly
- Add formatting pattern for ERC20 balances

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/settings/app/GetInTouchFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/settings/app/GetInTouchFragment.kt
@@ -91,6 +91,7 @@ class GetInTouchFragment : BaseViewBindingFragment<FragmentGetInTouchBinding>() 
                             "?subject=${getString(R.string.feedback_subject)}" +
                             "&body=${feedback}"
                 )
+                putExtra(Intent.EXTRA_EMAIL, getString(R.string.email_feedback))
                 putExtra(Intent.EXTRA_SUBJECT, getString(R.string.feedback_subject))
                 putExtra(Intent.EXTRA_TEXT, feedback)
             }
--- a/app/src/main/java/io/gnosis/safe/ui/settings/app/GetInTouchFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/settings/app/GetInTouchFragment.kt
@@ -19,7 +19,6 @@ import io.gnosis.safe.databinding.FragmentGetInTouchBinding
 import io.gnosis.safe.di.components.ViewComponent
 import io.gnosis.safe.ui.base.fragment.BaseViewBindingFragment
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.runBlocking
 import pm.gnosis.svalinn.common.utils.openUrl
 import pm.gnosis.svalinn.common.utils.snackbar
 import pm.gnosis.utils.asEthereumAddressString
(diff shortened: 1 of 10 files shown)


app/src/main/java/io/gnosis/safe/ui/settings/app/GetInTouchFragment.kt
```diff
@@ -86,7 +85,12 @@ class GetInTouchFragment : BaseViewBindingFragment<FragmentGetInTouchBinding>()
             val feedback = helper.createFeedbackText(requireContext())
             val intent = Intent(Intent.ACTION_SENDTO).apply {
                 type = "text/html"
-                data = Uri.parse("mailto:${getString(R.string.email_feedback)}")
+                // gmail is ignoring intent extras
+                data = Uri.parse(
```
DmitryBespalov, line 89: will it have any effect on other mail clients (non-gmail), i.e. would they have duplicate subject/text or body if this URI is used and the intent extras - as the code is doing now?

data/src/main/java/io/gnosis/data/adapters/MoshiBuilder.kt
```diff
@@ -14,8 +15,13 @@ internal val transactionDtoAdapter =
         .withDefaultValue(UnknownTransactionDto)
 
 val dataMoshi =
-    MoshiBuilderFactory.makeMoshiBuilder()
+    Moshi.Builder()
```
rmeissner, line 18: Why can the factory not be used anymore? Or other question why is the order important, shouldn't the annotation make it order independent.

app/src/main/java/io/gnosis/safe/ui/settings/app/GetInTouchFragment.kt
```diff
@@ -86,7 +85,12 @@ class GetInTouchFragment : BaseViewBindingFragment<FragmentGetInTouchBinding>()
...
+                    "mailto:${getString(R.string.email_feedback)}" +
+                            "?subject=${getString(R.string.feedback_subject)}" +
```
rmeissner, line 91: Does this work with other email clients as well?

app/src/main/java/io/gnosis/safe/ui/settings/app/GetInTouchFragment.kt
elgatovital: it is possible that some clients use intent extras and some just data uri. In worst case data will just be overridden.

app/src/main/java/io/gnosis/safe/ui/settings/app/GetInTouchFragment.kt
elgatovital: yes

data/src/main/java/io/gnosis/data/adapters/MoshiBuilder.kt
elgatovital: we still use factory. But I think it is more feasible to have moshi constructed in the app itself and not in the library. Library should not make too many assumptions about how moshi will be used or what type adapters will be needed.

rmeissner: Where do we use it? Shouldn't we use the same approach everywhere?

elgatovital: we probably won't need that many adapters/type converters in other places. Also having a complete builder at one place is better for readability I think.

rmeissner: This feels weird, if we think it is better to use then use it. In general I am more a fan of being explicit here and would remove the factory (as a lot of stuff in the moshi builder is order dependant)

elgatovital: here is pr for this: https://github.com/gnosis/svalinn-kotlin/pull/49 
I would release new patch version for savlinn and update dependencies within this pr @rmeissner

Issue #750, opened 2020-07-21
[Android] Safe settings: Edit name


liliya-soroka:
Edit name option is not implemented on the Safe settings page

elgatovital:
I would not label such tickets as bugs. This ticket was about implementing functionality. @tschubotz @liliya-soroka

liliya-soroka:
Verified
 app version 1.12.1-72 (576)

PR #755, opened 2020-07-24
Add screen for editing local safe name

elgatovital:
Handles #750 

- Add screen for editing local safe name
- Adjust nav graph
- React to local name changes in ui
     - Updating main toolbar
     - Updating name item

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/settings/safe/SafeSettingsEditNameFragment.kt
```diff
@@ -0,0 +1,67 @@
...
+        super.onViewCreated(view, savedInstanceState)
+        with(binding) {
+            backButton.setOnClickListener {
+                close()
+            }
+            safeName.showKeyboardForView()
+            safeName.setOnEditorActionListener listener@ { v, actionId, event ->
+                if (actionId == EditorInfo.IME_ACTION_DONE) {
+                    viewModel.saveLocalName(safeName.text.toString())
+                    return@listener true
```
rmeissner, line 45: So we only save the name if you press done on the keyboard? If you close the keyboard and go back the name will not be saved, right. Not sure if this is super intuitive, since there is no explicit save button. I would check this with @posthnikova and @tschubotz

app/src/test/java/io/gnosis/safe/ui/settings/safe/SafeSettingsEditNameViewModelTest.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    @get:Rule
+    val instantExecutorRule = TestLifecycleRule()
+
+    private val safeRepository = mockk<SafeRepository>(relaxed = true)
+    private lateinit var viewModel: SafeSettingsEditNameViewModel
+
+    @Test
+    fun `saveLocalName (name) - should change local name for safe`() = runBlockingTest {
+
```
rmeissner, line 31: This only tests the happy case. Would love to see test for some edge cases (e.g. no safe or error when interacting with the repo). But not super important right now.

Issue #714, opened 2020-07-09
[Android]Multisig reject transaction displays as incoming and nonce is missing in original transaction in Queue


liliya-soroka:
**Describe the bug**
When the one of the owners rejects transaction the incoming transaction appears in the transactions list in Queue

**To Reproduce**
1. Create transaction in web app in the safe with a few owners and a few confirmations
2. One of the owners rejects the transaction
3. Open transactions list in the mobile app for the same safe

**Expected behavior**
Multisig reject transaction should be displayed as outgoing

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/f7a1ac0c-2de4-4ea1-85a4-18ccca96d375)

**Smartphone (please complete the following information):**
 - Device: Pixel 4
 - Android Version: 10
 - App Version 0.4.2 - alpha -debug

**Additional context**

biafra23:
https://safe-team.staging.gnosisdev.com/app/#/safes/0x1C8b9B78e3085866521FE206fa4c1a67F49f153A/transactions has several unconfirmed txs which do not have a nonce in the android apps queued section.

See also: https://safe-transaction.staging.gnosisdev.com/api/v1/safes/0x1C8b9B78e3085866521FE206fa4c1a67F49f153A/all-transactions/?limit=100

liliya-soroka:
Verified
app version 1.12.1-76

Issue #712, opened 2020-07-08
[Android] Settings transaction doesn't have title in Queue section


liliya-soroka:
**Describe the bug**
Settings transaction doesn't have title in Queue section

**To Reproduce**
1. Add or remove an owner in the safe where a few approves are required
2. Check transaction view in the Queue

**Expected behavior**
The settings transaction should have title depends on the function name

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/b025dd70-48c1-4723-8b7e-2ec8ae061342)

**Smartphone (please complete the following information):**
 - Device: Pixel 4
 - Android Version: 10
 - App Version: dev version 0.4.2 -alpha-debug

**Additional context**

biafra23:
See 0x1C8b9B78e3085866521FE206fa4c1a67F49f153A for an example.

biafra23:
This is, what is missing in TransactionViewHolderFactory.kt to fix this bug:
 
![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/0ea9e87a-b210-48ba-bce9-858d84d1b0aa)

biafra23:
For a fix see branch `bug/716-712`

liliya-soroka:
Verified
app version 1.12.1-76

Issue #706, opened 2020-07-07
[Android] Outgoing transaction(transfer) where safe is a recipient is marked as incoming in Queue


liliya-soroka:
**Describe the bug**
The transaction where the initial safe is a recipient is saved as incoming in Queue. It looks strange that incoming transactions required confirmation

**To Reproduce**
1st situation:
1. Send token from safe to safe X
2. Check transactions list-  Queue

See transaction:
1. load safe 0x1230B3d59858296A31053C1b8562Ecf89A2f888b
2. transaction with nonce 115 or 97

**Expected behavior**
The transfer transaction in such case should be marked as outgoing - https://docs.google.com/document/d/1vhTu5NbB-6m8nE3cqTS3bRONigPO3NA5YkJqNTA6AOk/edit?usp=sharing

**Smartphone (please complete the following information):**
 - Device: Pixel 4
 - Android Version: 10
 - App Version 1.12.1-47

**Additional context**

biafra23:
This ist the tx in question (115):

```
      "safe": "0x1230B3d59858296A31053C1b8562Ecf89A2f888b",
      "to": "0x63704B63Ac04f3a173Dfe677C7e3D330c347CD88",
      "value": "0",
      "data": "0xa9059cbb0000000000000000000000001230b3d59858296a31053c1b8562ecf89a2f888b00000000000000000000000000000000000000000000000000038d7ea4c68000",
      "operation": 0,
      "gasToken": "0x0000000000000000000000000000000000000000",
      "safeTxGas": 33874,
      "baseGas": 0,
      "gasPrice": "0",
      "refundReceiver": "0x0000000000000000000000000000000000000000",
      "nonce": 115,
      "executionDate": null,
      "submissionDate": "2020-07-07T13:47:39.768304Z",
      "modified": "2020-07-07T13:47:39.768304Z",
      "blockNumber": null,
      "transactionHash": null,
      "safeTxHash": "0x8f630a28307189169dea89af92a87f29114c5d8b63c59099aebee03568ee2194",
      "executor": null,
      "isExecuted": false,
      "isSuccessful": null,
      "ethGasPrice": null,
      "gasUsed": null,
      "fee": null,
      "origin": null,
      "dataDecoded": {
        "method": "transfer",
        "parameters": [
          {
            "name": "to",
            "type": "address",
            "value": "0x1230B3d59858296A31053C1b8562Ecf89A2f888b"
          },
          {
            "name": "value",
            "type": "uint256",
[...]

PR #769, opened 2020-07-29
Fix incoming status on confirmations/rejections

biafra23:
Ticket #714 .

- Move calculation of incoming status to repository because we need access to data decoded to calculate it.
- Follow the rules laid out here: https://docs.google.com/document/d/1vhTu5NbB-6m8nE3cqTS3bRONigPO3NA5YkJqNTA6AOk/edit?skip_itp2_check=true **Except** the rules for custom tx because they are in a different issue/ticket and maybe obsoleted by the client service
- Tests added for new `incoming` calculation
- I sneaked in two fixes for missing confirmations text color and settings name (queued txs) 
- Also solves: https://github.com/gnosis/safe-android/issues/706 and  #712 

Before | After
------- | -----
Screenshot_20200729-185226  | Screenshot_20200729-185129

@gnosis/mobile-devs

### DIFF
--- a/data/src/main/java/io/gnosis/data/models/Transaction.kt
+++ b/data/src/main/java/io/gnosis/data/models/Transaction.kt
@@ -37,7 +37,8 @@ sealed class Transaction {
         val value: BigInteger,
         val date: String?,
         val tokenInfo: ServiceTokenInfo?,
-        val nonce: BigInteger?
+        val nonce: BigInteger?,
+        val incoming: Boolean
     ) : Transaction()
 }
 
--- a/data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
+++ b/data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
@@ -35,9 +35,9 @@ class TransactionRepository(
                     is ModuleTransactionDto -> add(custom(transactionDto))
                     is EthereumTransactionDto -> {
                         when {
-                            !transactionDto.transfers.isNullOrEmpty() -> addAll(transactionDto.transfers.map { transfer(it) })
+                            !transactionDto.transfers.isNullOrEmpty() -> addAll(transactionDto.transfers.map { transfer(it, safeInfo) })
                             transactionDto.transfers.isNullOrEmpty() && !transactionDto.data.hexStringNullOrEmpty() -> add(custom(transactionDto))
-                            else -> add(transfer(transactionDto))
+                            else -> add(transfer(transactionDto, safeInfo))
                         }
                     }
                     is MultisigTransactionDto -> {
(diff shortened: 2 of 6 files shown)


data/src/main/java/io/gnosis/data/models/Transaction.kt
```diff
@@ -37,7 +37,8 @@ sealed class Transaction {
         val value: BigInteger,
         val date: String?,
         val tokenInfo: ServiceTokenInfo?,
-        val nonce: BigInteger?
+        val nonce: BigInteger?,
+        val incoming: Boolean
```
jpalvarezl, line 41: Not sure if it's a good idea to add this as part of data model, as the client gateway will return this hopefully with need for mapping. Instead of adding an additional field could we just have it calculated as an extension function? Like `fun Transactions.Transfer.isInconming(): Boolean`? This is a field that to me is part of the UI layer.

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
```diff
@@ -76,35 +76,37 @@ class TransactionRepository(
                 SafeRepository.isSettingsMethod(transactionDto.dataDecoded?.method)
 
     // This is a big assumption for txType == ETHEREUM_TRANSACTION, it was agreed that this can be assumed successful, because only successful TXs trigger events
-    private fun transfer(transferDto: TransferDto): Transaction.Transfer {
-        val tokenInfo = serviceTokenInfo(transferDto)
+    private fun transfer(transfer: TransferDto, safeInfo: SafeInfo): Transaction.Transfer {
```
jpalvarezl, line 79: Here you are forced to to pass all the way down to the lowest level of mapping the `SafeInfo` which, if you did in the UI layer, you could simply call the method. I think it would be a cleaner solution to have it as an extension function like I mentioned on my previous comment.

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
```diff
@@ -76,35 +76,37 @@ class TransactionRepository(
...
-            transaction.value ?: BigInteger.ZERO,
-            transaction.blockTimestamp?.formatBackendDate(),
+            ethereumTransaction.to,
+            ethereumTransaction.from,
+            ethereumTransaction.value ?: BigInteger.ZERO,
+            ethereumTransaction.blockTimestamp?.formatBackendDate(),
             ETH_SERVICE_TOKEN_INFO,
-            null
+            null,
+            ethereumTransaction.isIncoming(safeInfo)
```
jpalvarezl, line 109: This is weird, aren't ethereum transactions generally incoming always? Maybe it is my misunderstanding. I think we should avoid the coupling with `safeInfo` here, just like for the other case.

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
```diff
@@ -253,6 +258,38 @@ class TransactionRepository(
...
+            return true
+        }
+        return false
+    } else {
+        // unknown method
+        return false
+    }
+}
+
+
```
jpalvarezl, line 292: All these methods would be better suited IMO in a mapping that happens closer to the UI. Here we are really coupling transaction requests with safe requests. It will hit us with composability of these objects. Besides, `Transactions.Transfer` should already contain `sender` and `recipient`. If there is a bug there, then the problem is in the mapping, we are maybe assigning the wrong address to the wrong field.

data/src/main/java/io/gnosis/data/models/Transaction.kt
jpalvarezl: After seeing the rest of the PR, the `incoming` calculation should definitely be moved up. The basic mapping should be as clean as possible, meaning, with fewer dependencies as possible (safeInfo and other stuff)

biafra23: We already had a dependency on safeInfo for the status calculation.

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
biafra23: If we want to calculate the incoming status in the UI layer then we need to pass up the dataDecoded (method, from and to ) section. Is that really worth it?

biafra23: Unfortunately sender & recipient do not determine alone whether a transfer is incoming or outgoing. Here is the full description: https://docs.google.com/document/d/1vhTu5NbB-6m8nE3cqTS3bRONigPO3NA5YkJqNTA6AOk/edit?skip_itp2_check=true

data/src/main/java/io/gnosis/data/models/Transaction.kt
biafra23: Why is the status calculation here and not in the UI?

jpalvarezl: The status calculation will come from the gateway too.

biafra23: Should the incoming calculation also be done in the gateway (in the future)?

biafra23: Since we just agreed, to have an incoming field coming from the client gateway at some point. Can this PR be merged and the code for incoming calculation be removed when the field is returned by the gateway?

PR #770, opened 2020-07-30
Dates in local  timezone

### LINKED ISSUE #705: [Android] Time in the app doesn't depend on the user's device time zone and date/time format
**Describe the bug**
Seems that the server or any other time zone and date/time format  is used for the transaction creation/execution time instead of user's device local time zone

**To Reproduce**
1. Go to the device settings
2. Change time zone, date and time format 
3. Come back to the app
4.refresh transactions list 
5. Go to the web app and create new transaction
6. Open transactions list in the mobile app

Current result:
As a user, I am confused that just created transaction has different time from actual for me.
Also, it's influence on the displayed day. Some transactions created 4th of July are marked as created 3rd of July because of using different time zone.
Also, date format yyyy-mm-dd can be confusing for the users with another date format.
We should either user date format from the design (https://app.zeplin.io/project/5e9efc30726e106958cc43de/screen/5eb1234bf944bdb12a3fd89c) or the user's device local date format 

**Expected behavior**
The user's device time zone, time format should be used.
Otherwise, it looks waired 

**Screenshots**

**Smartphone (please complete the following information):**
 - Device: Pixel 4
 - Android Version: 10
 - App Version 1.12.1-47

**Additional context**

biafra23:
Closes #705 .

- Uses date class instead of String
- Adds date parsing
- Add localised date formatting for tx list

@gnosis/mobile-devs


app/build.gradle
```diff
@@ -253,6 +253,9 @@ dependencies {
     implementation "com.google.firebase:firebase-analytics:$versions.firebase_analytics"
     implementation "com.google.firebase:firebase-crashlytics:$versions.firebase_crashlytics"
 //    androidTestImplementation "androidx.navigation:navigation-testing:2.3.0-alpha05" //TODO change to stable when available
+
```
rmeissner, line 256: Afaik this lib is only required if you have string timestamps with timezones, which we don't have, we always have utc timezones from the backend.

biafra23: It is needed to load the timezones database. And we need timezone data to calculate local times from UTC.

app/src/main/java/io/gnosis/safe/utils/DateUtils.kt
```diff
@@ -0,0 +1,12 @@
+package io.gnosis.safe.utils
+
+import org.threeten.bp.Instant
+import org.threeten.bp.ZoneId
+import org.threeten.bp.format.DateTimeFormatter
+import org.threeten.bp.format.FormatStyle
+import java.util.*
+
+fun Date.formatBackendDate(zoneId : ZoneId = ZoneId.systemDefault(), locale: Locale = Locale.getDefault()): String {
+    val customFormatter: DateTimeFormatter = DateTimeFormatter.ofLocalizedDateTime(FormatStyle.MEDIUM).withZone(zoneId).withLocale(locale)
```
biafra23, line 10: This is not a performance problem. I checked that the `customFormatter` is cached. The first call takes between 16ms and 40ms. All subsequent calls take between 0ms and 1ms.

app/src/test/java/io/gnosis/safe/ThreeTenTimezoneDatabaseInit.kt
```diff
@@ -0,0 +1,13 @@
+package io.gnosis.safe
+
+import org.threeten.bp.zone.TzdbZoneRulesProvider
+import org.threeten.bp.zone.ZoneRulesProvider
+
+fun Any.initThreeTen() {
```
biafra23, line 6: This is needed to run unit tests.

data/src/main/java/io/gnosis/data/backend/dto/TransactionDto.kt
```diff
@@ -26,10 +27,10 @@ data class MultisigTransactionDto(
     val nonce: BigInteger,
     val isExecuted: Boolean,
     val isSuccessful: Boolean? = null,
-    val executionDate: String? = null,
-    val submissionDate: String? = null,
-    val creationDate: String? = null,
-    val modified: String? = null,
+    val executionDate: Date? = null,
```
elgatovital, line 30: why java.util.Date and not Instant from threetenbp?

app/src/test/java/io/gnosis/safe/ThreeTenTimezoneDatabaseInit.kt
elgatovital: is there any specific reason for having this as an extension for Any?

data/src/main/java/io/gnosis/data/backend/dto/TransactionDto.kt
biafra23: 1. AFAICT the converter class dosen't support threeten classes
2. This way the timezone handling stuff can be local to the app and doesn't need to be in the data module

app/src/test/java/io/gnosis/safe/ThreeTenTimezoneDatabaseInit.kt
biafra23: The resource loader needs a class (to get the class loader).

app/build.gradle
rmeissner: Afaik this is only if the timestamp contains a timezone else you can use the timezone of the device (and for that the lib is not required). I am fine if you want to use the library, but everything that we want to do right now (e.g. display a utc timestamp in the device local time) can be done without.

Also the lib has a warning that the development is winding down and that a new feature should be used. Was this evaluated?

biafra23: Desugaring seems to work and is implemented in b837a17ae7d09b36f417b87be5ffb5cadb581ab9

app/src/test/java/io/gnosis/safe/ThreeTenTimezoneDatabaseInit.kt
biafra23: Since this is not needed with desugaring I will resolve this conversation.

PR #783, opened 2020-08-06
Client Gateway integration tests

### LINKED ISSUE #780: Transaction List test coverage
Improve test coverage for `TransactionRepository`

biafra23:
Closes #780 .

- Adding tests for TransactionRepository

@gnosis/mobile-devs

### DIFF
--- a/data/src/test/java/io/gnosis/data/repositories/TransactionRepositoryTest.kt
+++ b/data/src/test/java/io/gnosis/data/repositories/TransactionRepositoryTest.kt
@@ -28,19 +28,69 @@ import org.junit.Assert.assertEquals
 import org.junit.Assert.assertFalse
 import org.junit.Assert.assertTrue
 import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.Parameterized
 import pm.gnosis.crypto.utils.asEthereumAddressChecksumString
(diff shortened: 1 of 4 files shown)


data/src/test/java/io/gnosis/data/repositories/TransactionRepositoryTest.kt
```diff
@@ -46,322 +117,106 @@ class TransactionRepositoryTest {
...
+    fun `getTransactions (unknown txInfo) should throw exception Transfers`() = runBlockingTest {
+        val pagedResult = listOf(
+            buildGateTransactionDto(txInfo = object : TransactionInfo {
+                override val type: GateTransactionType
+                    get() = GateTransactionType.Transfer
+            })
+        )
+        coEvery { gatewayApi.loadTransactions(any()) } returns Page(1, null, null, pagedResult)
+
+        transactionRepository.getTransactions(defaultSafeAddress)
```
jpalvarezl, line 201: If we use `@Test(expected="...")` we can't really verify what's internally call within the mock, as the method exists early on the throw. I am much more inclined to to use `runCatching` then have additionally a `coVerify` block where we see that we don't call a method more times than we should, etc. And we still can look into the exception.

Issue #573, opened 2020-04-23
Confirmation list for tx details


elgatovital:
![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/c75c32fb-99ec-4af0-934e-c82c11ad6bdd)

biafra23:
Discussion on slack: https://gnosisinc.slack.com/archives/C63LQTGNB/p1588837780117500

elgatovital:
https://gnosisinc.slack.com/archives/C63LQTGNB/p1588837780117500

jpalvarezl:
@rmeissner @tschubotz So rejected transactions as far as we've discussed, are transactions with value, data set to 0 that have the same nonce as the transaction that is being attempted to be cancelled. How would this be reflected on the list, say we have a transaction coming up ahead that is meant to cancel something that is shown as pending?

tschubotz:
This ticket is about the confirmation list but your question is about the transaction list, correct?

So are you asking how this should be displayed on the transaction list? When there is already a transaction pending ( i.e. it's being executed according to https://github.com/gnosis/safe/wiki/Safe-transaction-states) but at the same time there is a competing cancel tx coming in?

rmeissner:
As far as I understood iOS is using the data that is returned in the tx list. So only if the list is refreshed would the confirmation change. @DmitryBespalov correct me if I am wrong :D else I would go for pull to refresh for now as we do not expect that these change often and you would need to check if executed (since for executed tx it makes no sense to poll).

tschubotz:
I agree to Richard. Let's not add some kind of polling yet.
I would use the data from the list that we have.
However I think there should be the option to "pull to refresh" on the tx details as we discussed in https://github.com/gnosis/safe/wiki/Network-requests-and-no-internet-connection

liliya-soroka:
Verified
app version 1.12.1 -93

PR #785, opened 2020-08-09
Add custom views to display execution details of transaction

elgatovital:
Handles #573

- Custom view for showing list of confirmaitons (steps of execution)
- Custom view for showing time info data
- Loading tx details through client gateway using tx id 
- Propagate tx id to transfer object
- Pass tx id to tx details fragment
- Load / Reload tx details 
- Show confirmations, created at, executed at, advanced item, view on etherscan

Testing will be done in a separate pr after we refine structure of the tx details object

Incoming ETH | Outgoing Transfer
-------------- | ------------------
Screenshot_20200811-093722 | Screenshot_20200811-093647

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
```diff
@@ -13,10 +14,17 @@ class TransactionDetailsViewModel
     appDispatchers: AppDispatchers
 ) : BaseStateViewModel<TransactionDetailsViewState>(appDispatchers) {
 
-    override fun initialState() = TransactionDetailsViewState(null, true)
+    override fun initialState() = TransactionDetailsViewState(null, ViewAction.Loading(true))
+
+    fun loadDetails(txId: String) {
+        safeLaunch {
+            val txDetails = transactionRepository.getTransactionDetails(txId)
```
elgatovital, line 21: here error handling is missing -> can be added in next pr. Currently, if something happens during tx details load ui won't be updated and spinner will continue to spin

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
```diff
@@ -77,6 +112,24 @@ class TransactionRepository(
...
+
+            is Creation -> Transaction.Creation(
+                id = id,
+                confirmations = null,
+                status = TransactionStatus.SUCCESS
+            )
+
+            is Unknown -> Transaction.Custom(
+                id = id,
+                address = "0x00".asEthereumAddress()!!,
```
elgatovital, line 124: why "0x00" and not BigInteger.ZERO? Solidity.Address(BigInteger.ZERO). I think it is more explicit and easier to read

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
```diff
@@ -48,9 +55,35 @@ class TransactionRepository(
...
+                getExecutor(it.txInfo),  //(it.txInfo as? Transfer)?.sender, // TODO: handle other transfer type
+                it.txData,
+                it.detailedExecutionInfo
+            )
+        }
+
+    private fun getExecutor(txInfo: TransactionInfo): Solidity.Address =
+         when (txInfo) {
+            is Custom -> txInfo.to // safe address ?
+            is SettingsChange -> "0x1234".asEthereumAddress()!! // safe address
```
elgatovital, line 77: what are those values: "0x1234", ...? And they should be extracted to constants with meaningful names

data/src/main/java/io/gnosis/data/models/TransferDetails.kt
```diff
@@ -0,0 +1,61 @@
+package io.gnosis.data.models
+
+import io.gnosis.data.backend.dto.DetailedExecutionInfo
+import io.gnosis.data.backend.dto.TxData
+import pm.gnosis.model.Solidity
+import java.util.*
+
+//FIXME: (WIP) revise structure
+// Designs:
+// incoming Transfer https://zpl.io/V1KKZq5 (with data section)
```
elgatovital, line 10: I don't think we should comments with links to design specs in code. And also we should not have comments about design specs at all, just code related comments

data/src/main/java/io/gnosis/data/models/Transaction.kt
```diff
@@ -40,6 +45,12 @@ sealed class Transaction {
         val nonce: BigInteger?,
         val incoming: Boolean
     ) : Transaction()
+
+    data class Creation(
+        override val id: String,
+        override val status: TransactionStatus,
+        override val confirmations: Int?
```
biafra23: I presume this will always be null?

data/src/main/java/io/gnosis/data/models/Transaction.kt
```diff
@@ -40,6 +45,12 @@ sealed class Transaction {
         val nonce: BigInteger?,
         val incoming: Boolean
     ) : Transaction()
+
+    data class Creation(
+        override val id: String,
+        override val status: TransactionStatus,
```
biafra23: This will always be successful?

data/src/main/java/io/gnosis/data/models/TransferDetails.kt
biafra23: Yes you're right.
Those were work in progress comments. While I was checking the data from the client gateway has everything we need in the layouts. I wanted to remove them when the layouts are done. But if you want, I can remove them right now.

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
biafra23: These are dummy values for now. They do not have any meaning. There is a TODO next to the call to getExecutor which will be covered in the next PR.

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
biafra23: Fixed in 9e9bc0cf8afce0f6a2b6a9a4774753181a84ffc1

data/src/main/java/io/gnosis/data/models/TransferDetails.kt
elgatovital: sure, when layouts are done, let's clean it up

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
jpalvarezl: The `BaseStateViewModel` emits errors automatically AFAIK with `safeLaunch`. The problem is that in line https://github.com/gnosis/safe-android/blob/bd5323fd5162ca4db94f61574ef5b3f2a3f51110/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt#L58 The case `viewAction is ShowError` is not handled at all.

app/src/main/java/io/gnosis/safe/ui/transactions/details/view/TxConfirmationsView.kt
```diff
@@ -0,0 +1,203 @@
...
+                TransactionStatus.AWAITING_EXECUTION -> {
+                    addExecutionStep(TxExecutionStep.Type.EXECUTE_READY)
+                }
+                TransactionStatus.SUCCESS -> {
+                    if(executor != null) {
+                        addExecutionStep(TxExecutionStep.Type.EXECUTE_DONE)
+                        addAddressItem(executor)
+                    } else {
+                        // fail silently
+                        addExecutionStep(TxExecutionStep.Type.EXECUTE_READY)
```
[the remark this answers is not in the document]
jpalvarezl: Also, is it correct to classify as `EXECUTE_READY`? Maybe we shouldn't make executor `nullable` at all.

app/src/main/java/io/gnosis/safe/ui/transactions/details/view/TxConfirmationsView.kt
```diff
@@ -0,0 +1,203 @@
...
+    private fun addAddressItem(address: Solidity.Address) {
+        val addressItem = AddressItem(context)
+        val layoutParams = LayoutParams(LayoutParams.MATCH_PARENT, dpToPx(ADDRESS_ITEM_HEIGHT))
+        layoutParams.setMargins(dpToPx(ADDRESS_ITEM_MARGIN_LEFT), 0, 0, dpToPx(MARGIN_VERTICAL))
+        addressItem.layoutParams = layoutParams
+        addressItem.address = address
+        addView(addressItem)
+    }
+
+    fun clear() {
```
jpalvarezl, line 94: do you need access from outside? I would either inline this method or make it `private`.

app/src/main/java/io/gnosis/safe/ui/transactions/details/view/TxConfirmationsView.kt
```diff
@@ -0,0 +1,203 @@
...
+            private set
+
+        enum class Type {
+            CREATED,
+            CONFIRMED,
+            CANCELED,
+            FAILED,
+            EXECUTE_WAITING,
+            EXECUTE_READY,
+            EXECUTE_DONE
```
jpalvarezl, line 140: `EXECUTE_WAITING`, `EXECUTE_READY` and `EXECUTE_DONE` don't sound very gramatical. It is not important, but also, don't we need to distinguish between awaiting execution and awaiting confirmations?

data/src/main/java/io/gnosis/data/backend/dto/GateTransactionDetails.kt
```diff
@@ -0,0 +1,52 @@
+package io.gnosis.data.backend.dto
+
+import io.gnosis.data.models.TransactionStatus
+import pm.gnosis.model.Solidity
+import java.math.BigInteger
+
+//FIXME: (WIP) revise structure
```
jpalvarezl, line 7: probably can get rid of this FIXME

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
```diff
@@ -37,12 +50,42 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
...
+    }
+
+    private fun updateUi(txDetails: TransferDetails?, isLoading: Boolean) {
+        binding.refresh.isRefreshing = isLoading
+        txDetails?.let {
+            binding.content.visible(true)
+            binding.txConfirmations.setExecutionData(
+                status = txDetails.txStatus,
+                confirmations = (txDetails.detailedExecutionInfo as? MultisigExecutionDetails)?.confirmations?.map { it.signer } ?: listOf(),
+                threshold = (txDetails.detailedExecutionInfo  as? MultisigExecutionDetails)?.confirmationsRequired ?: 0,
```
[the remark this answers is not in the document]
biafra23: I agree. This was a dirty quick workaround. Not even a fix.

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
```diff
@@ -37,12 +50,42 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
...
-
+            when (val viewAction = state.viewAction) {
+                is BaseStateViewModel.ViewAction.Loading -> updateUi(state.txDetails, viewAction.isLoading)
+            }
         })
     }
+
+    override fun onResume() {
+        super.onResume()
+        viewModel.loadDetails(txId)
```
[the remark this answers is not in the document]
elgatovital: if we have pending tx's onResume is good place to update tx details - maybe there are some additional confirmations or other changes. Generally, we should also distinguish btw loading and updating tx data. For creational transaction, there is no sense to have updates or pull to refresh. This can be handled in a separate pr.

app/src/main/java/io/gnosis/safe/ui/transactions/details/view/TxConfirmationsView.kt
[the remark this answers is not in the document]
elgatovital: this should never happen. If tx status is success - there should be an executor. But in other cases executor can be null. We can log it.

app/src/main/java/io/gnosis/safe/ui/transactions/details/view/TxConfirmationsView.kt
elgatovital: this can be potentially used from outside. It's part of public api of the view.

app/src/main/java/io/gnosis/safe/ui/transactions/details/view/TxConfirmationsView.kt
elgatovital: we do - EXECUTE_WAITING represents awaiting confirmations, EXECUTE_READY represents awaiting execution. Those names represent ui better. In each case we have execute-label

elgatovital: it's not about grammatical sounding. They have same prefix and are used in place of the same execution step element.

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
elgatovital: this was not that clear. I would place TODO into getExecutor itself and not near it's invocation

elgatovital: no action required now

app/src/main/java/io/gnosis/safe/ui/transactions/details/view/TxConfirmationsView.kt
jpalvarezl: Why not use the TransactionStatus enum directly then?

elgatovital: it does not map 1:1. You don't have created status. Or confirmed. This is very ui specific

PR #788, opened 2020-08-11
TX details screen

### LINKED ISSUE #574: Tx details screen: basic ui
screen: https://zpl.io/bPj0NQD
Layout except confirmation/signature list & status header
Navigation to advanced is excluded
All addresses lead to etherscan

Bundle arg: safe tx hash
Load tx data using safe tx hash on entering

biafra23:
Closes #574  .

- Move common layout out of ViewStub for TXInfo types (Custom, Settings, Transfer)
- Copy DTO details model to Domain model almost 1:1
- 

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
```diff
@@ -70,25 +76,77 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
...
-            binding.executed.value = txDetails.executedAt?.formatBackendDate() ?: ""
-            binding.etherscan.setOnClickListener {
-                requireContext().openUrl(
-                    getString(
-                        R.string.etherscan_transaction_url,
-                        txDetails.txHash
+        binding.content.visible(true)
+
+        
+        // TODO: How can we use viewBindings for the ViewStubs after they have been replaced?
```
rmeissner, line 85: what about `TxDetailsTransferBinding.bind(view)`?

Also theoretically you only need one stub as you can change the target layout with setLayoutResource

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
```diff
@@ -107,10 +111,11 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
                 }
             }
             is SettingsChangeDetails -> {
-
-                val viewStub = rootView.findViewById<ViewStub>(R.id.stub_setttings_change).inflate()
-                viewStub.visibility = View.VISIBLE
-                val txDetailsSettingsChangeBinding = TxDetailsSettingsChangeBinding.bind(viewStub)
+                val viewStub = binding.stubSetttingsChange
```
rmeissner, line 114: typo in name

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
```diff
@@ -18,13 +19,19 @@ class TransactionDetailsViewModel
 
     fun loadDetails(txId: String) {
         safeLaunch {
-            val txDetails = transactionRepository.getTransactionDetails(txId)
-            updateState { TransactionDetailsViewState(txDetails, ViewAction.Loading(false)) }
+
+            kotlin.runCatching {
```
rmeissner, line 23: This doesn't really make sense, as the `safeLaunch` has a catcher build in (https://github.com/gnosis/safe-android/blob/master/app/src/main/java/io/gnosis/safe/ui/base/BaseStateViewModel.kt#L50 , this can even be overwritten)

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
```diff
@@ -70,26 +79,120 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
...
+                        getString(
+                            R.string.etherscan_transaction_url,
+                            txDetails.txHash
+                        )
                     )
+                }
+                val txType = if (txDetails.incoming == true) TxStatusView.TxType.TRANSFER_INCOMING else TxStatusView.TxType.TRANSFER_OUTGOING
+                txDetailsTransferBinding.txStatus.setStatus(txType, txDetails.txStatus)
+
+                if (txDetails.txData != null) {
```
rmeissner, line 117: This view should not show the txData section (see https://zpl.io/2GJA7RJ)

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
```diff
@@ -70,26 +79,120 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
...
+                txDetailsSettingsChangeBinding.etherscan.setOnClickListener {
+                    requireContext().openUrl(
+                        getString(
+                            R.string.etherscan_transaction_url,
+                            txDetails.txHash
+                        )
+                    )
+                }
+                txDetailsSettingsChangeBinding.txStatus.setStatus(TxStatusView.TxType.MODIFY_SETTINGS, txDetails.txStatus)
+                if (txDetails.txData != null) {
```
rmeissner, line 150: This section should also not show the txData section (see https://zpl.io/VYgpw7k)

data/src/main/java/io/gnosis/data/models/TransferDetails.kt
```diff
@@ -5,6 +5,8 @@ import io.gnosis.data.backend.dto.TxData
 import pm.gnosis.model.Solidity
 import java.util.*
 
+interface TransactionDetails
```
rmeissner, line 8: Why do we not use a similar model structure as in the client gateway. A lot of information is duplicated here (which also shows in the view holder and mapping as a lot of code is duplicated).

The client gateway models were created with the UI design in mind. The top section which displays the transaction type specific details (transfer, settings or custom) is polymorphic and there you can put the specific parts for that into different layout. The other parts of the response are mapped to a section in the UI (status -> show in status section, if executed timestamp is present -> show in executed section, if txHash is present -> show Etherescan link, ...). The whole purpose of the client gateway is that the clients should be able to use nearly the same models as the backend returns and that in most cases no additional mapping is required. If this is not the case we should discuss the response data and check if it can be optimized.

data/src/main/java/io/gnosis/data/models/TransferDetails.kt
```diff
@@ -44,17 +48,19 @@ data class SettingsChangeDetails(
...
     val executedAt: Date?,
+    val executor: Solidity.Address,
     val detailedExecutionInfo: DetailedExecutionInfo?,
     val txData: TxData? // contains data decoded
-)
+) : TransactionDetails
 
 // Safe created
-data class CreationChangeDetails(
+data class CreationDetails(
```
rmeissner, line 57: I would display the creation details in the same screen/ models as the other transactions, as this is a very special transaction with a compeltely different design ... also this information will never be returned by the gateway.

data/src/main/java/io/gnosis/data/models/TransferDetails.kt
```diff
@@ -5,6 +5,8 @@ import io.gnosis.data.backend.dto.TxData
 import pm.gnosis.model.Solidity
```
rmeissner, line 5: Why is this file called `TransferDetails`?

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
```diff
@@ -55,29 +57,85 @@ class TransactionRepository(
...
+                        txHash = it.txHash,
+                        txData = it.txData,
+                        txStatus = it.txStatus,
+                        executedAt = null,
+                        detailedExecutionInfo = null,
+                        createdAt = null,
+                        executor = Solidity.Address(BigInteger.ZERO)
+                    )
+                }
+                else -> CustomDetails(
```
rmeissner, line 128: This and the `Unknown` branch are identical, why not collapse them?

app/src/main/res/layout/fragment_transaction_details.xml
```diff
@@ -59,68 +59,36 @@
...
-                    android:layout_width="wrap_content"
-                    android:layout_height="@dimen/item_setting_openable_height"
-                    android:drawableRight="@drawable/ic_link_green_24dp"
-                    android:drawablePadding="6dp"
-                    android:gravity="left|center_vertical"
-                    android:text="@string/tx_details_view_on_etherscan"/>
+                android:orientation="vertical"
+                tools:visibility="visible">
+            <!-- Put Transfer ViewStub here  -->
+            <ViewStub
```
rmeissner, line 65: View use a different stub for each tx detail layout ... this means if we add in the future more layout we need to add more and more stubs, not sure if this is the best idea (btw I like the ViewStub approach ;) )

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
```diff
@@ -55,29 +57,85 @@ class TransactionRepository(
...
+                        executor = (it.txInfo as? Transfer)?.sender!!, // TODO: handle other transfer type
+                        txData = it.txData,
+                        detailedExecutionInfo = it.detailedExecutionInfo,
+                        incoming = (it.txInfo as? Transfer)?.direction == TransactionDirection.INCOMING
+                    )
+                }
+                is Custom -> {
+                    CustomDetails(
+                        txHash = it.txHash,
+                        executedAt = it.executedAt?.let { date ->
```
rmeissner, line 78: Why no use `toDate` here to as above?

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
biafra23: What about: https://zpl.io/bJmODxx (Incoming Transfer)

I think the Designs are not enough as a specification.

rmeissner: Incoming custom transfers don't exist

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
```diff
@@ -70,26 +79,105 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
...
+                            txDetails.txHash
+                        )
+                    )
+                }
+
+                txDetailsCustomBinding.txStatus.setStatus(TxStatusView.TxType.CUSTOM, txDetails.txStatus)
+
+                if (txDetails.txData != null) {
+                    txDetailsCustomBinding.txData.visible(true)
+                    txDetails.txData?.hexData?.let { txDetailsCustomBinding.txData.setData(it,hexStringByteSize(it)) } // TODO: calculate right size or pass it through
```
rmeissner, line 171: The size is part of the custom details

app/src/main/res/layout/fragment_transaction_details.xml
biafra23: I moved the common sections out of the ViewStubs.

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
biafra23: I am not sure what to do here. 
Can I just remove the surrounding block and the error handling because we're logging it already in the base class?
The result qwould be this
```
    fun loadDetails(txId: String) {
        safeLaunch {
            val txDetails = transactionRepository.getTransactionDetails(txId)
            updateState { TransactionDetailsViewState(txDetails, ViewAction.Loading(false)) }
        }
    }
```

data/src/main/java/io/gnosis/data/models/TransferDetails.kt
biafra23: Legacy. I renamed it

app/src/main/java/io/gnosis/safe/ui/transactions/details/view/TxStatusView.kt
```diff
@@ -0,0 +1,81 @@
...
+            }
+        }
+    }
+
+    enum class TxType(@DrawableRes val iconRes: Int, @StringRes val titleRes: Int) {
+        TRANSFER_INCOMING(R.drawable.ic_arrow_green_10dp, R.string.tx_status_type_transfer_incoming),
+        TRANSFER_OUTGOING(R.drawable.ic_arrow_red_10dp, R.string.tx_status_type_transfer_outgoing),
+        MODIFY_SETTINGS(R.drawable.ic_settings_change_14dp, R.string.tx_status_type_modify_settings),
+        CUSTOM(R.drawable.ic_code_16dp, R.string.tx_status_type_custom),
+        CREATION(R.drawable.ic_settings_change_14dp, R.string.tx_status_type_creation)
```
jpalvarezl, line 79: Looks good 👍

PR #790, opened 2020-08-14
Add advanced tx details screen

elgatovital:
- Add advanced tx details screen
- Adjust nav graph

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
```diff
@@ -141,6 +142,11 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
             }
         }
 
+        binding.advanced.setOnClickListener {
+            //TODO: pass arguments
```
jpalvarezl, line 146: This is where we wire @biafra23 changes into yours?

app/src/main/java/io/gnosis/safe/ui/transactions/details/view/ValueItem.kt
```diff
@@ -9,7 +9,7 @@ import io.gnosis.safe.R
 import io.gnosis.safe.databinding.ViewTimeInfoItemBinding
 import timber.log.Timber
 
-class TimeInfoItem @JvmOverloads constructor(
```
jpalvarezl, line 12: I am a bit confused by this name. Is this what we use for dates as well as other stuff too? Like a label + some value?

elgatovital: You are absolutely right, that's why I have renamed it to ValueItem. Maybe LabeledValueItem would be more descriptive? The naming contest is officially open :)

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
elgatovital: yes

Issue #568, opened 2020-04-23
View transaction details


elgatovital:
### Transaction details:
- User sees the following details:
  - Sender (address & identicon, tapping copies the address to clipboard)
  - Recipient (address & identicon, tapping copies the address to clipboard)
  - ETH value
  - Data (Show only number of bytes by default, user can view and copy full data if desired)
  - Submission date
  - Execution date
  - Type of operation (call or delegatecall)
  - Transaction hash (User can copy it)
  - Status (See section below)
  - ~Network fee~ (Leave this one out since there could be different types and we don't need this complexity right now, cf. [Slack thread](https://gnosisinc.slack.com/archives/G9GEEHY79/p1585744219049200))
- User can open the tx on etherscan.
- If a transaction is an outgoing tx show:
  - Which of the owners confirmed/rejected

### Transaction decoding:
- Decode the following transactions by showing details:
  - ERC20 token transfers
  - Safe settings changes (addOwner, removeOwner, swapOwner, changeMastercopy)

#### Transaction status
https://docs.google.com/document/d/1oV3hdlBvbs5VB-j-4mZVWOZRekdFNBWIv92W4uRapto/edit#

elgatovital:
see web client for confirmation list. Ask them for specification

elgatovital:
What's missing:

- NFT details (action part for NFT's)
- Small layout adjustments
- Logo for ERC20 (not loaded, not part of the layout)
- Placeholder when data can not be loaded

liliya-soroka:
Verified
app version 1.12.1 -98

PR #791, opened 2020-08-16
Basic TxTransferActionView and TxSettingsActionView

### LINKED ISSUE #574: Tx details screen: basic ui

biafra23:
Regharding #568   .

- Add top view to tx details

New mastercopy | Outgoing ERC20 transfer | Custom tx 
------ | ----- | ------ 
Screenshot_20200820-123352 | Screenshot_20200820-123422 | Screenshot_20200820-123449

 

Incoming ERC721 transfer | threshold change
 ------- | --------- 
Screenshot_20200820-124304 | Screenshot_20200820-124331

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewModel.kt
```diff
@@ -416,21 +415,12 @@ class TransactionListViewModel
             confirmationsTextColor = if (thresholdMet) R.color.safe_green else R.color.medium_grey,
             confirmationsIcon = if (thresholdMet) R.drawable.ic_confirmations_green_16dp else R.drawable.ic_confirmations_grey_16dp,
             nonce = transaction.nonce.toString(),
-            version = version,
+            version = version ?: "Unknown",
```
elgatovital, line 418: hardcoded string alert :)

app/src/main/java/io/gnosis/safe/ui/transactions/details/view/TxSettingsActionView.kt
```diff
@@ -0,0 +1,127 @@
...
+            METHOD_ENABLE_MODULE, METHOD_DISABLE_MODULE -> {
+                addAddressItem(params.getValueByName("module")?.asEthereumAddress()!!)
+            }
+        }
+    }
+
+    private fun addStringItem(text: String) {
+        val addressItem = TextView(context)
+        val layoutParams = LinearLayout.LayoutParams(LinearLayout.LayoutParams.MATCH_PARENT, dpToPx(TxTransferActionView.ADDRESS_ITEM_HEIGHT))
+        layoutParams.setMargins(dpToPx(TxTransferActionView.ADDRESS_ITEM_MARGIN_LEFT), 0, 0, dpToPx(TxTransferActionView.MARGIN_VERTICAL))
```
elgatovital, line 90: either there should be a global margin for all items inside this view, or string item should have it's own. Using margin from a different view doesn't look good. You could define margin inside of the companion object or have then in dimens.xml

app/src/main/java/io/gnosis/safe/ui/transactions/details/view/TxSettingsActionView.kt
```diff
@@ -0,0 +1,127 @@
...
+        addressItem.layoutParams = layoutParams
+        addressItem.address = address
+        addressItem.label = label
+        addView(addressItem)
+    }
+
+    private val settingsMethodTitle = mapOf(
+        METHOD_ADD_OWNER_WITH_THRESHOLD to "Add owner:",
```
elgatovital, line 116: these should be defined in strings.xml

app/src/main/java/io/gnosis/safe/ui/transactions/details/view/TxTransferActionView.kt
```diff
@@ -0,0 +1,89 @@
...
+    private val binding by lazy { ViewTxActionBinding.inflate(LayoutInflater.from(context), this) }
+    private fun clear() {
+        removeAllViews()
+    }
+
+    fun setActionInfo(txInfo: TransactionInfo) {
+
+        clear()
+
+        // Step 1
```
elgatovital, line 38: these comments are not adding any value

elgatovital:
image
missing bottom margin

app/src/main/res/layout/view_tx_action.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<merge xmlns:android="http://schemas.android.com/apk/res/android"
```
elgatovital, line 2: you don't need this layout

elgatovital:
it would be better to rebase this pr on top of master branch

app/src/main/java/io/gnosis/safe/ui/transactions/details/view/AmountView.kt
```diff
@@ -0,0 +1,70 @@
...
+                        transferInfo.logoUri
+                    }
+                    is TransferInfo.Erc721Transfer -> {
+                        transferInfo.logoUri
+                    }
+                    else -> {
+                        "local::ethereum"
+                    }
+                }
+                binding.logo.loadTokenLogo(logoUri)
```
jpalvarezl, line 59: There is a lot of code duplication here that I think could be avoided, if we added a method in `TransferInfo` where we can do the type check and return the correct field. Not terribly important though, but then we could really take advantage of polymorphism.

jpalvarezl: This `Amount` view should be as "dumb" as possible, in a sense that it receives an amount and logo to display. Maybe that would simplify things a bit.

app/src/main/java/io/gnosis/safe/ui/transactions/details/view/TxSettingsActionView.kt
```diff
@@ -0,0 +1,113 @@
...
+            }
+            METHOD_SWAP_OWNER -> {
+                params.getValueByName("oldOwner")?.asEthereumAddress()?.let { addAddressItem(it) }
+                settingsMethodTitle[METHOD_ADD_OWNER_WITH_THRESHOLD]?.let { addStringItem(it) }
+                params.getValueByName("newOwner")?.asEthereumAddress()?.let { addAddressItem(it) }
+            }
+            METHOD_ENABLE_MODULE, METHOD_DISABLE_MODULE -> {
+                params.getValueByName("module")?.asEthereumAddress()?.let { addAddressItem(it) }
+            }
+        }
```
jpalvarezl, line 81: I am wondering if these custom views should be aware of the domain objects. This could be maybe later discussed and potentially refactored if we agree on it. I would say it is a more common pattern to pass text, assets, etc. to views and they just render it, rather than having decision trees inside them.

app/src/main/java/io/gnosis/safe/ui/transactions/details/view/AmountView.kt
biafra23: I totally agree. The View should be as dumb as possible. If only because it cannot be unit tested. I don't know why I hesitated to put helper methods into TransferInfo/TransactionInfo

app/src/main/java/io/gnosis/safe/ui/transactions/details/view/TxSettingsActionView.kt
```diff
@@ -0,0 +1,120 @@
...
+
+    fun setActionInfo(txInfo: TransactionInfo) {
+        clear()
+        val settingsChange = txInfo as TransactionInfo.SettingsChange
+        settingsMethodTitle[settingsChange.dataDecoded.method]?.let { addStringItem(it) }
+
+        val params = settingsChange.dataDecoded.parameters
+        when (settingsChange.dataDecoded.method) {
+            METHOD_CHANGE_MASTER_COPY -> {
+                val mainCopy = params.getValueByName("_masterCopy")?.asEthereumAddress()
```
elgatovital, line 48: as mentioned view should be dumb, get everything it needs, and not have this complex parsing logic

elgatovital (approved):

PR #798, opened 2020-08-19
Fix logic for showing correct execution step

elgatovital:
Waiting for timestamps for confirmations from client gw. Till then it's a draft 

- Fix tx confirmations list

@gnosis/mobile-devs


biafra23 (approved):
LGTM 👍

Issue #775, opened 2020-08-04
Use same signing key for app distribution & playstore


elgatovital:
Using same signing key will enable testers to get updates from playstore.
Keep in mind: we might want to switch to app bundles at some point to support dynamic features. In this case we need to use two keys: upload key & signing key (used by google to sign apk's generated from app bundle)

- [x] Create / Update keystores
- [x] Adjust buildkite pipelines

Dependencies:
playstore app project has to be created in order to configure upload & signing keys

liliya-soroka:
@tschubotz , just to confirm the expected result we agreed on:
The users with beta version will need to uninstall beta version and install the google play public release 2.0.0 version .
It will not be possible to update beta version to the public release version without uninstalling

tschubotz:
That's correct.

liliya-soroka:
Verified
app version 1.12.1 -98

PR #804, opened 2020-08-20
Use env variable to specify keystore for signing

elgatovital:
Handles #775  

- Use signing key for app distribution uploads
- Use upload key for github uploads

Buildkite script was adjusted to select correct keystore.

@gnosis/mobile-devs


app/build.gradle
```diff
@@ -58,7 +58,7 @@ android {
             storeFile file("../debug.keystore")
         }
         release {
-            storeFile file(getKey("SIGNING_KEYFILE_PATH", "../gnosis-upload.jks"))
```
biafra23, line 61: Just to clarify: This just changes the environment variable to look for the keytore file name in

elgatovital: yes

biafra23 (approved):
LGTM 👍

Issue #697, opened 2020-07-02
Handle errors & reloads


elgatovital:
Notify user about internet connection & data loading related issues

see API errors handling doc: https://docs.google.com/spreadsheets/d/1U_cM0Tg3WecJ8XTcoP2VK1ZiCXytAIlm0Y0kdorp4J4/edit#gid=0

network requests & no internet connection: https://github.com/gnosis/safe/wiki/Errors%2C-Network-requests-and-no-internet-connection

notification example screen: https://app.zeplin.io/project/5e8f27b9963e55bb79b13634/screen/5ebbb3f2036ca578104c1a81

slack discussion: https://gnosisinc.slack.com/archives/C63LQTGNB/p1592296491004400

elgatovital:
Split into no internet error & other errors

elgatovital:
error state for coins: https://zpl.io/VDJA5P3

liliya-soroka:
Verified
app version 1.12.1 - 105

PR #823, opened 2020-08-25
Handle connectivity & error issues

elgatovital:
Handles #697 

- Get connectivity updates
- Notify user if offline
- 

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/base/BaseStateViewModel.kt
```diff
@@ -16,7 +20,36 @@ data class AppDispatchers(
...
+                    }
+                }
+            }
+            val builder = NetworkRequest.Builder()
+            builder
+                .addTransportType(NetworkCapabilities.TRANSPORT_CELLULAR)
+                .addTransportType(NetworkCapabilities.TRANSPORT_WIFI)
+            it.registerNetworkCallback(builder.build(), networkCallback)
+        }
+    }
```
jpalvarezl, line 51: Oh man not sure this is the best decision. I would absolutely **not** couple the `ConnectivityManager` to our BaseViewModel class. This doesn't match our modelling at all. It feels put here out of convenience rather than architectural design.

app/src/main/java/io/gnosis/safe/ui/assets/coins/CoinsFragment.kt
```diff
@@ -51,6 +52,16 @@ class CoinsFragment : BaseViewBindingFragment<FragmentCoinsBinding>() {
                         when (action) {
                             is BaseStateViewModel.ViewAction.ShowError -> handleError(action.error)
                             is UpdateBalances -> adapter.setItems(action.newBalances)
+                            is Connectivity -> {
```
jpalvarezl, line 55: This would have to be duplicated everywhere.

app/src/main/java/io/gnosis/safe/ui/base/BaseStateViewModel.kt
```diff
@@ -16,7 +20,36 @@ data class AppDispatchers(
     val background: CoroutineDispatcher = Dispatchers.IO
 )
 
-abstract class BaseStateViewModel<T>(private val dispatchers: AppDispatchers) : ViewModel() where T : BaseStateViewModel.State {
+abstract class BaseStateViewModel<T>(private val dispatchers: AppDispatchers, private val connectivityManager: ConnectivityManager? = null) :
+    ViewModel() where T : BaseStateViewModel.State {
+
+    private lateinit var networkCallback: ConnectivityManager.NetworkCallback
```
jpalvarezl, line 26: Couldn't we add our own wrapper to `ConnectivityManager`, something like a `ConnectivityMonitor` or `ConnectivityObserver` that we consume in the `MainActivity`? Or even everywhere, that doesn't matter so much. There are times that we would like to have `viewModels` that aren't necessarily `BaseStateViewModel` which have simpler logic, but may need to also monitor the network state. This is again an argument against this sort of code coupling.

jpalvarezl (changes requested):
We should not be coupling the `BaseStateViewModel` with the `ConnectivityManager`. This , in my opinion will lead to forceful inheritance of this viewModel in every case we need to poll the `ConnectivityManager` and simply mixes responsibilities and deteriorates our code abstraction.

app/src/main/java/io/gnosis/safe/ui/assets/coins/CoinsFragment.kt
jpalvarezl: Maybe that's not that terrible... but still. We should find a solution that doesn't mix responsibilities like this.

app/src/main/java/io/gnosis/safe/ui/base/BaseStateViewModel.kt
elgatovital: not every screen needs connectivity updates, so I would not place observers in MainActivity. Also, it is fine to not provide connectivity manager and there will be no updates. Adding wrapper will not change much - it will still be connectivity manager registering and unregistering network callback.

app/src/main/java/io/gnosis/safe/ui/assets/coins/CoinsFragment.kt
elgatovital: You duplicate this only if you need to react to connectivity changes. All UI actions are explicitly listed for a screen, so I think this kind of duplication is good.

app/src/main/java/io/gnosis/safe/ui/base/BaseStateViewModel.kt
elgatovital: could you elaborate more? Why does it not match our modeling? Why having ConnectivityManager provided to a viewModel is a bad thing architecturally?

elgatovital: what do you mean by a forceful inheritance? Why would we need to poll the connectivity manager?

app/src/main/java/io/gnosis/safe/ui/assets/coins/CoinsFragment.kt
jpalvarezl: I agree with that. This would be good duplication

app/src/main/java/io/gnosis/safe/ui/base/BaseStateViewModel.kt
jpalvarezl: It would add the possibility of of providing this information as `flow` or somehow that allows subscription based consumption. Then that is easily couple-able with screen lifecycles. The network callback will be registered once and then consumers can subscribe to its changes. Then we can isolate this responsibility in a single dependency.

jpalvarezl:
I would be much more in favour of a solution that leverages this for of architecture: https://medium.com/@andrea.bresolin/kotlin-coroutines-interacting-with-external-libraries-and-existing-code-70001a835c30

I can't link the direct section of the article, but please have a look at the code snippet that contains this fragment of code: 

```kotlin
suspend fun execute(): MyResultType {
    return suspendCoroutine<MyResultType> { continuation ->
        ...execute the asynchronous code with callbacks...
        ...call onSuccess(continuation, result) passing the continuation in case of a successful execution...
        ...call onError(continuation, throwable) passing the continuation in case of an error during the execution...
    }
}
private fun onSuccess(
    continuation: Continuation,
    result: MyResultType) {
    ...custom processing logic..
    continuation.resume(result)
}
private fun onError(
    continuation: Continuation,
    throwable: Throwable) {
    ...custom processing logic...
    continuation.resumeWithException(throwable)
}
```

data/src/main/java/io/gnosis/data/repositories/ConnectivityRepository.kt
```diff
@@ -0,0 +1,45 @@
...
+class ConnectivityRepository(private val connectivityManager: ConnectivityManager) {
+
+    var offline: Boolean = false
+
+    interface ConnectivityCallback {
+        fun onConnectivityChange(offline: Boolean)
+    }
+
+    private val networkCallback: ConnectivityManager.NetworkCallback
+    private lateinit var connectivityCallback: ConnectivityCallback
```
jpalvarezl, line 17: Like this you can only have a single listener to this at any given time.

data/src/main/java/io/gnosis/data/repositories/ConnectivityRepository.kt
```diff
@@ -0,0 +1,45 @@
+package io.gnosis.data.repositories
+
+import android.net.ConnectivityManager
+import android.net.Network
+import android.net.NetworkCapabilities
+import android.net.NetworkRequest
+
+class ConnectivityRepository(private val connectivityManager: ConnectivityManager) {
```
jpalvarezl, line 8: This does not belong in the `:data` layer of the app. It is not a repository as it doesn't store any data and it is specific to Android. I think this belongs in the upper layer, in `:app`

app/src/main/java/io/gnosis/safe/ui/assets/coins/CoinsViewModel.kt
```diff
@@ -1,5 +1,6 @@
 package io.gnosis.safe.ui.assets.coins
 
+import android.net.ConnectivityManager
```
rmeissner, line 3: unused import?

app/src/main/java/io/gnosis/safe/ui/safe/add/AddSafeFragment.kt
```diff
@@ -60,7 +60,7 @@ class AddSafeFragment : BaseViewBindingFragment<FragmentAddSafeBinding>() {
             when (val action = state.viewAction) {
                 is ShowValidSafe -> handleValid(action.address)
                 is BaseStateViewModel.ViewAction.Loading -> binding.progress.visible(action.isLoading)
-                is BaseStateViewModel.ViewAction.ShowError -> handleError(action.error)
+                is BaseStateViewModel.ViewAction.ShowError ->  handleError(action.error)
```
rmeissner, line 63: space too much?

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListFragment.kt
```diff
@@ -84,6 +88,17 @@ class TransactionListFragment : SafeOverviewBaseFragment<FragmentTransactionList
...
                         }
                     }
+                    is ShowError -> {
+                        binding.progress.visible(false)
+                        when(viewAction.error) {
+                            is Offline -> {
+                                snackbar(requireView(), R.string.error_no_internet)
+                            }
+                            else -> {
+                                //TODO: handle error
```
rmeissner, line 98: TODO?

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
```diff
@@ -73,7 +75,14 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
                 }
                 is BaseStateViewModel.ViewAction.ShowError -> {
                     binding.refresh.isRefreshing = false
-                    //TODO: handle error here
+                    when(viewAction.error) {
+                        is Offline -> {
+                            snackbar(requireView(), R.string.error_no_internet)
+                        }
+                        else -> {
+                            //TODO: handle error
```
rmeissner, line 83: Where should we address these todos?

app/src/main/java/io/gnosis/safe/helpers/ConnectivityInfoProvider.kt
```diff
@@ -0,0 +1,42 @@
...
+    var offline: Boolean = true
+
+    private val networkCallback = object : ConnectivityManager.NetworkCallback() {
+        override fun onAvailable(network: Network?) {
+            super.onAvailable(network)
+            offline = false
+        }
+
+        override fun onLost(network: Network?) {
+            super.onLost(network)
```
rmeissner, line 20: So if I have wifi and cellular and then wifi disconnects ... what happens here?

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListFragment.kt
elgatovital: at this point we should show empty placeholder instead of content

app/src/main/java/io/gnosis/safe/helpers/ConnectivityInfoProvider.kt
elgatovital: fixed it - offline flag is set only if there is no network available

biafra23: It is a problem. If either disconnect, the state goes to offline. Which is false as the phone can use the other network.

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
biafra23: This TODO is addressed in origin/feature/793-error-and-empty-state-for-details

app/src/main/java/io/gnosis/safe/helpers/ConnectivityInfoProvider.kt
biafra23: Actually when I tried this again, it is ok. We lose wifi but gain cellular a moment later.

app/src/main/java/io/gnosis/safe/helpers/ConnectivityInfoProvider.kt
```diff
@@ -18,7 +18,7 @@ class ConnectivityInfoProvider(private val connectivityManager: ConnectivityMana
 
         override fun onLost(network: Network?) {
             super.onLost(network)
-            offline = true
+            offline = connectivityManager.activeNetworkInfo == null
```
rmeissner, line 21: > Do NOT call ConnectivityManager.getNetworkCapabilities(android.net.Network) or ConnectivityManager.getLinkProperties(android.net.Network) or other synchronous ConnectivityManager methods in this callback as this is prone to race conditions ; calling these methods while in a callback may return an outdated or even a null object.

from https://developer.android.com/reference/android/net/ConnectivityManager.NetworkCallback#onLost(android.net.Network) 😄

rmeissner: If @biafra23 says it works, I am fine with the solution ... for me it was more about what happens if 2 networks are active at the same time (is that even possible?) ... since this method is used to block all network traffic of the app we need to be sure how this works.

biafra23 (approved):

PR #824, opened 2020-08-25
Error and empty state for details

### LINKED ISSUE #793: Transaction details error/empty state
All static views should be hidden. We display generic error with a snackbar at the bottom of the screen.

If the error comes after a reload, the view state is preserved, but the user is still notified with a snackbar.

biafra23:
Closes #793  .

- Don't show emtpy sections on error
- Show error screen with "Data cannot be loaded"
- Show Snackbar with error message
- Do not show error screen on reload (if we already show data) then only show snackbar

Initial error | Reload error
-------- | ---------
initial_error |refresh_error

Not sure about the error message shown in the snackbar. It's probably too technical? Do we have a mapping/source for better error messages?

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
```diff
@@ -73,7 +75,15 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
                 }
                 is BaseStateViewModel.ViewAction.ShowError -> {
                     binding.refresh.isRefreshing = false
-                    //TODO: handle error here
+
+                    binding.content.visibility = View.GONE
+
+                    snackbar(requireView(), viewAction.error.message ?: getString(R.string.error_cannot_load_tx_details))
+                    Timber.e(viewAction.error)
```
elgatovital, line 80: we already log error in base state view model. In general, I think view model is a better place for error logs that are not intended for an end-user.

rmeissner:
Imo it should just display "No internet connection" in all these cases (we did something like that in the legacy app)

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
```diff
@@ -73,7 +75,13 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
                 }
                 is BaseStateViewModel.ViewAction.ShowError -> {
                     binding.refresh.isRefreshing = false
-                    //TODO: handle error here
+
+                    snackbar(requireView(), viewAction.error.message ?: getString(R.string.error_cannot_load_tx_details))
+
+                    if (binding.executed.value.isNullOrBlank() && binding.created.value.isNullOrBlank()) {
```
rmeissner, line 81: This seems arbitrary to me, could you add a comment why this is done?

biafra23: The reason is to not show the empty screen if we had either executed or created date before. Is there a better way to detect a reload, which had a result before?

biafra23:
@rmeissner What other exception (besides UnknownHostException) should be mapped to "No internet connection"? 

All IOExceptions?

elgatovital:
we shouldn't just show propagated error messages as is

rmeissner:
https://github.com/gnosis/safe-android/blob/v1.12.1/app/src/main/java/pm/gnosis/heimdall/ui/exceptions/SimpleLocalizedException.kt#L85 for reference

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
rmeissner: No I think it is fone, I would just add a code comment to avoid future confusion ;)

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
```diff
@@ -73,12 +80,38 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
...
+                        binding.content.visibility = View.GONE
+                        binding.error.visible(true)
+                    }
                 }
             }
         })
     }
 
+    private fun getErrorMessageForException(throwable: Throwable): String? =
```
elgatovital, line 95: it will be the same message logic for all screens - better extract it to utils

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
biafra23: I will add a comment in the next PR

elgatovital (approved):

Issue #808, opened 2020-08-21
[Android] Incorrect app icon in the notification


liliya-soroka:
**Describe the bug**
The black gnosis icon is used for all app version instead of valid app icon 

**To Reproduce**
1. Turn on the dark mode on the device
2. Check the notification in notifications 

**Expected behavior**
the icon should be seen proper

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/7705255c-f958-4da4-b142-73f9c0eccd76)
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/6facf260-c577-4b80-b4a1-8aeb9884c14c)

**Smartphone (please complete the following information):**
 - Device: huawei P30
 - Android Version: 10
 - App Version 1.12.1 - 93

**Additional context**

elgatovital:
@posthnikova @rmeissner @tschubotz We agreed to use the monochrome icon for notifications. This is how it looks like in a day mode 
![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/1518f20a-80a2-4e6d-a677-47440fb40818)
But what should we do in a dark mode? We could use white color for the symbol

posthnikova:
@elgatovital White for dark mode is fine

elgatovital:
![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/48adebdd-b88e-4658-9ac2-e328b373ac50)
![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/9d73de7c-4b40-461c-a858-c5bae103ad67)

tschubotz:
beautiful!

liliya-soroka:
Can we use the same nice app icons ( yellow, black, red , with and without rinleby ) that are used by ios app instead of just black icon?
Example: ![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/867bca4a-1caf-4359-b222-79469b900e68)
app version 1.12.1 -105 ( staging and prod rinkeby)

elgatovital:
no, we have to use monochrome icons

elgatovital:
also, we agreed to use icons without text. This issue can be closed

PR #829, opened 2020-08-26
Use icon without padding for notifications

elgatovital:
Handles #808
result
image


biafra23 (commented):
Does this really handle #808 . 808 is about different icons for dark background.

biafra23 (approved):

elgatovital:
> #808

check my comment in #808 with a screenshot for a dark mode

Issue #742, opened 2020-07-16
Snackbar should be on top of bottom bar


elgatovital:
- Device:
- Android OS version:

tschubotz:
Is this a bug or which Epic is this related to?

tschubotz:
This sounds like a bug, right? I'll mark it as such. Let me know if objections. @elgatovital @jpalvarezl

jpalvarezl:
@tschubotz 
> Is this a bug or which Epic is this related to?

This was found by Richard back in the Kyoto sprint I believe. It blocks the bottom navigation bar for as long as the snackbar is displayed, AFAIK.

liliya-soroka:
Verified
The bottom navigation bar is not blocked with snackbar
app version 1.12.1 - 98

PR #831, opened 2020-08-26
Display snackbar on top of bottom bar

elgatovital:
Handles #742 

- Add coordinator layouts

@gnosis/mobile-devs


app/src/main/res/layout/fragment_collectibles.xml
```diff
@@ -1,18 +1,27 @@
...
+            android:text="@string/coming_soon"
+            android:textSize="16sp"
+            app:layout_constraintBottom_toBottomOf="parent"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toTopOf="parent" />
+
+    </androidx.constraintlayout.widget.ConstraintLayout>
+
+</androidx.coordinatorlayout.widget.CoordinatorLayout>
```
jpalvarezl, line 27: The nesting here doesn't seem necessary. Would it be possible to use `layout_gravity="center"` instead?

app/src/main/res/layout/fragment_coins.xml
```diff
@@ -1,43 +1,50 @@
...
+                android:id="@+id/coins"
+                android:layout_width="match_parent"
+                android:layout_height="match_parent"
+                tools:listitem="@layout/item_coin_balance" />
+        </androidx.swiperefreshlayout.widget.SwipeRefreshLayout>
+
+    </androidx.constraintlayout.widget.ConstraintLayout>
 
-</androidx.constraintlayout.widget.ConstraintLayout>
+</androidx.coordinatorlayout.widget.CoordinatorLayout>
```
jpalvarezl, line 50: Wondering if we can do without the nesting of `CoordinatorLayout` and `ConstraintLayout` . The progress bar can probably use `layout_gravity="center"` and the other 2 views are `match_parent` anyway.

app/src/main/res/layout/fragment_settings_safe.xml
```diff
@@ -1,159 +1,167 @@
 <?xml version="1.0" encoding="utf-8"?>
-<androidx.constraintlayout.widget.ConstraintLayout
+<androidx.coordinatorlayout.widget.CoordinatorLayout
```
jpalvarezl, line 2: Same as the previous 2 comments. Is it absolutely necessary to nest? If it is feel free to ignore this comment.

app/src/main/res/layout/fragment_transaction_list.xml
```diff
@@ -1,75 +1,81 @@
...
 
-    </androidx.swiperefreshlayout.widget.SwipeRefreshLayout>
+        <androidx.swiperefreshlayout.widget.SwipeRefreshLayout
+            android:id="@+id/refresh"
+            android:layout_width="0dp"
+            android:layout_height="0dp"
+            app:layout_constraintBottom_toBottomOf="parent"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toTopOf="parent">
```
jpalvarezl, line 35: This could probably be `match_parent` in both width and height and then possibly get rid of the nesting.

jpalvarezl (approved):
I would just try to get rid of what is seemingly unnecessary nesting. If it is strictly necessary just go for it, but at a first glance it doesn't seem to be the case.

Issue #644, opened 2020-06-10
[Android]Enter ENS Name: wrong message when no internet connection


liliya-soroka:
**Describe the bug**
Couldn't find address message is displayed instead of no internet connection notification when internet connection is turned off on the device
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/ed7dcfd8-c5b3-4418-8f7b-34cb3f2ed031)

**To Reproduce**
1. Open Load Safe 
2. Select Enter ENS name
3. Turn off internet connection
4. Start to type ENS name 

**Expected behavior**
No internet connection notification should be displayed
Example of No internet notification - https://app.zeplin.io/project/5e9efc30726e106958cc43de/screen/5ebbb32d9ff6f1622643863e

**Smartphone (please complete the following information):**
 - Device: Huawei p30
 - Android Version: 10
 - App Version 1.12.1- 28

liliya-soroka:
The new issue with the incorrect message is created
https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-android/840

This one will be closed as the popup view was changed to the page view

PR #834, opened 2020-08-26
Handle offline state in ens input screen

elgatovital:
Handles #644 

@gnosis/mobile-devs


jpalvarezl (approved):
Looks good.

app/src/main/java/io/gnosis/safe/ui/dialogs/EnsInputViewModel.kt
```diff
@@ -18,7 +19,10 @@ class EnsInputViewModel
                 it ?: throw EnsResolutionError()
             }
             .onFailure {
-                throw EnsResolutionError(it.cause?.localizedMessage ?: it.localizedMessage)
+                if (it is Offline)
+                    throw it
```
biafra23, line 23: I really prefer braces on multiline if-else. But this is just a matter of taste and bad experience without blocks.

biafra23 (approved):

PR #836, opened 2020-08-26
Send only x.y.z versions to notification service

elgatovital:
semantic versioning validation for registering push notification devices on the transaction service was added => the versions android is sending have to be adjusted. iOS is sending x.y.z, so we will do the same.

- Adjust version sent to notification service when registering device

@gnosis/mobile-devs


jpalvarezl (approved):
Are the versions here not prefixed with `v` maybe that could mess up the regex, otherwise looks good 👍

biafra23 (approved):
We need to be careful not to change the BuildConfig.VERSION_NAME and maybe add a unit test. But not necessarily in this PR

Issue #852, opened 2020-08-27
[Android] There is no amount in the incoming ETH tx notifications


liliya-soroka:
**Describe the bug**
There is no amount in the incoming ETH tx notifications

**To Reproduce**
1. add safe to the app
2. send ETH to the added safe
3. check notification
Current result: There is no amount

**Expected behavior**
From Epic:
ETH was received
Client should display the Safe name.
Client just displays that ETH was received incl. amount
Tx list is opened upon opening notification
For amount: Please use regular amount formatting rules
Texts:
<safe_name: <amount> ETH received

**Screenshots**

**Smartphone (please complete the following information):**
 - Device: Pixel 4
 - Android Version: 10.1
 - App Version 1.12.1-105

**Additional context**

liliya-soroka:
Verified without formatting check
app version 1.12.1-107

PR #853, opened 2020-08-28
Show amount for incoming eth notifications

elgatovital:
Handles #852 

image

@gnosis/mobile-devs


jpalvarezl (approved):
I left a comment in the bug regarding balance formatting to include a fix for this too.

Issue #860, opened 2020-08-28
[Android] Transactions list is not updated after calling the app from background


liliya-soroka:
**Describe the bug**
Transactions list is not updated after calling the app from background
**To Reproduce**
1. send the app to the background
2. send tx to the safe and send from the safe
3. call the app from the background

**Expected behavior**
The transactions list should be updated after calling the app from background

**Smartphone (please complete the following information):**
 - Device: Pixel 4
 - Android Version: 10.1
 - App Version 1.12.1-110

**Additional context**

liliya-soroka:
Verified
app version 1.12.1 -114

Issue #858, opened 2020-08-28
Opening TX List from push notification


jpalvarezl:
**Describe the bug**
Transaction list does not show the new transaction when clicking the notification. This maybe have been solved by https://github.com/gnosis/safe-android/pull/857

**Additional context**
This came from `bug bashing` session

liliya-soroka:
Not fixed in 1.12.1-110
When the incoming tx notification has clicked the app with opened txs list is opened but it doesn't contain the tx from the notification
@tschubotz , I think it's a major issue

liliya-soroka:
Verified
app version 1.12.1 -114

Issue #855, opened 2020-08-28
Transactions list is not updated after moving from transaction details page


liliya-soroka:
**Describe the bug**
Transactions list is not updated after moving from the transaction details page. In some cases, it's really confusing.
Case:
1. The app is located in the background on the user's device
2. Send a few outgoing txs without app reopening
3. Click on the latest received notification
4. Click notification 
5. Go back from the transaction details for the latest executed tx

Current result: The transactions list is no up to date and the user is confusing why he can not see just checked transaction in the list 
Expected result: I think that the txs list should be updated after going back from the tx details

**Smartphone (please complete the following information):**
 - Device: Pixel 4
 - Android Version: 10.1
 - App Version 1.12.1 - 107

tschubotz:
Does this happen to all notifications which open the tx details? I yes I think this is major, if not I think it should be minor.

liliya-soroka:
The issue still actual.
Pull to refresh required to update transactions list after coming back from the transaction details page.
Current result: The transactions list page is blinked but not updated

app version 1.12.1-110 ( staging and prod Rinkeby)

liliya-soroka:
Verified
app version 1.12.1 - 114

Issue #835, opened 2020-08-26
[Android]Transactions list is not updated for 10 min if it was updated with pull to refresh before changes from back-end: txs from Queue is not moved to History


liliya-soroka:
**Describe the bug**
1. Already executed txs are displayed as awaiting confirmation and awaiting executing for 10 min on transaction list after it was actually executed ( executed state confirmed on web and ios) if the transactions list was opened when txs had statuses awaiting confirmation/awaiting execution
2. Queue section with created tx doesn't appear when the new tx is created if the transaction list was updated with pull to refresh right before tx creation 
Reproduced both on staging and prod 
staging safe 0x1230B3d59858296A31053C1b8562Ecf89A2f888b
prod safe 0x8675B754342754A30A2AeF474D114d8460bca19b

**To Reproduce**
1. Go to the safe in android app and on web
2. Create tx that requires confirmation from a few owners
3. Check mobile app to make sure that tx appears in Queue
4. Reject tx as owner 
5. Pull to refresh txs list in mobile app
Current result: rejection tx doesn't appear for 10 min ( was reproduced on prod)

6. Wait till both transactions are displayed in the Queue in mobile app
7. Confirm original tx without executing it
8. Pull to Refresh txs list in mobile app
Current result: The states are updated correctly

9. Execute original tx on web
10 Pull to refresh txs list

Current result: The original and rejection tx states are not updated for 10 min and not moved to History section

**Expected behavior**
The txs list should be updated on pull to refresh

**Smartphone (please complete the following information):**
 - Device: Pixel 4, Huawei P30
 - Android Version: 10.1
 - App Version 1.12.1-98 
- client gateway version 1.0.0 build 168

**Additional context**

jpalvarezl:
I tried reproducing this bug with no success in staging/rinkeby. The cancellation transaction was shown (basically step 5 didn't go as according to the description of this issue.) 

The states appear to be updated correctly for me. I think yesterday @rmeissner was updating some webhook tokens that could have had an impact on this. I opened an issue in the client gateway to keep track of this anyway (https://github.com/gnosis/safe-client-gateway/issues/131) 

@liliya-soroka If you find at some point of the day you have spare time, maybe give it another try. I personally could not reproduce the error :(

jpalvarezl:
I have just tried to reproduce this bug in prod/rinkeby with no success. The Image below was taken right after submitting the cancellation transaction in web and pulling down to refresh in the app: 

[...]

PR #861, opened 2020-08-28
Always load transaction to display latest state; remove caching

elgatovital:
Handles #623
Handles #809
Handles #828
Handles #835
Handles #855 
Handles #858
Handles #860

- Always load the latest state of the transactions
- Remove caching from tx list
- Fix retry handling for tx list
- Show retry error messages
- Fix tx list flashing issues
- Update list when navigating back from tx details
- Keep showing old data in case of an error on the pull to refresh

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListFragment.kt
@@ -54,7 +54,7 @@ class TransactionListFragment : SafeOverviewBaseFragment<FragmentTransactionList
         adapter.addLoadStateListener { loadState ->
             if (lifecycle.currentState.isAtLeast(Lifecycle.State.STARTED)) {
 
-                binding.progress.isVisible = loadState.refresh is LoadState.Loading
+                binding.progress.isVisible = loadState.refresh is LoadState.Loading && adapter.itemCount == 0
                 binding.refresh.isRefreshing = false
 
                 val append = loadState.append
@@ -85,7 +85,7 @@ class TransactionListFragment : SafeOverviewBaseFragment<FragmentTransactionList
         binding.refresh.setOnRefreshListener { viewModel.load() }
 
         viewModel.state.observe(viewLifecycleOwner, Observer { state ->
-            binding.progress.visible(state.isLoading)
+
             binding.contentNoData.root.visible(false)
 
             state.viewAction.let { viewAction ->
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListFragment.kt
@@ -54,6 +54,14 @@ class TransactionListFragment : SafeOverviewBaseFragment<FragmentTransactionList
         adapter.addLoadStateListener { loadState ->
             if (lifecycle.currentState.isAtLeast(Lifecycle.State.STARTED)) {
                 binding.progress.isVisible = loadState.refresh is LoadState.Loading
+                val append = loadState.append
+                if(append is LoadState.Error) {
+                    handleError(append.error)
+                }
+                val prepend = loadState.prepend
+                if(prepend is LoadState.Error) {
+                    handleError(prepend.error)
+                }
             }
         }
         adapter.addDataRefreshListener { isEmpty ->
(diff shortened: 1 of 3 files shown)


app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListFragment.kt
```diff
@@ -66,6 +73,7 @@ class TransactionListFragment : SafeOverviewBaseFragment<FragmentTransactionList
                     showEmptyState()
                 } else {
                     showList()
+                    binding.transactions.scrollToPosition(0)
                 }
```
rmeissner, line 77: Doesn't that scroll to the top each time the list is updated?

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListFragment.kt
```diff
@@ -53,8 +53,15 @@ class TransactionListFragment : SafeOverviewBaseFragment<FragmentTransactionList
...
+                if(append is LoadState.Error) {
+                    handleError(append.error)
+                }
+                val prepend = loadState.prepend
+                if(prepend is LoadState.Error) {
+                    handleError(prepend.error)
+                }
             }
         }
         adapter.addDataRefreshListener { isEmpty ->
```
rmeissner, line 67: Is this method deprecated? https://developer.android.com/reference/kotlin/androidx/paging/PagingDataAdapter#adddatarefreshlistener

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListFragment.kt
```diff
@@ -53,8 +53,15 @@ class TransactionListFragment : SafeOverviewBaseFragment<FragmentTransactionList
 
         adapter.addLoadStateListener { loadState ->
             if (lifecycle.currentState.isAtLeast(Lifecycle.State.STARTED)) {
-                binding.transactions.isVisible = loadState.refresh is LoadState.NotLoading
                 binding.progress.isVisible = loadState.refresh is LoadState.Loading
```
rmeissner, line 56: Is this the full page loading spinner?

elgatovital: yes

rmeissner: And it shows whenever we load data?

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListFragment.kt
elgatovital: it won't do it by default, so we have to do it manually

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListFragment.kt
elgatovital: only if the list is empty. For lists with content we show only pull to refresh spinner

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListFragment.kt
```diff
@@ -119,9 +115,21 @@ class TransactionListFragment : SafeOverviewBaseFragment<FragmentTransactionList
 
     override fun onResume() {
         super.onResume()
+        binding.transactions.scrollToPosition(0)
```
rmeissner, line 118: Why are we doing this? I would only scroll to top if we have new data ... it is a major inconvenience that the android app scrolls up even if nothing in the list changed. Did you ever try browsing a long transaction list?

biafra23: I agree. But AFAICT it is specified that way. We should change the specification as it is really annoying with longer tx lists.

elgatovital: yes, we could discuss it after release and then change this behaviour for both android and ios.

biafra23 (approved):
LGTM 👍

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListFragment.kt
rmeissner: iOS only scrolls up if they really load data ... they don't do a preemptive scrolling up when nothing changed ... this is bad ux and just so that it feels the same we should not introduce bad ux ....

rmeissner: Also could you reference where this is specified please.

Issue #819, opened 2020-08-24
[Android]TX details: Typo: "Canceled" instead if "Cancelled" 


liliya-soroka:
**Describe the bug**
Typo: "Canceled" instead if "Cancelled" 

**To Reproduce**
1. Go to 0x938bae50a210b80EA233112800Cd5Bc2e7644300 safe
2. Open cancelled tx

**Expected behavior**
two ll should be in the cancelled state name in confirm block 

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/3cf6ce0a-615b-46ec-b3e4-100893104bf4)

**Smartphone (please complete the following information):**
 - Device: Pixel 4
 - Android Version: 10
 - App Version 1.12.1 - 93

**Additional context**

biafra23:
How do we know "Cancelled" is the correct spelling?

Miriam Webster on this Subject: "... both canceled and cancelled are acceptable for the past tense..." 

Source: https://www.merriam-webster.com/words-at-play/canceled-or-cancelled

liliya-soroka:
I offer to keep Cancelled as we are using it in other places in the android app , ios app and web app

liliya-soroka:
Verified
app version 1.12.1 -114

PR #862, opened 2020-08-28
Change spelling: Canceled -> Cancelled

elgatovital:
Handles #819

@gnosis/mobile-devs


jpalvarezl (approved):

biafra23 (approved):
Shouldn't we renamed `TxExecutionStep.Type.CANCELED` also, to avoid confusion of the spelling in the future? Just to be consistent?

Issue #859, opened 2020-08-28
Push notification: No ERC721 incoming tx


liliya-soroka:
**Describe the bug**
No ERC721 incoming tx

**To Reproduce**
1. Send ERC721 token to the safe in the app on the device

**Expected behavior**
The incoming ERC721 tx notification should be received.
Notification receiving works for the ios app , so It's not back-end issue

**Smartphone (please complete the following information):**
 - Device: Pixel 4
 - Android Version: 10.1
 - App Version 1.12.1-110

**Additional context**

liliya-soroka:
Verified
app version 1.12.1 - 114

PR #863, opened 2020-08-31
Fix parsing of  incoming ERC721 notification

elgatovital:
Handles #859 
 
@gnosis/mobile-devs


jpalvarezl (approved):
I would be nice to have a bit more polymorphism here. Then we would be able to avoid having to have nullable fields. Still looks good though 👍

biafra23 (approved):

Issue #854, opened 2020-08-28
Track user properties for all events.


tschubotz:
**Describe the bug**
We have the user property `num_safes`. This seems to be set only when Safes are added / deleted (cf. [Slack](https://gnosisinc.slack.com/archives/C63LQTGNB/p1598604914314600?thread_ts=1598528873.294700&cid=C63LQTGNB). 

However the intention of the user property was to have it tracked always, so it can also be figured out when a user does never add a Safe.

**To Reproduce**

I used BigQuery to figure this out (cf. [Slack](https://gnosisinc.slack.com/archives/C63LQTGNB/p1598604045314000?thread_ts=1598528873.294700&cid=C63LQTGNB), [Wiki](https://github.com/gnosis/safe/wiki/How-to-check-tracking-events-with-BigQuery)

```SQL
SELECT event_name, count(*)
FROM `safe-firebase-mainnet.analytics_243403902.events_*` 
where platform = 'ANDROID'
  and app_info.version like '1.12.1-98-gf%'
  and not EXISTS(SELECT key FROM UNNEST(user_properties) WHERE key='num_safes')
group by 1
order by 1 asc
```

**Expected behavior**

`num_safes` is tracked with any event.
All user props should be tracked with any event actually.

liliya-soroka:
Verified
app version 1.12.1 -115

liliya-soroka:
Reopened.
We have expected that the all user_props are tracked with all events.
app version 1.12.1 -115

liliya-soroka:
@DmitryBespalov , the same issue is actual for the ios app.
According the stand up discussion ( week ago) the expected result should be discussed one more time

DmitryBespalov:
@tschubotz I'm confused, as user properties are attributes that you would use to describe segments of the user base. Do I understand that what you actually mean is that every event should have **event** parameters, including num_safes?

tschubotz:
I thought that event params should be specific to events and user attributes are something that is specific to the user. They can be used for segmentation yes. So for instance a segment could be "users with no Safe" or "users with more than 2 Safes". Hence I think `num_safes` should be user a property.

DmitryBespalov:
Refinement meeting: the debug view doesn't show the user properties right away. The values of the user properties are added retroactively. If the user changes the property within the same session and goes back to the initial value, then that change is not tracked.

tschubotz:
After implementing #1045 it seems to work now:

```
SELECT event_name, app_info.version, count(*)
[...]

PR #867, opened 2020-09-01
Set user properties on app start

elgatovital:
Handles #854

- Check and set user properties on app start

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/splash/SplashActivity.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/splash/SplashActivity.kt
@@ -6,22 +6,14 @@ import androidx.lifecycle.Observer
 import androidx.lifecycle.lifecycleScope
 import io.gnosis.safe.ScreenId
 import io.gnosis.safe.databinding.ActivitySplashBinding
-import io.gnosis.safe.notifications.NotificationManager
-import io.gnosis.safe.notifications.NotificationRepository
-import io.gnosis.safe.ui.base.activity.BaseActivity
 import io.gnosis.safe.ui.base.BaseStateViewModel.ViewAction
+import io.gnosis.safe.ui.base.activity.BaseActivity
 import io.gnosis.safe.ui.terms.TermsBottomSheetDialog
 import kotlinx.coroutines.launch
 import pm.gnosis.svalinn.common.utils.visible
 import javax.inject.Inject
 
 class SplashActivity : BaseActivity() {
-
-    @Inject
-    lateinit var notificationRepo: NotificationRepository
-
-    @Inject
-    lateinit var notificationManager: NotificationManager
 
     @Inject
     lateinit var viewModel: SplashViewModel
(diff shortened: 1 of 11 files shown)


app/src/main/java/io/gnosis/safe/ui/splash/SplashActivity.kt
```diff
@@ -69,8 +58,7 @@ class SplashActivity : BaseActivity() {
         )
 
         lifecycleScope.launch {
-            notificationRepo.checkPermissions()
-            notificationRepo.register()
+            viewModel.onAppStart()
```
rmeissner, line 61: Just a question no blocker (as the behaviour is not changed): If I have a bad internet connection, is the splash screen then shown longer, as we wait for the app to register for push on our backend service?

rmeissner: Ohh wait, so after 500ms (or when the user agrees to the terms, we continue) ... doesn't that then kill the lifecycleScope? ... anyways I don'T think it is a big issue as the request is already fired and even if the lifecycleScope is cancelled it should not be able to cancel the processing of the request in the backend

app/src/main/java/io/gnosis/safe/ui/splash/SplashViewModel.kt
```diff
@@ -12,13 +15,28 @@ import javax.inject.Inject
...
 
+    suspend fun onAppStart() {
+
+        notificationRepository.register()
+
+        notificationRepository.clearNotifications()
+        val pushEnabled = notificationRepository.checkPermissions()
+        tracker.setPushInfo(pushEnabled)
+
+        val numSafes = safeRepository.getSafes().size
```
jpalvarezl, line 36: Should we encapsulate this in a method within the `SafeRepository` just in case at some point there is some magic involved on the safe list returned by `getSafes()` ?

app/src/test/java/io/gnosis/safe/ui/splash/SplashViewModelTest.kt
```diff
@@ -102,4 +110,27 @@ class SplashViewModelTest {
         }
         coVerify(exactly = 1) { termsChecker.getTermsAgreed() }
     }
+
+    @Test
+    fun `onAppStart should register notification service and set user properties`() = runBlockingTest {
```
jpalvarezl, line 115: Maybe we could test when we have more than 0 safes?

jpalvarezl (approved):
Minor comments. Looks good!

app/src/main/java/io/gnosis/safe/ui/splash/SplashViewModel.kt
elgatovital: safeRepository.getNumberOfSafes() ?

jpalvarezl: `getSafeCount` or what you said. Either option is fine :)

biafra23 (approved):

Issue #848, opened 2020-08-27
Details screen for creation transaction


DmitryBespalov:
- Data is taken from the list item
- NO pull to refresh
- "Implementation/Master copy" and "Factory" address may be not available

rmeissner:
this depends on https://github.com/gnosis/safe-android/issues/588 right?

liliya-soroka:
@biafra23 , Do we have separate task for the creation tx that should be tested ?

liliya-soroka:
Verified
app version 1.12.1 - 123 
The case with no mastercopy and factory will be retested with task https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-android/847

Issue #588, opened 2020-05-07
Display creation transaction


elgatovital:
# What is this feature about? (1 sentence)
This will display the creation transactions in the list of transactions of a Safe.

# Why is it needed? What is the value? For whom do we build it?
There is a transaction list (#289), however the creation transaction is not part of that list yet. But it is something that users expect to see.

Also the web interface shows that info (gnosis/safe-react#554)

# High-level overview of the feature
- There is an entry for this transaction on the transaction list.
- The details page shows the following details:
  - Tx hash (Link to Etherscan)
  - Owner addresses (Link to Etherscan, can be null if setupData is not available)
  - Required confirmations (n out of m, can be null if setupData is not available)
  - Creator address (Link to Etherscan)
  - Mastercopy used (If available, can be null for some Safes). If available, then contract version should be displayed similar to the Safes settings
  - Factory used (If available, can be null for some Safes)
  - Setup data (If available, can be null for some Safes)

# Misc
There is an endpoint for this info: https://safe-transaction.gnosis.io/api/v1/safes/0xA063Cb7CFd8E57c30c788A0572CBbf2129ae56B6/creation/
(Owner addresses and threshold can be parsed from setupData)

PR #873, opened 2020-09-03
Adds creation tx to tx list

### LINKED ISSUE #847: Creation transaction in the list of transactions

biafra23:
Closes #847  .

- Adds creation tx to the end of the tx list

Screenshot_20200903-163845

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewHolderFactory.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewHolderFactory.kt
@@ -7,13 +7,7 @@ import androidx.core.content.res.ResourcesCompat
 import androidx.navigation.Navigation
 import androidx.viewbinding.ViewBinding
 import io.gnosis.safe.R
-import io.gnosis.safe.databinding.ItemTxChangeMastercopyBinding
-import io.gnosis.safe.databinding.ItemTxQueuedChangeMastercopyBinding
-import io.gnosis.safe.databinding.ItemTxQueuedSettingsChangeBinding
-import io.gnosis.safe.databinding.ItemTxQueuedTransferBinding
-import io.gnosis.safe.databinding.ItemTxSectionHeaderBinding
-import io.gnosis.safe.databinding.ItemTxSettingsChangeBinding
-import io.gnosis.safe.databinding.ItemTxTransferBinding
+import io.gnosis.safe.databinding.*
 import io.gnosis.safe.ui.base.adapter.Adapter
 import io.gnosis.safe.ui.base.adapter.BaseFactory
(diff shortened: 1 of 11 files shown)


app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewHolderFactory.kt
```diff
@@ -329,6 +327,27 @@ class CustomTransactionViewHolder(private val viewBinding: ItemTxTransferBinding
...
+        with(viewBinding) {
+            finalStatus.setText(viewTransfer.statusText)
+            finalStatus.setTextColor(ResourcesCompat.getColor(resources, viewTransfer.statusColorRes, theme))
+
+            dateTime.text = viewTransfer.dateTimeText
+            settingName.setText(viewTransfer.label)
+
+            // TODO: Add in future PR
+//            root.setOnClickListener {
+//                // TBD
```
elgatovital, line 345: next time please be more specific - instead of just tbd something like navigate to tx details page would be more descriptive. No action needed

data/src/main/java/io/gnosis/data/backend/dto/GateTransaction.kt
```diff
@@ -50,7 +50,11 @@ sealed class TransactionInfoDto {
     ) : TransactionInfoDto()
 
     data class Creation(
-        override val type: GateTransactionType = GateTransactionType.Creation
+        override val type: GateTransactionType = GateTransactionType.Creation,
+        val creator: Solidity.Address,
+        val transactionHash: String,
+        val implementation: Solidity.Address,
```
elgatovital, line 56: what is implementation? Master copy address?

elgatovital:
Having some unit tests for creation tx would be nice

data/src/main/java/io/gnosis/data/backend/dto/GateTransaction.kt
jpalvarezl: Yeah, it was part of the inclusive renaming. Anywhere you see `implementation`, as far as contracts are concerned, think of it as `masterCopy`

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewHolderFactory.kt
biafra23: At the time it wasn't clear whether we would need to fetch the creation tx from the backend or just show, what we get in the tx list. That's why there is a TBD. See https://github.com/gnosis/safe-android/issues/848 and https://github.com/gnosis/safe-android/issues/588

elgatovital (approved):

Issue #889, opened 2020-09-11
[Android]Collectible details: It should be possible to copy asset contract to clipboard


liliya-soroka:
**Describe the bug**
There is no possibility to copy asst address. Instead of copy the page about address in Etherscan is opened

**To Reproduce**
1. Go to safe 0x1230B3d59858296A31053C1b8562Ecf89A2f888b
2. Open collectibles
3. Open details for any collectible
4. click asset address

**Expected behavior**
It should be possible to copy asset contract to clipboard

**Screenshots**

**Smartphone (please complete the following information):**
 - Device: Pixel 4
 - Android Version: 10.1
 - App Version 2.0.0 ( 10)

**Additional context**

elgatovital:
same here (advanced safe settings)? ![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/dfee7ea8-32c4-4aa3-b3f7-b7db5716bfc8) 
Click on address -> copy to clipboard; click on link icon -> go to etherscan?

tschubotz:
yes :)

liliya-soroka:
Verified
app version 2.0.0 ( 16)

Issue #886, opened 2020-09-11
[Android]Collectibles: no shadow above top border of the card


liliya-soroka:
**Describe the bug**
Collectibles: no shadow above top border of the card. It's a little bit complicated to see the top of the card 

**To Reproduce**
1. Go to 0x1230B3d59858296A31053C1b8562Ecf89A2f888b
2. Open Collectibles list

**Expected behavior**
It should be done as on design - https://app.zeplin.io/project/5e9efc30726e106958cc43de/screen/5e9efc63f095427a65137844

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/06923179-2cc4-4988-8e31-dff1e0db9ce8)

**Smartphone (please complete the following information):**
 - Device: Pixel 4
 - Android Version: 10.1
 - App Version 2.0.0 - 10

**Additional context**

rmeissner:
By default cards don't have a shadow on top according to material design: https://material.io/components/cards (Android uses the library by google for material design)

@posthnikova is this required?

posthnikova:
> By default cards don't have a shadow on top according to material design

@rmeissner Not everywhere, it depends on the elevation. At 1dp the shadow on top is barely visible. Cards are 1dp so this looks correct. But still in the official UI kit you can see the shadow on top, on production it's a bit different. 

![image](https://user-images.githubusercontent.com/45787536/92890007-04ca1880-f417-11ea-93f5-0fe909ff79a0.png)

Tab bar and Nav bar are elevated more therefore the shadows should be more visible.

rmeissner:
Yes the different elevations should be supported by the Android sdk (can be probably applied with `app:elevation="4dp"`)

liliya-soroka:
Verified
app version 2.0.0 ( 16)

PR #892, opened 2020-09-11
Fix collectible issues

elgatovital:
Handles #886 
Handles #889
Handles #894 

- Make shadow on top of collectible list item more visible
- Copy address to clipboard on click for LabeledAddressItem
- Hide view on token link on collectible details page

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/res/layout/item_collectible_collectible.xml
+++ b/app/src/main/res/layout/item_collectible_collectible.xml
@@ -12,6 +12,7 @@
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:layout_gravity="center"
+        android:layout_marginTop="2dp"
         android:layout_marginHorizontal="16dp"
         android:layout_marginStart="16dp"
         android:layout_marginBottom="12dp"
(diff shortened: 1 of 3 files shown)


app/src/main/res/layout/item_collectible_collectible.xml
```diff
@@ -12,6 +12,7 @@
         android:layout_width="match_parent"
```
rmeissner, line 12: Why is it required to wrap this card view in a constraint layout?

app/src/main/res/layout/item_collectible_collectible.xml
```diff
@@ -12,6 +12,7 @@
...
         android:layout_height="wrap_content"
         android:layout_gravity="center"
+        android:layout_marginTop="2dp"
```
rmeissner, line 15: This only increases the distance between the items, right? It does not add any more shadow.

elgatovital: top shadow/line was clipped by view bounds. We can fix it by either adding padding to parent or having this margin

app/src/main/res/layout/item_collectible_collectible.xml
elgatovital: we want to have a white background for the whole width of the item

biafra23 (approved):

Issue #885, opened 2020-09-11
[Android] Groups and items are not sorted alphabetically in Collectibles list


liliya-soroka:
**Describe the bug**
Groups and items in each group are not sorted by name alphabetically 

**To Reproduce**
1. Go to 0x1230B3d59858296A31053C1b8562Ecf89A2f888b
2. Open collectibles

**Expected behavior**
Groups and items in each group should be sorted by name alphabetically 
**Screenshots**

**Smartphone (please complete the following information):**
 - Device: Pixel 4
 - Android Version: 10.1
 - App Version 2.0.0 ( 10)

**Additional context**

liliya-soroka:
Verified
app version 2.0.0 (17)

PR #896, opened 2020-09-11
Fix sorting for collectible list

elgatovital:
Handles #885

- Sort groups alphabetically
- Sort collectibles in groups alphabetically
- Put `unknown` elements at the end of the list

@gnosis/mobile-devs


biafra23 (commented):
Could you add one more "SpecialToken" with null as the `name` and assert that null names are sorted after non-null names within the group?

biafra23 (approved):

PR #899, opened 2020-09-14
Improve connectivity info provider

elgatovital:
@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/helpers/ConnectivityInfoProvider.kt
+++ b/app/src/main/java/io/gnosis/safe/helpers/ConnectivityInfoProvider.kt
@@ -4,6 +4,7 @@ import android.net.ConnectivityManager
 import android.net.Network
 import android.net.NetworkCapabilities
 import android.net.NetworkRequest
+import android.os.Build
 import java.io.IOException
 
 class ConnectivityInfoProvider(private val connectivityManager: ConnectivityManager) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/gnosis/safe/helpers/ConnectivityInfoProvider.kt
```diff
@@ -18,7 +19,19 @@ class ConnectivityInfoProvider(private val connectivityManager: ConnectivityMana
...
+                capabilities?.let {
+                    when {
+                        capabilities.hasTransport(NetworkCapabilities.TRANSPORT_CELLULAR) -> false
+                        capabilities.hasTransport(NetworkCapabilities.TRANSPORT_WIFI) -> false
+                        else -> true
+                    }
+                } ?: true
+            } else {
+                val activeNetworkInfo = connectivityManager.activeNetworkInfo
+                activeNetworkInfo == null || !activeNetworkInfo.isConnectedOrConnecting
```
biafra23, line 33: So if the phone is currently trying to reestablish the connection `isConnectedOrConnecting` the were not offline?
Should we consider "connecting" online?

elgatovital: yes, we can be more strict in this regard

biafra23 (approved):

Issue #903, opened 2020-09-14
DataDecoded: support decoding arrays of values and nested arrays (non-UI)


DmitryBespalov:
Decoding of arrays containing string values or arrays (recursive).

liliya-soroka:
closed as no testing required

Issue #329, opened 2019-02-13
Error "404( )"


damiandragobetzky:
When you enter the recovery phrase, a 404( ) sign shows.

Pre-Conditions:
Have a Safe wallet address made and its recovery phrase.

Steps to reproduce:

1. Enter the Safe Wallet Mobile App
2. Tap on "Recover a Safe"
3. Write your existent recovery phrase
4. Observe

Expected result:
You can access the application normally

Actual result:
A 404( ) message is shown, and the user is unable to access.

Smartphone:
Device: LG Nexus X5
Android Version: 8.1.0
App version: 1.1.0-Staging (Rinkeby)

Capture: 
https://drive.google.com/open?id=1sTeKc4xFN5Zwdh7hJ237_erHiWQ-8Ex0

elgatovital:
not possible to reproduce. Currently safe existence is already being checked when entering safe address

PR #906, opened 2020-09-15
Parse tx params

elgatovital:
Handles #903

- Create subtypes for params and map solidity types to them

@gnosis/mobile-devs


data/src/main/java/io/gnosis/data/backend/dto/TransactionDto.kt
```diff
@@ -1,16 +1,80 @@
...
+        Solidity.types.keys.filter { it.contains("int") }
+    ),
+    BOOL(listOf("bool")),
+    STRING(listOf("string"));
+
+    companion object {
+        fun of(value: String): ParamType =
+            when {
+                value == "address" -> ADDRESS
+                value.contains("[]") -> ARRAY
```
rmeissner, line 68: This doesn't cover fixed sized arrays

data/src/main/java/io/gnosis/data/backend/dto/TransactionDto.kt
```diff
@@ -1,16 +1,80 @@
...
+        override val type: String,
+        override val name: String,
+        override val value: Any
+    ) : ParamDto()
+}
+
+enum class ParamType(val mappedSolidityTypes: List<String>) {
+    ADDRESS(listOf("address")),
+    ARRAY(
+        Solidity.types.keys.map { "${it}[]" }
```
rmeissner, line 53: There can be arrays of arrays

data/src/main/java/io/gnosis/data/backend/dto/TransactionDto.kt
```diff
@@ -1,16 +1,80 @@
...
+    ),
+    BYTES(
+        Solidity.types.keys.filter { it.contains("byte") }
```
rmeissner, line 56: What types is this? There is `bytes` and `byteX` they are very different

data/src/test/java/io/gnosis/data/adapters/DataMoshiTest.kt
```diff
@@ -119,6 +123,24 @@ class DataMoshiTest {
...
+
+        val params = paramAdapter.fromJson(jsonString)!!
+
+        assertEquals(6, params?.size)
+        assertTrue(params[0] is ParamDto.AddressParam)
+        assertTrue(params[1] is ParamDto.AddressParam)
+        assertTrue(params[2] is ParamDto.ArrayParam)
+        assertTrue(params[3] is ParamDto.ArrayParam)
+        assertTrue(params[4] is ParamDto.BytesParam)
+        assertTrue(params[5] is ParamDto.BytesParam)
```
rmeissner, line 138: Should the content be checked too?

data/src/main/java/io/gnosis/data/backend/dto/TransactionDto.kt
```diff
@@ -1,16 +1,60 @@
...
+        override val type: String,
+        override val name: String,
+        override val value: List<Any>
+    ) : ParamDto()
+
+    data class BytesParam(
+        override val type: String,
+        override val name: String,
+        override val value: String,
+        val valueDecoded: List<ValueDecodedDto>?
```
rmeissner, line 39: Value decoded ist not always a list ... It can be basically anything in this case it should fallback to null here. I would add a test for this

elgatovital: Does valueDecoded have the structure of ValueDecodedDto? So it is should be at least a ValueDecodedDto object or null? What do you mean by saying it can be anything? 
`valueDecoded` is nullable, so it should fallback to null if that''s what you mean.

rmeissner: If it is not a list of this object moshi will throw an error, you would need to catch that error in your custom adapter and set it to null.

And it can be literally anything ... E.g. `Any`

rmeissner: To be fair this is not used in v1 of the data decoding epic afaik, so you could also just igbore it

elgatovital: Done (setting valueDecoded to null if not a list). I will add tests

data/src/main/java/io/gnosis/data/adapters/ParamAdapter.kt
```diff
@@ -85,9 +85,15 @@ class ParamAdapter {
                                 type == "bytes" -> {
                                     val value = reader.nextString()
                                     if(reader.peek() ==  JsonReader.Token.NAME && reader.nextName() == "valueDecoded") {
-                                        val valueDecoded = valueDecodedAdapter.fromJson(reader)
-                                        reader.endObject()
-                                        return ParamDto.BytesParam(type, name, value, valueDecoded)
+                                        if(reader.peek() == JsonReader.Token.BEGIN_ARRAY) {
+                                            val valueDecoded = valueDecodedAdapter.fromJson(reader)
```
rmeissner, line 89: This will still throw if it is not the expected struct

data/src/main/java/io/gnosis/data/backend/dto/TransactionDto.kt
elgatovital: Even if valueDecoded could be Any we would care only about specific cases on the client-side, otherwise, it would be hard to parse/display such data. Is this correct?

rmeissner: In v1 we dont care about the decoded value, see https://github.com/gnosis/safe/issues/329

elgatovital: sure, but what will we do for the next versions, where we will sho multisend data? Specify specific polymorphic subtypes?

rmeissner: That is something you should worry about in the next version not now

elgatovital: I would like to anticipate change. That's what good software is about

elgatovital: but for now, everything is in place and #329 can be implemented

rmeissner: >I would like to anticipate change. That's what good software is about

You are implementing logic for features that have not been fully speced out and have little impact on this feature. You could leave out the field and it would have 0 impact here. This is overthinking and slows down development. I do agree that there is a fine line and that you should check how future potential changes can impact current implementation (that is why I mentioned it should fallback to null in an error case). The rest is unknown ... as the issues have not been discussed in detail. Good software *development* is not only about thinking about edge cases, but also about *not overspecifying* (and knowing when not to do things). 

The client gateway returns it, because it is used in that form by the web client, but maybe we decide while discussing v2 of data decoding that we want to use a different structure and do some preprocessing on the client gateway (who knows).

elgatovital: good point

data/src/main/java/io/gnosis/data/backend/dto/TransactionDto.kt
```diff
@@ -1,16 +1,60 @@
...
+    data class AddressParam(
+        override val type: String,
+        override val name: String,
+        override val value: Solidity.Address
+    ) : ParamDto()
+
+    data class ArrayParam(
+        override val type: String,
+        override val name: String,
+        override val value: List<Any>
```
jpalvarezl, line 32: Does this have to be `Any` ? I think that for the sake of regularity it would make more sense to leave it as a `String`

jpalvarezl: Would also be easier to put into the UI

data/src/main/java/io/gnosis/data/backend/dto/TransactionDto.kt
```diff
@@ -1,16 +1,60 @@
...
-    val value: Any
+data class ValueDecodedDto(
+    val operation: Operation,
+    val to: Solidity.Address,
+    val value: Long,
+    val data: String,
+    val dataDecoded: DataDecodedDto?
 )
 
+sealed class ParamDto {
```
jpalvarezl, line 18: Would maybe be better to have a type parameter for `value` that is later defined foreach concretisation. Although, it is either `String`, `List<String>` or `Address` as far as I could see.

data/src/main/java/io/gnosis/data/backend/dto/TransactionDto.kt
```diff
@@ -1,16 +1,60 @@
...
+        override val value: Solidity.Address
+    ) : ParamDto()
+
+    data class ArrayParam(
+        override val type: String,
+        override val name: String,
+        override val value: List<Any>
+    ) : ParamDto()
+
+    data class BytesParam(
```
jpalvarezl, line 35: Not sure I see the point for `BytesParam` if you already have a `ValueParam` that is a flat scalar value.

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
```diff
@@ -227,13 +227,16 @@ class TransactionRepository(
...
 }
 
-fun List<ParamsDto>?.getValueByName(name: String): String? {
-    this?.map {
-        if (it.name == name && it.value is String) {
-            return (it.value as String)
-        }
-    }
-    return null
+fun List<ParamDto>?.getAddressValueByName(name: String): Solidity.Address? {
```
jpalvarezl, line 230: Would get rid of the `?` for the extension function as it doesn't add value, AFAICT. Some for the method below 👇

data/src/test/java/io/gnosis/data/adapters/DataMoshiTest.kt
```diff
@@ -119,6 +127,61 @@ class DataMoshiTest {
         assertEquals(expected, transactionDto)
     }
 
+    @Test
+    fun `fromJson (param list) - should map to correct param types`() {
```
jpalvarezl, line 131: What happens if the backend returns an empty non-null list?

jpalvarezl: I would say it is worth having a test for that. Asserting against `emptyList`

data/src/main/java/io/gnosis/data/backend/dto/TransactionDto.kt
elgatovital: It can be list of list of list ....

data/src/main/java/io/gnosis/data/backend/dto/TransactionDto.kt
elgatovital: BytesParam is for "bytes" type and it can have valueDecoded field. ValueParam only has value field for a flat scalar value

data/src/main/java/io/gnosis/data/backend/dto/TransactionDto.kt
jpalvarezl: What we did in the gateway was to have something like a sealed class called `ParamValue` and 2 subclasses `SingleValue` and `ArrayValue`. Then you could just have `val value: ParamValue` and the type gets resolve upon deserialization.

data/src/main/java/io/gnosis/data/backend/dto/TransactionDto.kt
elgatovital: for v1 it should be fine I think. And we only deal with Strings for ui purposes

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
elgatovital: we would have to use nullablility checks everywhere we use this function

data/src/main/java/io/gnosis/data/backend/dto/TransactionDto.kt
elgatovital: Unfortunately, moshi needs to know the concrete type or have a concrete polymorphic adapter registered. We also cannot register polymorphic adapters for all solidity type combinations

jpalvarezl (approved):

Issue #905, opened 2020-09-14
DataDecoded: implement the "action" in transaction details and opening data decoded screen with tracking


DmitryBespalov:
- Action displayed in transaction details screen
- Action opens data decoded screen with actual parameters
- Tracking is implemented for the action details screen (see the epic)

liliya-soroka:
Verified
app version 2.1.0 -1172

Issue #904, opened 2020-09-14
DataDecoding: UI for arrays and nested arrays


DmitryBespalov:
Display arrays of values and nested arrays of values

To be refined:
- Indentation of the nested arrays; arrays are collapsible
- limit the nesting levels to 10

Designs 
- web: https://zpl.io/amplpRQ
- app: [zpl://screen?pid=5e9efc30726e106958cc43de&sid=5f4e3d1cef6167b00769a540](zpl://screen?pid=5e9efc30726e106958cc43de&sid=5f4e3d1cef6167b00769a540)

liliya-soroka:
waiting for ABI from Uxio to test nested array

liliya-soroka:
Verified
app version 2.3.0 - 1277rc

Issue #902, opened 2020-09-14
DataDecoding: UI for simple parameter types


DmitryBespalov:
Different UI for:
- Address
- Bytes (hexadecimal), collapsible/expandable
- Other types of values (non arrays), like integers, bool, or string values

See designs: https://zpl.io/amplpRQ

liliya-soroka:
Verified
app version 2.2.0 -1259

PR #923, opened 2020-09-17
Add transaction details action screen

elgatovital:
Handles #905
Handles #902
Handles #904

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/settings/view/AddressItem.kt
```diff
@@ -21,7 +21,7 @@ class AddressItem @JvmOverloads constructor(
     defStyleAttr: Int = 0
 ) : ConstraintLayout(context, attrs, defStyleAttr) {
 
-    private val binding by lazy { ViewAddressItemBinding.inflate(LayoutInflater.from(context), this) }
```
biafra23, line 24: What was/is the problem with the lazy initialization? Should we change this in other places, too?

elgatovital: No, I have changed it back. Better to have it as lazy, so that inflation happens deferred

app/src/main/res/layout/view_labeled_address_item.xml
```diff
@@ -4,65 +4,28 @@
...
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
-    android:layout_height="68dp"
+    android:layout_height="wrap_content"
     tools:parentTag="androidx.constraintlayout.widget.ConstraintLayout">
 
     <TextView
-        android:id="@+id/label"
-        style="@style/TextDark.Bold"
+        android:id="@+id/name"
```
biafra23, line 11: Why is this called `name` and not `label` as the file name suggests? And in `view_named_address_item.xml` the id is called `label` instead of `name`.

data/src/main/java/io/gnosis/data/backend/dto/TransactionDto.kt
```diff
@@ -30,7 +31,14 @@ sealed class ParamDto {
         override val type: String,
         override val name: String,
         override val value: List<Any>
-    ) : ParamDto()
+    ) : ParamDto() {
+
+        fun getItemType(): ParamType = when (type.split("[")[0]) {
```
biafra23, line 36: Just to clarify: This could me more than one type. ATM we only look at the first to determine the array param type?

elgatovital: we just need to know if it's an array

app/src/main/res/layout/view_named_address_item.xml
```diff
@@ -0,0 +1,68 @@
...
+<merge
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="68dp"
+    tools:parentTag="androidx.constraintlayout.widget.ConstraintLayout">
+
+    <TextView
+        android:id="@+id/label"
```
biafra23, line 11: One more thing :-) Could this be called `name` instead? For consistency?

biafra23 (approved):
LGTM 👍

app/src/main/res/layout/view_named_address_item.xml
elgatovital: wow.. you are meticulous today :) Sure, renaming

app/src/main/java/io/gnosis/safe/ui/transactions/details/view/LabeledArrayItem.kt
```diff
@@ -65,38 +65,38 @@ class LabeledArrayItem @JvmOverloads constructor(
...
+                addressItem.address = value as Solidity.Address
                 container.addView(addressItem)
             }
             ParamType.BYTES -> {
                 val bytesItem = TxDataView(context)
                 val layoutParams = LinearLayout.LayoutParams(ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.WRAP_CONTENT)
                 layoutParams.setMargins(0, dpToPx(6), 0, 0)
                 bytesItem.layoutParams = layoutParams
-                bytesItem.setData(value, value.length / 2)
+                bytesItem.setData(value as String, value.length / 2)
```
biafra23, line 91: The length calculation might be off, if there is a `0x` prefix. Assuming value is already a hex string.

Issue #792, opened 2020-08-17
NFT details (action part for NFT's)


jpalvarezl:
Design:https://zpl.io/VDyzLp3 & https://zpl.io/bLMwYBD

![image](https://user-images.githubusercontent.com/246473/92596070-800bbd00-f2a5-11ea-8c54-a3d48be0a089.png)

biafra23:
@jpalvarezl Does the client gateway send the token description in the tx list?

jpalvarezl:
There is not going to be a client gateway endpoint for this, at least not in the foreseeable future. The only logic that the clients should implement is to aggregate by `tokenAddress` and sort alphabetically by `tokenName`, as far as I understood. It was deemed in the refinement meeting that this was not a priority to push to the gateway for now.

liliya-soroka:
Reopened:
The item details should be instead of group details on the tx details page
Current result: ![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/0b660ea2-9fb3-4311-8fba-de5421715d3f)
Expected result: https://app.zeplin.io/project/5e9efc30726e106958cc43de/screen/5ee8ddc8e175644aa196ad65
safe : 0x1230B3d59858296A31053C1b8562Ecf89A2f888b
tx nonce : 234

biafra23:
>  The item details should be instead of group details on the tx details page

I don't understand, what you mean by that.

tschubotz:
Let's remove "NFT action" from the requirements https://gnosisinc.slack.com/archives/C63LQTGNB/p1600077831459800?thread_ts=1600072639.442800&cid=C63LQTGNB

liliya-soroka:
Verified
app version 2.2.0 -1174rc

PR #926, opened 2020-09-21
Show amount (+/-1 <SYMBOL>) instead of name for NFTs

biafra23:
Regading #792 .

- We don't show the token name only amount (and symbol) and tokenId below 
See: https://zpl.io/2GKm9xE and https://zpl.io/VDyzLp3

ENS | LBB in   | LBB out
---- | -------- | --------
Screenshot_20200921-160542 | Screenshot_20200921-155704 | Screenshot_20200921-160518

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/details/view/Erc721View.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/details/view/Erc721View.kt
@@ -7,6 +7,7 @@ import androidx.constraintlayout.widget.ConstraintLayout
 import io.gnosis.safe.R
 import io.gnosis.safe.databinding.ViewTxErc721Binding
 import io.gnosis.safe.utils.loadTokenLogo
+import pm.gnosis.svalinn.common.utils.getColorCompat
 
 class Erc721View @JvmOverloads constructor(
     context: Context,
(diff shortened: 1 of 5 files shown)


app/src/main/java/io/gnosis/safe/ui/transactions/details/view/Erc721View.kt
```diff
@@ -16,9 +17,16 @@ class Erc721View @JvmOverloads constructor(
 
     private val binding by lazy { ViewTxErc721Binding.inflate(LayoutInflater.from(context), this) }
 
-    fun setToken(logoUri: String, tokenId: String?, tokenName: String?, tokenDescription: String?) {
-        binding.tokenName.text = tokenName
+    fun setToken(logoUri: String, tokenId: String?, outgoing: Boolean, amount: String?) {
         binding.logo.loadTokenLogo(logoUri, R.drawable.ic_nft_placeholder)
         binding.tokenId.text = tokenId
```
elgatovital, line 22: with(binding) { ... would look nicer

elgatovital (approved):

app/src/main/java/io/gnosis/safe/ui/transactions/details/view/Erc721View.kt
biafra23: changed in 21a7ced2831d447598816cd2a3cc59b65a8b4575

rmeissner:
Shouldn't the amount (+/- 1) and the token id, be aligned with the from address and asset address

elgatovital (approved):

Issue #817, opened 2020-08-24
[Android] Call is missed in the name of type of operation on Advanced details page "delegateCall" 


liliya-soroka:
**Describe the bug**
There is "delegate" instead of "delegateCall" on the Advanced details page in the Type of operation 

**To Reproduce**
1. Add save 0x938bae50a210b80EA233112800Cd5Bc2e7644300
2. Open transaction with nonce 57
3. Open advanced

**Expected behavior**
"delegateCall" should be used instead of "delegate" in the type of operation field

**Smartphone (please complete the following information):**
 - Device: Pixel 4
 - Android Version: 10
 - App Version 1.12.1 - 93

**Additional context**

liliya-soroka:
Verified
app version 2.1.0 -1105

PR #929, opened 2020-09-21
Add displayName() to Operation enum

biafra23:
Regarding #817 .

- Add displayName to Operation and use it

@gnosis/mobile-devs

### DIFF
--- a/data/src/main/java/io/gnosis/data/backend/dto/TransactionDto.kt
+++ b/data/src/main/java/io/gnosis/data/backend/dto/TransactionDto.kt
@@ -45,7 +45,7 @@ sealed class ParamDto {
         override val value: Any
     ) : ParamDto()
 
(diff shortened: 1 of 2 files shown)


data/src/main/java/io/gnosis/data/backend/dto/TransactionDto.kt
```diff
@@ -57,5 +57,11 @@ sealed class ParamDto {
 
 enum class Operation(val id: Int) {
```
elgatovital, line 58: I would not assume any knowledge about UI layer here. Better to have an extension function in the fragment

biafra23: Done in 4ed1267d9732e101ba665de5cdec3515f1817a7a

elgatovital (approved):

PR #933, opened 2020-09-22
Activate LeakCanary for debug builds

### LINKED ISSUE #868: Check for memory leaks
After the public release we could check with https://github.com/square/leakcanary/ in debug builds

biafra23:
Regarding #868 .

- Activate Leak Canary for buildType `debug`
- This installs a separate app called "Leaks" to access previously detected leaks.

See https://square.github.io/leakcanary/fundamentals-how-leakcanary-works/ 

Remarks: QA can only check that Leak Canary is not active in any of the builds. 

@gnosis/mobile-devs


jpalvarezl (approved):
Provided that LeakCanary doesn't show up in prod this LGTM

PR #937, opened 2020-09-24
Add serializer for params

elgatovital:
- Fix crash due to marshaling problem when moving app to background

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/utils/ParamSerializer.kt
```diff
@@ -0,0 +1,35 @@
...
+    init {
+        decodedDataAdapter = moshi.adapter(DataDecodedDto::class.java)
+        decodedValuesAdapter = moshi.adapter(Types.newParameterizedType(List::class.java, ValueDecodedDto::class.java))
+    }
+
+    fun serializeDecodedData(decodedData: DataDecodedDto): String {
+        return decodedDataAdapter.toJson(decodedData)
+    }
+
+    fun unserializeDecodedData(decodedDataString: String): DataDecodedDto? {
```
biafra23, line 24: I would call it `deserializeDecodedData()` and `deserializeDecodedValues()` but that's not a blocker.

app/src/main/res/navigation/main_nav.xml
```diff
@@ -176,7 +176,7 @@
 
         <argument
             android:name="decodedData"
-            app:argType="java.io.Serializable" />
```
biafra23, line 179: Does that mean we cannot use Serializable arguments in general? Or is it a problem in this specific case only, and if so, why?

data/src/main/java/io/gnosis/data/backend/dto/TransactionDto.kt
```diff
@@ -1,17 +1,17 @@
...
     val method: String,
     val parameters: List<ParamDto>?
-) : Serializable
+)
 
 data class ValueDecodedDto(
     val operation: Operation,
     val to: Solidity.Address,
-    val value: Long,
+    val value: BigInteger,
```
biafra23, line 14: 👍

biafra23 (approved):
LGTM 

Why did Serialize not work? Is it a general problem? Why would Android offer that method if it doesn't survive the lifecycle? Would Parcelable work or fail for the same reason?

elgatovital:
> LGTM
> 
> Why did Serialize not work? Is it a general problem? Why would Android offer that method if it doesn't survive the lifecycle? Would Parcelable work or fail for the same reason?

I haven't investigated it in detail. Probably some child elements were not serializable.

PR #938, opened 2020-09-24
Disable synthetics

### LINKED ISSUE #789: Disable Kotlin Synthetics
We decided to go with ViewwBindings instead of Kotlin synthetics. Some classes still use synthetics.

Synthetics can be disabled via 

```
androidExtensions {
    features = ["parcelize"]
}
```

in app/build.gradle. In doing so, the build will break for several files still using synthetics. Those need to be migrated to ViewBindings. 

```
e: app/src/main/java/io/gnosis/safe/helpers/AddressInputHelper.kt
e: app/src/main/java/io/gnosis/safe/ui/safe/add/AddSafeNameFragment.kt
e: app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
e: app/src/main/java/io/gnosis/safe/utils/DialogBuilders.kt
```

biafra23:
Closes #789 .

- Disable Synthetics
- Convert remaining usages to ViewBindings 
- Housekeeping

@gnosis/mobile-devs

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -156,7 +156,7 @@ android {
 
 androidExtensions {
     experimental = true
-    features = ["parcelize"]
+    features = []
 }
 
 dependencies {
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -156,6 +156,7 @@ android {
 
 androidExtensions {
     experimental = true
+    features = ["parcelize"]
 }
 
 dependencies {
--- a/app/src/main/java/io/gnosis/safe/helpers/AddressInputHelper.kt
+++ b/app/src/main/java/io/gnosis/safe/helpers/AddressInputHelper.kt
@@ -14,15 +14,13 @@ import io.gnosis.safe.utils.handleAddressBookResult
 import io.gnosis.safe.utils.handleQrCodeActivityResult
 import io.gnosis.safe.utils.parseEthereumAddress
 import pm.gnosis.model.Solidity
-import pm.gnosis.svalinn.common.utils.visible
 import pm.gnosis.utils.exceptions.InvalidAddressException
 
 class AddressInputHelper(
     fragment: BaseFragment,
     tracker: Tracker,
     private val addressCallback: (Solidity.Address) -> Unit,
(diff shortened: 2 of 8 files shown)


app/src/main/java/io/gnosis/safe/helpers/AddressInputHelper.kt
```diff
@@ -30,37 +28,41 @@ class AddressInputHelper(
     private val dialog =
         BottomSheetDialog(fragment.requireContext()).apply {
             val clipboard = fragment.activity?.getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager
+            val binding = BottomSheetAddressInputBinding.inflate(layoutInflater)
+            with(binding) {
+                setContentView(root)
+
+                bottomSheetAddressInputBook.visible(allowAddressBook)
```
biafra23, line 35: Should I remove disabled AddressBook support?

elgatovital:
Where exactly are synthetics disabled?

app/build.gradle
```diff
@@ -156,6 +156,7 @@ android {
 
 androidExtensions {
     experimental = true
+    features = ["parcelize"]
```
elgatovital, line 159: we are not using @Parcelize yet

biafra23: So should we disable parcelize? This disables synthetics, by not mentioning them.

elgatovital: this was also my question. So if you are not listing features explicitly everything is enabled?

app/src/main/java/io/gnosis/safe/helpers/AddressInputHelper.kt
elgatovital: yes, we don't have ux for it and it might look different in the future. Better to add it when we need it

biafra23:
> Where exactly are synthetics disabled?

In app/build.gradle:
```
androidExtensions {
    experimental = true
    features = []
}
```

elgatovital (approved):

PR #939, opened 2020-09-24
Show decoded multisend data

elgatovital:
- Add custom view for multisend action item
- Extend nav graph

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsActionFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsActionFragment.kt
@@ -12,6 +12,7 @@ import androidx.navigation.fragment.navArgs
 import io.gnosis.data.backend.dto.DataDecodedDto
 import io.gnosis.data.backend.dto.ParamDto
 import io.gnosis.data.backend.dto.ParamType
+import io.gnosis.safe.R
 import io.gnosis.safe.ScreenId
 import io.gnosis.safe.databinding.FragmentTransactionDetailsActionBinding
 import io.gnosis.safe.di.components.ViewComponent
@@ -30,7 +31,8 @@ class TransactionDetailsActionFragment : BaseViewBindingFragment<FragmentTransac
 
     private val navArgs by navArgs<TransactionDetailsActionFragmentArgs>()
 
-    private val decodedData by lazy { paramSerializer.deserializeDecodedData(navArgs.decodedData) }
+    private val data by lazy { navArgs.data }
+    private val decodedData by lazy { navArgs.decodedData?.let { paramSerializer.deserializeDecodedData(it) } }
     private val address by lazy { navArgs.address?.asEthereumAddress() }
     private val amount by lazy { navArgs.amount }
 
@@ -63,6 +65,9 @@ class TransactionDetailsActionFragment : BaseViewBindingFragment<FragmentTransac
                 content.addView(getTransferItem(it, amount ?: ""))
             }
         }
+
+        //TODO: add separators
+        binding.content.addView(getDataItem(getString(R.string.tx_details_data), data))
 
         decodedDto?.let {
             with(binding) {
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsActionFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsActionFragment.kt
@@ -32,6 +32,7 @@ class TransactionDetailsActionFragment : BaseViewBindingFragment<FragmentTransac
 
     private val navArgs by navArgs<TransactionDetailsActionFragmentArgs>()
 
+    private val action by lazy { navArgs.action }
     private val data by lazy { navArgs.data }
     private val decodedData by lazy { navArgs.decodedData?.let { paramSerializer.deserializeDecodedData(it) } }
     private val address by lazy { navArgs.address?.asEthereumAddress() }
@@ -50,6 +51,7 @@ class TransactionDetailsActionFragment : BaseViewBindingFragment<FragmentTransac
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
         with(binding) {
+            title.text = action
             backButton.setOnClickListener {
                 findNavController().navigateUp()
             }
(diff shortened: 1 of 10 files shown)


app/src/main/res/layout/fragment_transaction_details_action_multisend.xml
```diff
@@ -0,0 +1,62 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.coordinatorlayout.widget.CoordinatorLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:background="@color/white">
+
+    <com.google.android.material.appbar.AppBarLayout
```
biafra23, line 10: It seems we have this same snippet in several place only with different title texts. Can we replace this with a custom view? Or an include if it is possible to parametrise them. This change is probably outside the scope of this PR.

elgatovital: we can think about it at some later point. With view binding, we do not have a possibility to pass arguments to the included layout. I am against enabling data binding as it will increase compile time. A custom view could be an option. Right now I don't mind this small duplication as it makes layouts more explicit and flexible.

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsActionFragment.kt
```diff
@@ -80,6 +99,15 @@ class TransactionDetailsActionFragment : BaseViewBindingFragment<FragmentTransac
         }
     }
 
+    private fun getTransferItem(address: Solidity.Address, amount: String): TxTransferActionView {
+        val item = TxTransferActionView(requireContext())
+        val layoutParams = LayoutParams(MATCH_PARENT, WRAP_CONTENT)
+        layoutParams.setMargins(0, dpToPx(16), 0, -dpToPx(8))
```
jpalvarezl, line 105: negative margin? seems a bit hacky, what are you trying to do with that?

app/src/main/java/io/gnosis/safe/ui/transactions/details/view/MultisendActionView.kt
```diff
@@ -0,0 +1,27 @@
...
+import android.content.Context
+import android.util.AttributeSet
+import android.view.LayoutInflater
+import androidx.constraintlayout.widget.ConstraintLayout
+import io.gnosis.safe.databinding.ViewMultisendActionBinding
+import io.gnosis.safe.utils.abbreviateEthAddress
+import pm.gnosis.crypto.utils.asEthereumAddressChecksumString
+import pm.gnosis.model.Solidity
+
+class MultisendActionView @JvmOverloads constructor(
```
jpalvarezl, line 12: Should we not be using `LabeledAddressItem`? I think we are creating too many custom views, maybe we should start trying to generalise a bit more?

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsActionFragment.kt
```diff
@@ -118,4 +146,13 @@ class TransactionDetailsActionFragment : BaseViewBindingFragment<FragmentTransac
...
         return item
     }
+
+    private fun getDivider(): View {
+        val item = View(requireContext())
+        val layoutParams = LayoutParams(MATCH_PARENT, dpToPx(1))
+        layoutParams.setMargins(0, dpToPx(16), 0, 0)
+        item.layoutParams = layoutParams
+        item.setBackgroundColor(ContextCompat.getColor(requireContext(), R.color.light_grey))
+        return item
```
jpalvarezl, line 156: maybe using `View(requireContext()).apply { ... }` would be a bit more readable.

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsActionMultisendFragment.kt
```diff
@@ -0,0 +1,89 @@
...
+    override fun screenId() = ScreenId.TRANSACTIONS_DETAILS_ACTION_LIST
+
+    private val navArgs by navArgs<TransactionDetailsActionMultisendFragmentArgs>()
+    private val decodedValues by lazy { paramSerializer.deserializeDecodedValues(navArgs.decodedValues) }
+
+    @Inject
+    lateinit var balanceFormatter: BalanceFormatter
+
+    @Inject
+    lateinit var paramSerializer: ParamSerializer
```
jpalvarezl, line 34: I haven't seen what this class looks like, but couldn't we just use `@Parcelize` on `DecodedValue` class?

app/src/main/java/io/gnosis/safe/ui/transactions/details/view/MultisendActionView.kt
elgatovital: This view has different layout and shows not only blockie, address, but also method and potentially (in v3) tx type icon. I think we have custom views that are specific enough for their task and I would not generalize them at this point in time

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsActionFragment.kt
elgatovital: TxTransferAction has an internal margin that is too big for decoded action screen.

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsActionFragment.kt
elgatovital: maybe :)

elgatovital: you'll have to cast layoutParams then in order to set margins. I would leave it as is. IMO it is straightforward and easy to read.

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsActionMultisendFragment.kt
elgatovital: some parameter types are not directly supported by parcelize

jpalvarezl (approved):

Issue #911, opened 2020-09-15
[Android] Copy to clipboard doesn't work for contract address on the safe settings


liliya-soroka:
**Describe the bug**
Copy to clipboard doesn't work for contract address on the safe settings

**To Reproduce**
1. Go to safe settings
2. Click on contract address
Current result: The contract address is opened on etherscan

**Expected behavior**
Tap on contract address should call copy to clipboard for address

**Smartphone (please complete the following information):**
 - Device: Pixel 4
 - Android Version: 10.1
 - App Version 2.0.0 ( 19)
- env staging rinkeby

**Environment**

**Additional context**

liliya-soroka:
Verified
app version 2.1.0 - 1160

PR #942, opened 2020-09-25
Make address copiable to clipboard for MasterCopyItem

elgatovital:
Handles #911 

- Copy address to clipboard on click
- Disable leak canary

@gnosis/mobile-devs

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -260,8 +260,7 @@ dependencies {
 
     coreLibraryDesugaring "com.android.tools:desugar_jdk_libs:$versions.desugar_jdk_libs"
 
-    debugImplementation "com.squareup.leakcanary:leakcanary-android:$versions.leak_canary"
-
+    //debugImplementation "com.squareup.leakcanary:leakcanary-android:$versions.leak_canary"
 }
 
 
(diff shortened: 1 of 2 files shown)


rmeissner:
I would love a comment in the PR description on why leak canary was deactivated (even so it was just added yesterday) ;)

elgatovital:
> I would love a comment in the PR description on why leak canary was deactivated (even so it was just added yesterday) ;)

sure. That is something I wanted to discuss. In my understanding memory leak deteftion was a profiling task. Wether it is leak canary or something else is irrelevant. It is fine to use leak canary but I don't want profiler being enabled all the time on all of my devices. We should profile app from time to time though.

rmeissner:
In my opinion this makes sense, and theoretically you could have a separate build type for that, right? (e.g. like profiling and make it a `profilingImplementation` dependency)

elgatovital:
> In my opinion this makes sense, and theoretically you could have a separate build type for that, right? (e.g. like profiling and make it a `profilingImplementation` dependency)

yes, this would be better than having it for dev builds

rmeissner (approved):

Issue #917, opened 2020-09-16
[Android]Txs list: The Set FallbackHandler and Change Master Copy recprds are not grey out in cancelled tx


liliya-soroka:
**Describe the bug**
The whole record is greyed out if the tx was cancelled except the SetFallbackHandler and ChangeMasterCopy

**To Reproduce**
1. Go to 0x8675B754342754A30A2AeF474D114d8460bca19b ( prod mainnet)
2. Open txs list
3. Scroll to the cancelled SetFallback and Change masterCopy txs

**Expected behavior**
The text should be greyed out the same was as the whole record in the txs list

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/90b2ad2b-1f27-49eb-bc1c-eeb75dbde2fa)

**Smartphone (please complete the following information):**
 - Device: Pixel 4
 - Android Version: 10.1
 - App Version 2.0.0 (23)

**Environment**
Prod Mainnet
**Additional context**

liliya-soroka:
Verified
app version 2.1.0 - 1172
staging, rinkeby

PR #943, opened 2020-09-25
Grey out label for canceled transaction

elgatovital:
Handles #917

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewHolderFactory.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewHolderFactory.kt
@@ -207,6 +207,7 @@ class ChangeMastercopyViewHolder(private val viewBinding: ItemTxChangeMastercopy
             finalStatus.alpha = OPACITY_FULL
             dateTime.alpha = viewTransfer.alpha
             version.alpha = viewTransfer.alpha
+            label.alpha = viewTransfer.alpha
             blockies.alpha = viewTransfer.alpha
             ellipsizedAddress.alpha = viewTransfer.alpha
             nonce.text = viewTransfer.nonce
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewHolderFactory.kt
```diff
@@ -207,6 +207,7 @@ class ChangeMastercopyViewHolder(private val viewBinding: ItemTxChangeMastercopy
             finalStatus.alpha = OPACITY_FULL
             dateTime.alpha = viewTransfer.alpha
             version.alpha = viewTransfer.alpha
+            label.alpha = viewTransfer.alpha
```
biafra23, line 210: Shouldn't nonce and moduleAddress also have the alpha set? Maybe reorder, so that we have three blocks: texts, alphas and visibility.

biafra23:
It seems the nonce is never having its alpha value set.

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewHolderFactory.kt
biafra23: IMHO txTypeIcon should also be greyed out: 
`txTypeIcon.alpha = viewTransfer.alpha`

biafra23: **Not the chevron!** Sorry for the confusion. 
But the txTyoeIcon needs to be alpha'ed too.

biafra23: And there is no example wit a nonce in the design. So I will ask Sasha.

biafra23 (approved):
👍

Issue #953, opened 2020-09-29
Key derivation & import


elgatovital:
- Calculate seed phrase from mnemonic
- Derive accounts from seed phrase
- Provide pages of accounts for account selection screen

elgatovital:
This proof of concept generates the first three owner addresses from a given Mnemonic phrase. It also generates the secret keys, but I haven't checked those, yet.

```
        val bip39: Bip39 = Bip39Generator(AndroidWordListProvider(this))
        val validatedMnemonic = bip39.validateMnemonic("creek banner employ mix teach sunny sure mutual pole mom either lion")
        Timber.i("---> ValidatedMnemonic: $validatedMnemonic")

        val mnemonicSeed = bip39.mnemonicToSeed(validatedMnemonic)
        Timber.i("---> Seed: ${mnemonicSeed.toHexString()}")

        val hdNode = KeyGenerator.masterNode(ByteString.of(*mnemonicSeed))
        val masterKey = hdNode.derive(KeyGenerator.BIP44_PATH_ETHEREUM)
        (0L..2L).map {
            val key = masterKey.deriveChild(it).keyPair
            val privateKey = key.privKeyBytes ?: throw IllegalStateException("Private key must not be null")
            val address = key.address.asBigInteger()
            Timber.i("---> Owner: ${Solidity.Address(address).asEthereumAddressChecksumString()}")
        }
```

biafra23:
This proof of concept generates the first three owner addresses from a given Mnemonic phrase. It also generates the secret keys, but I haven't checked those, yet.

```
        val bip39: Bip39 = Bip39Generator(AndroidWordListProvider(this))
        val validatedMnemonic = bip39.validateMnemonic("creek banner employ mix teach sunny sure mutual pole mom either lion")
        Timber.i("---> ValidatedMnemonic: $validatedMnemonic")

        val mnemonicSeed = bip39.mnemonicToSeed(validatedMnemonic)
        Timber.i("---> Seed: ${mnemonicSeed.toHexString()}")

        val hdNode = KeyGenerator.masterNode(ByteString.of(*mnemonicSeed))
        val masterKey = hdNode.derive(KeyGenerator.BIP44_PATH_ETHEREUM)
        (0L..2L).map {
            val key = masterKey.deriveChild(it).keyPair
            val privateKey = key.privKeyBytes ?: throw IllegalStateException("Private key must not be null")
            val address = key.address.asBigInteger()
[...]

PR #962, opened 2020-09-30
Address & Key derivation with tests

biafra23:
Handles #953 

- Derive owner addresses from Mnemonic seed
- Derive secret key from Mnemonic and index
-

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/utils/MnemonicAddressDerivation.kt
```diff
@@ -0,0 +1,53 @@
...
+
+import okio.ByteString
+import pm.gnosis.crypto.HDNode
+import pm.gnosis.crypto.KeyGenerator
+import pm.gnosis.mnemonic.Bip39
+import pm.gnosis.model.Solidity
+import pm.gnosis.utils.asBigInteger
+import java.math.BigInteger
+
+interface MnemonicAddressDerivation {
```
elgatovital, line 11: Derivation sounds like a verb and normally we use nouns for class-names

elgatovital: Derivation might represend one single derivation. It would be more correct to call this class MnemonicAddressDerivator

app/src/main/java/io/gnosis/safe/utils/MnemonicAddressDerivation.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    override fun initialize(mnemonic: String) {
+        val mnemonicSeed = bip39.mnemonicToSeed(mnemonic)
+        val hdNode = KeyGenerator.masterNode(ByteString.of(*mnemonicSeed))
+        masterKey = hdNode.derive(KeyGenerator.BIP44_PATH_ETHEREUM)
+    }
+
+    override fun addressesForRange(range: LongRange): List<Solidity.Address> {
+        val result = mutableListOf<Solidity.Address>()
+        (range).map {
```
elgatovital, line 36: do we need to enclose range in parentheses?

app/src/main/java/io/gnosis/safe/utils/MnemonicAddressDerivation.kt
biafra23: changed in 164da5b29bba2da94c71e9fe5ca461c071398f90

app/src/main/java/io/gnosis/safe/utils/MnemonicAddressDerivation.kt
biafra23: Changed in 164da5b29bba2da94c71e9fe5ca461c071398f90

elgatovital (approved):

Issue #719, opened 2020-07-09
[Android]DefaultFallbackHandler title is used even for unknow fallback handlers


liliya-soroka:
**Describe the bug**
DefaultFallbackHandler title is used even for unknow fallback handlers

**To Reproduce**
1. Change FallBack Handler to any address ( I used one of the owner's address)
2. Check created transaction

**Expected behavior**
"Unknown" should be used if it does not default fallback handler
**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/374142d0-bd96-4c9a-a44f-0111dd23b835)

**Smartphone (please complete the following information):**
 - Device: Pixel 4
 - Android Version: 10
 - App Version 0.4.2 - alpha-debug

**Additional context**

biafra23:
The default fallback handler address is 0xd5D82B6aDDc9027B22dCA772Aa68D5d74cdBdF44 and it is called DEFAULT_FALLBACK_HANDLER in the android app.

liliya-soroka:
Verified
app version 2.2.0 -1239

PR #963, opened 2020-09-30
Add workaround for displaying fallback handler version

elgatovital:
Handles #719

hardcoded string extraction will happen in scope of https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-android/964

@gnosis/mobile-devs

### DIFF
--- a/data/src/main/java/io/gnosis/data/repositories/SafeRepository.kt
+++ b/data/src/main/java/io/gnosis/data/repositories/SafeRepository.kt
@@ -93,6 +93,8 @@ class SafeRepository(
 
         val DEFAULT_FALLBACK_HANDLER = BuildConfig.DEFAULT_FALLBACK_HANDLER.asEthereumAddress()!!
         const val DEFAULT_FALLBACK_HANDLER_DISPLAY_STRING = "DefaultFallbackHandler"
+        const val DEFAULT_FALLBACK_HANDLER_UNKNOWN_DISPLAY_STRING = "Unknown"
+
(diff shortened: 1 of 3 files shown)


data/src/main/java/io/gnosis/data/repositories/SafeRepository.kt
```diff
@@ -93,6 +93,8 @@ class SafeRepository(
 
         val DEFAULT_FALLBACK_HANDLER = BuildConfig.DEFAULT_FALLBACK_HANDLER.asEthereumAddress()!!
         const val DEFAULT_FALLBACK_HANDLER_DISPLAY_STRING = "DefaultFallbackHandler"
+        const val DEFAULT_FALLBACK_HANDLER_UNKNOWN_DISPLAY_STRING = "Unknown"
+
         const val SAFE_MASTER_COPY_UNKNOWN_DISPLAY_STRING = "Unknown"
```
rmeissner, line 98: Why are we hardcoding display strings?

elgatovital: that is the reason I called it workaround. Here is quick fix that will allow users to see correct information. Harcoded strings will be dealt with in https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-android/964

biafra23 (approved):
LGTM 👍

Issue #946, opened 2020-09-28
Entry point to import owner key in App Settings


elgatovital:
On the App settings screen there is an “Import owner key” row, if there is no key imported yet.

Selecting it should lead to the “Import Owner Key”

<img src="https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/e6aa6330-bf6e-440f-87cc-b929eb4064ac" width=300/>

On the App settings screen, there is an “Imported owner key” row, if there is a key imported.

![image.png](https://images.zenhubusercontent.com/5c62952584111d4401fa50d5/d938453e-139a-4081-9a53-35ad0323d378)

liliya-soroka:
Verified
app version 2.2.0 - 1239

PR #968, opened 2020-10-01
Adds UI elements for navigation/removal for owner key interactions

jpalvarezl:
Handles #946 

- Adds UI elements in `App Settings` fragment
- Provides callbacks with `snackbars` for now to which the actions should be attached once they are implemented

@gnosis/mobile-devs


biafra23 (approved):
LGTM 👍

Issue #948, opened 2020-09-28
Select account screen 


elgatovital:
<img src="https://images.zenhubusercontent.com/5dea66701abe006b54a3a941/ed72d898-b449-482d-aa53-58e0246738c8" width=300/>

- Header scrolls with the content.

- The derivation shall use MetaMask HD path.
- Accounts start from # 1
- By default 20 keys are generated.
- There is “Show more” button after the last generated key in the list.
- Selecting “Show more” should generate additional 20 accounts. 
- The limit on displayed keys is 100.
- After the limit is reached, we do not display “Show more” button.

<img src="https://images.zenhubusercontent.com/5dea66701abe006b54a3a941/d217e42a-c37f-4bb7-b3cc-51abf4834406" width=200/>

**“Import” should:**
- Store the selected private key in the secure storage.
- Save the key address in the app storage.
- Lead to the App Settings screen

liliya-soroka:
Verified
app version 2.4.0 - 1324

PR #969, opened 2020-10-05
Select signing owner screen 

elgatovital:
Handles #948

- Add screen for selecting signing owner
- Paginated list of owners with potentially infinite scrolling capability
- Owner list adapter 
- Showing next page with owners only if `show more` is clicked
- Placeholder for storing selected owner key

@gnosis/mobile-devs


biafra23:
Showing the list can take a bit of time. I did not see a progress indicator, while waiting for the list. The layout has a ProgressBar, but I did not see it.

biafra23:
When scrolling beyond the ninth element the blockies icons shifts to the right. 

image
image

app/src/main/java/io/gnosis/safe/ui/signing/owners/OwnerListAdapter.kt
```diff
@@ -0,0 +1,116 @@
...
+
+class OwnerListAdapter() : PagingDataAdapter<Solidity.Address, RecyclerView.ViewHolder>(COMPARATOR) {
+
+    var pagesVisible = 1
+    private var selectedOwnerPosition: Int = 1
+
+    private var listener: WeakReference<OnOwnerItemClickedListener>? = null
+
+    fun setListener(listener: OnOwnerItemClickedListener) {
+        this.listener = WeakReference(listener)
```
biafra23, line 24: Refreshing my memory about WeakReferences...

biafra23: I think I got it. When the Fragment is garbage collected. So will the listener reference. 👍

biafra23:
There is no feedback when hitting "Show more" and no progress. I guess, that is not a blocker since it was probably not specified? Should we have at least a feedback that the "button" was pressed?

app/src/main/java/io/gnosis/safe/ui/signing/owners/OwnerPagingProvider.kt
```diff
@@ -0,0 +1,34 @@
...
+            initialKey = 0,
+            config = PagingConfig(
+                pageSize = PAGE_SIZE,
+                prefetchDistance = 1,
+                enablePlaceholders = false,
+                initialLoadSize = PAGE_SIZE,
+                maxSize = PAGE_SIZE * PAGES_THRESHOLD
+            ),
+            pagingSourceFactory = {
+                OwnerPagingSource(derivator, PAGES_THRESHOLD)
```
biafra23, line 25: I don't understand the word threshold here. The parameter is called maxPages. That makes more sense.

biafra23 (approved):
LGTM 👍

Issue #947, opened 2020-09-28
Import owner key screen


elgatovital:
- Scrolling should be available to support small screen resolutions.
- When the screen is opened the “Enter seed phrase” is focused by default.
- This field does not have any text input limitations.
- When selecting the “Next” button OR selecting “Enter” button on the keyboard, the form “submission” should happen.
  - Strip or ignore extra whitespace and punctuation symbols in the phrase. For example, words can be separated by more than one space or by new line symbol
  - Any letter case should be accepted (upper, lower, mixed)
  - Seed phrase validated according to underlying ETH/crypto library
  - Words should be lower cased
- If the input is valid, the next screen “Import Owner Key” should be presented.
- If the input is not valid, the error state should be displayed.

![image](https://user-images.githubusercontent.com/2284385/94420115-f8b6b880-0183-11eb-90b3-f359d220854b.png)

**ERROR STATE**
- In error state text input should be highlighted, and the “Seed phrase is not valid” message should be displayed.
- Typing in the text field should return it back to the “normal” state immediately.

![image](https://user-images.githubusercontent.com/2284385/94420194-18e67780-0184-11eb-8ad3-0e5d0e3a0e07.png)

Edit:
**Tracking**
Check seed phrase [screen names](https://github.com/gnosis/safe/wiki/Overview-screen-tracking)

Edit2:
**Designs**
https://zpl.io/Vkd0Xpg
https://zpl.io/amo0Md6

elgatovital:
@biafra23 move it to https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-android/953

biafra23:
This code validates the seed phrase ATM the result is unmodified parameter : 
```
val bip39: Bip39 = Bip39Generator(AndroidWordListProvider(this))
bip39.validateMnemonic("creek banner employ mix teach sunny sure mutual pole mom either lion")
```

liliya-soroka:
Verified
app version 2.4.0 - 1324

PR #973, opened 2020-10-05
Handles #947 Seed phrase input

jpalvarezl:
Handles #947 

- Adds `ImportOwnerKeyFragment` and `ViewModel`
- Validation according to (comment)[https://github.com/gnosis/safe-android/issues/947#issuecomment-700753569]
- Seed phrase `String` is cleaned up according to the task description

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/di/components/ApplicationComponent.kt
+++ b/app/src/main/java/io/gnosis/safe/di/components/ApplicationComponent.kt
@@ -21,8 +21,7 @@ import io.gnosis.safe.ui.terms.TermsChecker
 import io.gnosis.safe.utils.BalanceFormatter
 import io.gnosis.safe.utils.MnemonicKeyAndAddressDerivator
 import io.gnosis.safe.utils.ParamSerializer
-import pm.gnosis.mnemonic.Bip39Generator
-import pm.gnosis.mnemonic.android.AndroidWordListProvider
+import pm.gnosis.mnemonic.Bip39
 import javax.inject.Singleton
 
 @Singleton
@@ -73,6 +72,5 @@ interface ApplicationComponent {
 
     fun inject(fragment: BaseDialogFragment)
 
-    //    fun inject(service: BridgeService)
     fun inject(service: HeimdallFirebaseService)
 }
--- a/app/src/main/java/io/gnosis/safe/di/components/ApplicationComponent.kt
+++ b/app/src/main/java/io/gnosis/safe/di/components/ApplicationComponent.kt
@@ -19,9 +19,7 @@ import io.gnosis.safe.ui.base.fragment.BaseDialogFragment
 import io.gnosis.safe.ui.base.fragment.BaseFragment
 import io.gnosis.safe.ui.terms.TermsChecker
 import io.gnosis.safe.utils.BalanceFormatter
-import io.gnosis.safe.utils.MnemonicKeyAndAddressDerivator
 import io.gnosis.safe.utils.ParamSerializer
-import pm.gnosis.mnemonic.Bip39
 import javax.inject.Singleton
 
 @Singleton
@@ -62,8 +60,6 @@ interface ApplicationComponent {
 
     //TODO: remove
     fun safeRepository(): SafeRepository
-
-    fun mnemonicKeyAndAddressDerivator(): MnemonicKeyAndAddressDerivator
 
     // Base injects
     fun inject(activity: BaseActivity)
--- a/app/src/main/java/io/gnosis/safe/di/components/ApplicationComponent.kt
+++ b/app/src/main/java/io/gnosis/safe/di/components/ApplicationComponent.kt
@@ -21,6 +21,8 @@ import io.gnosis.safe.ui.terms.TermsChecker
 import io.gnosis.safe.utils.BalanceFormatter
 import io.gnosis.safe.utils.MnemonicKeyAndAddressDerivator
 import io.gnosis.safe.utils.ParamSerializer
+import pm.gnosis.mnemonic.Bip39Generator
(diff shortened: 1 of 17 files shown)


app/src/main/java/io/gnosis/safe/di/components/ApplicationComponent.kt
```diff
@@ -71,6 +73,6 @@ interface ApplicationComponent {
 
     fun inject(fragment: BaseDialogFragment)
 
-//    fun inject(service: BridgeService)
```
elgatovital, line 74: it is totally fine to delete this comment

app/src/main/java/io/gnosis/safe/di/components/ApplicationComponent.kt
```diff
@@ -21,6 +21,8 @@ import io.gnosis.safe.ui.terms.TermsChecker
 import io.gnosis.safe.utils.BalanceFormatter
 import io.gnosis.safe.utils.MnemonicKeyAndAddressDerivator
 import io.gnosis.safe.utils.ParamSerializer
+import pm.gnosis.mnemonic.Bip39Generator
```
elgatovital, line 24: unused imports

app/src/main/java/io/gnosis/safe/di/modules/RepositoryModule.kt
```diff
@@ -55,4 +59,9 @@ class RepositoryModule {
     @Singleton
     fun providesTransactionRepository(gatewayApi: GatewayApi): TransactionRepository =
         TransactionRepository(gatewayApi)
+
+    @Provides
+    @Singleton
+    fun providesBip39Generator(@ApplicationContext context: Context): Bip39Generator =
```
elgatovital, line 65: already provided in the ApplicationModule

app/src/main/java/io/gnosis/safe/ui/settings/owner/ImportOwnerKeyFragment.kt
```diff
@@ -0,0 +1,65 @@
...
+        viewModel.state.observe(viewLifecycleOwner, Observer { state ->
+            when (state) {
+                is ImportOwnerKeyState.Error -> {
+                    with(binding) {
+                        seedPhraseLayout.error = getString(R.string.enter_seed_phrase_error)
+                        seedPhraseLayout.isErrorEnabled = true
+                    }
+                }
+                is ImportOwnerKeyState.ValidSeedPhraseSubmitted -> {
+                    // TODO navigate to what goes next
```
elgatovital, line 49: findNavController().navigate(SettingsFragmentDirections.actionSettingsFragmentToOwnerSelectionFragment("insert seed phrase here"))

rmeissner: Be aware that passing the seed phrase in an intent might cause it to be persisted on device for some time (only if the app goes into background on that screen afaik) ;)

it is hard to find too many information on this and should probably only be an issue on rooted devices (see https://stackoverflow.com/questions/49759331/security-of-critical-data-passing-between-activities-via-intent-extras)

app/src/main/java/io/gnosis/safe/ui/settings/owner/ImportOwnerKeyViewModel.kt
```diff
@@ -0,0 +1,50 @@
...
+                        } else {
+                            ImportOwnerKeyState.Error(ViewAction.ShowError(InvalidSeedPhrase))
+                        }
+                    }
+                }
+            }
+    }
+
+    private fun cleanupSeedPhrase(seedPhrase: String): String {
+        return seedPhrase.split("\\s+?|\\p{Punct}+?".toRegex())
```
rmeissner, line 34: This is a nice addition, but please communicate this also to liliya and the ios team, so that it can be tested and is alligned between different clients. The specs say different things at different points.

Docs: `Before performing validation, strip whitespaces and make text lowercase (not visible to a user)`
Android issue: `Strip or ignore extra whitespace and punctuation symbols in the phrase.`

app/src/main/java/io/gnosis/safe/di/modules/RepositoryModule.kt
jpalvarezl: Have you tried using this dependency in a `viewModel` ? This dependency is not visible within the `ViewModelFactoryModule`.

elgatovital: If this generator is used only in viewModels we should remove provide statements from application module. 
```
 @Provides
    fun providesBip39(wordListProvider: WordListProvider): Bip39 = Bip39Generator(wordListProvider)

    @Provides
    fun providesWordListProvider(@ApplicationContext context: Context): WordListProvider = AndroidWordListProvider(context)

    @Provides
    fun providesMnemonicKeyAndAddressDerivator(bip39: Bip39): MnemonicKeyAndAddressDerivator = MnemonicKeyAndAddressDerivator(bip39)
```  could be moved to RepositoryModule then

app/src/main/java/io/gnosis/safe/ui/settings/owner/ImportOwnerKeyFragment.kt
jpalvarezl: Could we fork this into a new issue? I need to do rust work for the rest of the week. I think that a sensible solution would be to somehow share the viewModel between fragments and only keep the seed phrase on memory.

elgatovital (approved):

biafra23:
Is toLower() not in the task description anymore?

jpalvarezl:
It's part of the epic. I will update the issue too.

Issue #970, opened 2020-10-05
[Android] Crashes on Terms & Conditions (ActivityThread.java line 4195) 


liliya-soroka:
**Describe the bug**
The app is crashed after calling from background with opened T&C ( one more send to background before it required)
https://console.firebase.google.com/u/2/project/safe-firebase-mainnet/crashlytics/app/android:io.gnosis.safe/issues/1540c48a0e8188eb53486507839c3a7d?time=last-seven-days&sessionEventKey=5F74886A02D200013D9079EA7C0DB696_1456721774167076019
**To Reproduce**
1. Install the app
2. Launch app
3. Click Get started
4. Selec No thanks and send the app to the background 
5. wait for some time ( 1 min) and call the app from the background using the click on the app icon
Current result: The app is opened with opened "Terms and conditions"dialog. The strange thing that the dialog should be closed in step 
6. Send the app to the background with open Terms and Conditiona dialog
7. wait for some time
8. Call the app from the background
Current result: The app is crashed 
**Expected behavior**
After the step 5 the app should be opened with closed T&C dialog and the app should not crash 

**Smartphone (please complete the following information):**
 - Device: Pixel 4, Samsung A70
 - Android Version: [10
 - App Version 2.2.0 -1206

**Environment**
Prod mainnet

biafra23:
I can reproduce the crash by opening the app and just putting it in the background for a while. It usually crashes, when it was in the background for a minute.

biafra23:
This could be a way to fix it.

biafra23:
When the app is restarted (after it was in background) the event SplashViewModel.ShowTerms is received again. That's why the bottom sheet is reopened on app start. Apparently this event is sticky. It probably shouldn't be. 
The following is a workaournd in SplashActivity to prevent the crash:

```
                is SplashViewModel.ShowTerms -> {
                    termsBottomSheetDialog.apply {
                        onAgreeClickListener = {
                            viewModel.handleAgreeClicked()
                        }
                    }
                    if (!termsBottomSheetDialog.isAdded) {
                        termsBottomSheetDialog.show(supportFragmentManager, TermsBottomSheetDialog::class.simpleName)
                    }
                }
```

jpalvarezl:
Absolutely, agree. I would suggest coming up with an implementation of `BaseViewModel` that doesn't cache the last emitted event.

liliya-soroka:
Verified
app version 2.3.0 - 1278rc

PR #975, opened 2020-10-05
Prevent opening Terms Sheet if already added

biafra23:
Handles #970

- Prevents opening the Terms Sheet if it is already added
- Also does not re-open it (when it was dismissed) when coming back from background, because the last event is HideTerms in that case.

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/splash/SplashActivity.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/splash/SplashActivity.kt
@@ -40,7 +40,10 @@ class SplashActivity : BaseActivity() {
                         onAgreeClickListener = {
                             viewModel.handleAgreeClicked()
                         }
-                    }.show(supportFragmentManager, TermsBottomSheetDialog::class.simpleName)
+                    }
+                    if (!termsBottomSheetDialog.isAdded) {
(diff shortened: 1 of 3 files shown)


jpalvarezl (approved):
I would give this a try using the `SplashViewModel` inheriting from a `PublishViewModel` but if that doesn't do the trick, let's keep the proposed solution in this PR.

I will approve nonetheless so that you can merge at your earliest convenience.

elgatovital (changes requested):
what about handling the case when user clicks "no thanks", dismisses bottom sheet, and then navigates away from the app

jpalvarezl:
I am not sure where in the epic the expected behaviour is specified, but I think we are supposed to pester the users until we get an `Accept` 😆  not sure though.

elgatovital (approved):

elgatovital:
In case the user dismisses the dialog and navigates away it would be more intuitive to return to the page with the same state as it was before leaving the app: dialog dismissed and get started button is shown. So user has to click it to see the dialog again. In the current implementation user will return to the page with the dialog being dispayed already @jpalvarezl @biafra23

elgatovital (approved):

elgatovital:
lgtm 👍

Issue #951, opened 2020-09-29
Remove owner key: confirmation dialog


elgatovital:
Selecting the “remove” icon should display a confirmation.

![image.png](https://images.zenhubusercontent.com/5c62952584111d4401fa50d5/7f1e95fc-2e72-4056-abf8-18c14a8ab711)

On confirming:
- the key should be removed from secure storage.
- The key address should be removed from the app storage.

liliya-soroka:
verified
app version 2.3.0 - 1293

PR #976, opened 2020-10-06
Add remove owner confirmation dialog

elgatovital:
Handles #951

- Extract remove dialog to app utils
- Add confirmation dialog for removing owner key

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -189,7 +189,7 @@
 
     <string name="signing_owner_dialog_remove">Remove</string>
     <string name="signing_owner_dialog_cancel">Cancel</string>
-    <string name="signing_owner_dialog_description">Removing the owner key only removes it from this app. It doesn’t delete any Safes from this app or from blockchain. Safes controlled by this key will no longer be able to sign transactions in this app.</string>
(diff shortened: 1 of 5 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -187,6 +187,9 @@
     <string name="tx_confirmations_execute_waiting">Execute (%1$d more confirmation needed)</string>
     <string name="tx_confirmations_executed">Executed</string>
 
+    <string name="signing_owner_dialog_remove">Remove</string>
+    <string name="signing_owner_dialog_cancel">Cancel</string>
+    <string name="signing_owner_dialog_description">Removing the owner key only removes it from this app. It doesn’t delete any Safes from this app or from blockchain. Safes controlled by this key will no longer be able to sign transactions in this app.</string>
```
biafra23, line 192: I find the last sentence unclear. "Safes [...] will no longer be able to sign transactions [...]" Since when do safes sign transactions. I assume this is the specified wording?

@tschubotz

biafra23 (approved):
👍

app/src/main/res/values/strings.xml
tschubotz: yea, seem like that's the current wording from the designs (https://zpl.io/VYe4pXM). Agree, it's not 100% technically correct.

What would you write instead?

How about using the following last sentence instead? 

`For Safes controlled by this owner key, you will no longer be able to sign transactions in this app.`

cc @posthnikova

posthnikova: @tschubotz @biafra23 Agree.

Issue #954, opened 2020-09-29
Display confirm Button


elgatovital:
<img src="https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/3f2849b6-9986-47df-84d1-e3b2ca3cd5cd" width=300/>

Confirm button should be displayed, if the transaction is in “Awaiting your confirmation” state (the imported key is among safe owners, but not among confirmations)

The confirm button should be stick to the bottom tab bar.

### Selecting “Confirm” should
- Display a confirmation request
![Screenshot 2020-09-28 at 12.09.58.png](https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/8a418e2e-d8f6-4963-95aa-98b5843331b9)
- When user confirmed the pop-up, app makes networking request to the backend to submit the confirmation (see below)
- The activity indicator should be displayed:
  - the confirmation button is blocked, and pull to refresh activity indicator is present.
- When successful, the transaction details data updates the screen, so that "Confirm" button is hidden and a notification should be displayed 
![Screenshot 2020-09-28 at 12.12.26.png](https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/49a9aaf8-77d7-4632-a100-62c11bc36311)

elgatovital:
Blocked by https://github.com/gnosis/safe-client-gateway/issues/162

liliya-soroka:
Verified
app version 2.4.0 - 1329rc

PR #994, opened 2020-10-14
Added tx confirm button to the UI

jpalvarezl:
Handles #954 

- Added button for confirm tx to the UI

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
@@ -244,7 +244,7 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
                 binding.txConfirmations.visible(true)
                 binding.txConfirmButton.visible(true)
                 binding.txConfirmButton.setOnClickListener {
-                    showRemoveDialog(
+                    showConfirmDialog(
                         requireContext(),
                         message = R.string.confirm_transaction_dialog_message,
                         confirm = R.string.confirm,
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
@@ -244,7 +244,7 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
                 binding.txConfirmations.visible(true)
 
                 if (viewModel.isAwaitingOwnerConfirmation(executionInfo, txDetails.txStatus)) {
-                    binding.txConfirmButton.visible(true)
+                    binding.txConfirmButtonContainer.visible(true)
                     binding.txConfirmButton.setOnClickListener {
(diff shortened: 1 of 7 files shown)


app/src/main/java/io/gnosis/safe/utils/AppUtils.kt
```diff
@@ -88,6 +89,8 @@ fun Throwable.getErrorResForException(): Int =
 fun showRemoveDialog(
```
elgatovital, line 89: it should be renamed to showConfirmDialog

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
```diff
@@ -240,6 +242,17 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
         when (val executionInfo = txDetails.detailedExecutionInfo) {
             is DetailedExecutionInfo.MultisigExecutionDetails -> {
                 binding.txConfirmations.visible(true)
+                binding.txConfirmButton.visible(true)
```
elgatovital, line 245: why is the confirm button always visible?

app/src/main/res/layout/fragment_transaction_details.xml
```diff
@@ -151,6 +153,19 @@
                     android:drawablePadding="6dp"
                     android:gravity="start|center_vertical"
                     android:text="@string/tx_details_view_on_etherscan" />
+
+                <TextView
```
elgatovital, line 157: The confirm button should be stick to the bottom tab bar and not scroll (see ticket description)

jpalvarezl: there is no bottom tab bar in the transaction details.

jpalvarezl: Unless we have to change the navigation, but what I remember from planning is that you mentioned you also believed the mocks were wrong.

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
jpalvarezl: How do I know which owner address is stored in the device?

elgatovital: see `OwnerCredentialsRepository`. If hasCredentials() returns true we can retrieve key and address with retrieveCredentials. For example check AppSettingsFragment and its view model.

app/src/main/res/layout/fragment_transaction_details.xml
elgatovital: Here it could mean that confirm button is not scrollable and is always at the bottom. But please clarify it with UX

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
jpalvarezl: Thanks!

elgatovital (approved):

Issue #957, opened 2020-09-29
Safe TxHash validation


elgatovital:
### Additional information
- https://github.com/gnosis/safe-contracts/blob/94f9b9083790495f67b661bfa93b06dcba2d3949/contracts/GnosisSafe.sol#L403
- https://docs.gnosis.io/safe/docs/contracts_signatures/ (ECDSA signature)
- https://docs.gnosis.io/safe/docs/contracts_tx_execution/#transaction-hash (Transaction hash)

liliya-soroka:
Verified
app version 2.4.0-debug with turned off pinning

PR #995, opened 2020-10-15
Add safe tx hash validation

elgatovital:
Handles #957

- Calculate safe tx hash
- Compare calculated hash with hash returned by backend

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
@@ -39,11 +39,11 @@ class TransactionDetailsViewModel
             executionInfo.signers.contains(credentials.address) && !executionInfo.confirmations.map { it.signer }.contains(credentials.address)
         }
 
-    suspend fun validateSafeTxHash(transaction: TransactionDetails, executionDetails: DetailedExecutionInfo.MultisigExecutionDetails): Boolean {
+    suspend fun validateSafeTxHash(transaction: TransactionDetails, executionInfo: DetailedExecutionInfo.MultisigExecutionDetails): Boolean {
         return kotlin.runCatching {
             val safe = safeRepository.getActiveSafe()
-            val safeTxHash = executionDetails.safeTxHash
-            val calculatedSafeTxHash = calculateSafeTxHash(safe!!.address, transaction)?.toHexString()?.addHexPrefix()
+            val safeTxHash = executionInfo.safeTxHash
+            val calculatedSafeTxHash = calculateSafeTxHash(safe!!.address, transaction, executionInfo)?.toHexString()?.addHexPrefix()
             safeTxHash == calculatedSafeTxHash
         }.onSuccess {
             return it
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
@@ -45,11 +45,7 @@ class TransactionDetailsViewModel
             val safeTxHash = executionInfo.safeTxHash
             val calculatedSafeTxHash = calculateSafeTxHash(safe!!.address, transaction, executionInfo)?.toHexString()?.addHexPrefix()
             safeTxHash == calculatedSafeTxHash
-        }.onSuccess {
-            return it
-        }.onFailure {
-            return false
-        }.getOrDefault(false)
+       }.getOrDefault(false)
     }
 }
 
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
@@ -9,6 +9,7 @@ import io.gnosis.data.utils.calculateSafeTxHash
 import io.gnosis.safe.ui.base.AppDispatchers
 import io.gnosis.safe.ui.base.BaseStateViewModel
 import io.gnosis.safe.utils.OwnerCredentialsRepository
+import pm.gnosis.utils.addHexPrefix
 import pm.gnosis.utils.toHexString
 import javax.inject.Inject
 
@@ -42,7 +43,7 @@ class TransactionDetailsViewModel
         return kotlin.runCatching {
             val safe = safeRepository.getActiveSafe()
             val safeTxHash = (transaction.detailedExecutionInfo as DetailedExecutionInfo.MultisigExecutionDetails).safeTxHash
-            val calculatedSafeTxHash = calculateSafeTxHash(safe!!.address, transaction)?.toHexString()
+            val calculatedSafeTxHash = calculateSafeTxHash(safe!!.address, transaction)?.toHexString()?.addHexPrefix()
             safeTxHash == calculatedSafeTxHash
         }.onSuccess {
             return it
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
@@ -39,10 +39,10 @@ class TransactionDetailsViewModel
             executionInfo.signers.contains(credentials.address) && !executionInfo.confirmations.map { it.signer }.contains(credentials.address)
         }
 
-    suspend fun validateSafeTxHash(transaction: TransactionDetails): Boolean {
+    suspend fun validateSafeTxHash(transaction: TransactionDetails, executionDetails: DetailedExecutionInfo.MultisigExecutionDetails): Boolean {
         return kotlin.runCatching {
             val safe = safeRepository.getActiveSafe()
-            val safeTxHash = (transaction.detailedExecutionInfo as DetailedExecutionInfo.MultisigExecutionDetails).safeTxHash
+            val safeTxHash = executionDetails.safeTxHash
             val calculatedSafeTxHash = calculateSafeTxHash(safe!!.address, transaction)?.toHexString()?.addHexPrefix()
             safeTxHash == calculatedSafeTxHash
         }.onSuccess {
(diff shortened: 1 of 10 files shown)


data/src/main/java/io/gnosis/data/utils/SafeTxHash.kt
```diff
@@ -0,0 +1,108 @@
...
+                }
+            }
+        }
+        is TransactionInfo.Custom -> {
+            txInfo.to
+        }
+        is TransactionInfo.SettingsChange -> {
+            safeAddress
+        }
+        is TransactionInfo.Creation -> {
```
jpalvarezl, line 40: I think this branch doesn't make sense, I would let it go to `else` in this case.

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
```diff
@@ -34,6 +37,20 @@ class TransactionDetailsViewModel
                 true == ownerCredentialsRepository.retrieveCredentials()?.let { credentials ->
             executionInfo.signers.contains(credentials.address) && !executionInfo.confirmations.map { it.signer }.contains(credentials.address)
         }
+
+    suspend fun validateSafeTxHash(transaction: TransactionDetails): Boolean {
+        kotlin.runCatching {
+            val safe = safeRepository.getActiveSafe()
+            val safeTxHash = (transaction.detailedExecutionInfo as DetailedExecutionInfo.MultisigExecutionDetails).safeTxHash
```
jpalvarezl, line 44: if The only thing you need from the `MultisigExecutionDetails` is the `safeTxHash`, I would say you modify this method to require the minimum information necessary for that. Meaning, the `safeTxHash: String` directly. (or whichever type that is).

That way you can get rid of the hard cast.

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
```diff
@@ -34,6 +37,20 @@ class TransactionDetailsViewModel
...
+            val calculatedSafeTxHash = calculateSafeTxHash(safe!!.address, transaction)?.toHexString()
+            safeTxHash == calculatedSafeTxHash
+        }.onSuccess {
+            return it
+        }.onFailure {
+            return false
+        }
+        return false
```
jpalvarezl, line 52: instead of all these callbacks, I would suggest using `return Result{...}.getOrDefault(true)`

data/src/main/java/io/gnosis/data/utils/SafeTxHash.kt
```diff
@@ -0,0 +1,108 @@
...
+    val data = Sha3Utils.keccak(transaction.txData?.hexData?.hexToByteArray() ?: ByteArray(0)).toHex().padStart(64, '0')
+    val operationString = (transaction.txData?.operation?.id?.toBigInteger() ?: BigInteger.ZERO).paddedHexString()
+
+    val executionInfo = when (val executionInfo = transaction.detailedExecutionInfo) {
+        is DetailedExecutionInfo.MultisigExecutionDetails -> {
+            executionInfo
+        }
+        else -> {
+            throw UnsupportedTransactionType(transaction::javaClass.name)
+        }
```
jpalvarezl, line 58: This is just a hard cast with extra steps 😄  I would require the `MultisigExecutionDetails` as an input to the method, so you don't need to hard cast everywhere.

data/src/main/java/io/gnosis/data/utils/SafeTxHash.kt
```diff
@@ -0,0 +1,108 @@
...
+import pm.gnosis.crypto.utils.Sha3Utils
+import pm.gnosis.model.Solidity
+import pm.gnosis.utils.hexToByteArray
+import pm.gnosis.utils.toHex
+import java.math.BigInteger
+
+private const val ERC191_BYTE = "19"
+private const val ERC191_VERSION = "01"
+
+fun calculateSafeTxHash(
```
jpalvarezl, line 16: I would love to see unit tests for this. There are some hardcoded values too, that I am guessing come from the legacy app and are just what they are. Could we port the unit tests from the legacy app over to this one?

jpalvarezl (changes requested):

data/src/main/java/io/gnosis/data/utils/SafeTxHash.kt
elgatovital: ok

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
elgatovital: we'll have to do hard cast at the upper level. I think having it encapsulated in validateSafeTxHash is more elegant than doing it at the viewModel level

elgatovital: besides executionInfo is already part of the transaction object

jpalvarezl: The only subtype of `DetailedExecutionInfo` that has a `safeTxHash` is the  `MultisigExecutionDetails`, so I would say it would be far more elegant to leverage the strong typing of kotlin to your advantage. 

I don't think that the cast is necessary at all, as this callback is probably triggered from a context the type is already known ? I am imagining that it comes from the big `when` block for `executionInfo`

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
```diff
@@ -34,6 +38,19 @@ class TransactionDetailsViewModel
...
+        return kotlin.runCatching {
+            val safe = safeRepository.getActiveSafe()
+            val safeTxHash = (transaction.detailedExecutionInfo as DetailedExecutionInfo.MultisigExecutionDetails).safeTxHash
+            val calculatedSafeTxHash = calculateSafeTxHash(safe!!.address, transaction)?.toHexString()?.addHexPrefix()
+            safeTxHash == calculatedSafeTxHash
+        }.onSuccess {
+            return it
+        }.onFailure {
+            return false
+        }.getOrDefault(false)
```
jpalvarezl, line 52: If you use `getOrDefault` you don't need to have `onSuccess` and `onFailure` at all. Just return what `getOrDefault` returns.

jpalvarezl (changes requested):

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
jpalvarezl:
```suggestion
return kotlin.runCatching {
            val safe = safeRepository.getActiveSafe()
            val safeTxHash = (transaction.detailedExecutionInfo as DetailedExecutionInfo.MultisigExecutionDetails).safeTxHash
            val calculatedSafeTxHash = calculateSafeTxHash(safe!!.address, transaction)?.toHexString()?.addHexPrefix()
            safeTxHash == calculatedSafeTxHash
        }.getOrDefault(false)
```

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
```diff
@@ -34,6 +38,19 @@ class TransactionDetailsViewModel
                 true == ownerCredentialsRepository.retrieveCredentials()?.let { credentials ->
             executionInfo.signers.contains(credentials.address) && !executionInfo.confirmations.map { it.signer }.contains(credentials.address)
         }
+
+    suspend fun validateSafeTxHash(transaction: TransactionDetails): Boolean {
+        return kotlin.runCatching {
+            val safe = safeRepository.getActiveSafe()
+            val safeTxHash = (transaction.detailedExecutionInfo as DetailedExecutionInfo.MultisigExecutionDetails).safeTxHash
+            val calculatedSafeTxHash = calculateSafeTxHash(safe!!.address, transaction)?.toHexString()?.addHexPrefix()
```
jpalvarezl, line 46: Maybe it is worth having a method that encapsulates all these as `calculateSafeTxHash().asString()` ?

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
elgatovital: you proposing passing both, transaction and executionInfo as parameters, right?

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
elgatovital: curently it is used only in one place (tests not counting). I would wait and if we use it more than in one place - then this method could be added

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
jpalvarezl: Yeah, just passing both. Or even you could create a specific structure for this, if you prefer. because we are not taking advantage of typing here. If we specificially use `MultisigExecutionDetails` it is better documented that this method only works for these types. If you feel this is too knit-picky we can drop it, but I think that we should use types when we have them for the sake of clarity.

data/src/main/java/io/gnosis/data/utils/SafeTxHash.kt
jpalvarezl: Why not use a hard cast here? It would have the same effect and `ClassCastException` is sufficiently informative?

elgatovital: it's been updated. Please recheck

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
elgatovital: Please recheck the latest state

jpalvarezl (approved):
Looks good 👍

Issue #958, opened 2020-09-29
Submitting the transaction confirmation


elgatovital:
- The app uses /safes/{address}/transactions/ endpoint of the transaction service to submit  the confirmation
- After getting a successful response, the app will request updated tx details using the gateway service.
  - The screen should be updated with the latest tx info: awaiting execution or awaiting confirmations.
- In case of error, the app hides the loading indicator and displays the actual (latest) state with “Confirm” button, and a snackbar with the error message.
  - In V1 we will always show a “Failed to sign” message for all 422 errors.
  - Error message codes (TBD)
    - Networking error
    - Signing the tx when the key is already not an owner
    - Signing already executed tx
    - What else? (Unexpected error)

- In case of the transaction details request failure, the app displays “Data can not be loaded” placeholder and a standard snackbar with an error message.

### Additional information
- https://github.com/gnosis/safe-contracts/blob/94f9b9083790495f67b661bfa93b06dcba2d3949/contracts/GnosisSafe.sol#L403
- https://docs.gnosis.io/safe/docs/contracts_signatures/ (ECDSA signature)
- https://docs.gnosis.io/safe/docs/contracts_tx_execution/#transaction-hash (Transaction hash)

elgatovital:
Blocked by https://github.com/gnosis/safe-client-gateway/issues/162

jpalvarezl:
https://app.zeplin.io/project/5e9efc30726e106958cc43de/screen/5f7447bf583784456fd188f1

https://app.zeplin.io/project/5e9efc30726e106958cc43de/screen/5f7447c17a3d6a33302fd47a

https://app.zeplin.io/project/5e9efc30726e106958cc43de/screen/5f7447c46973a665771cd69c

liliya-soroka:
Verified with one note:
The next requirement will be covered later
"Error message codes (TBD)
Networking error
Signing the tx when the key is already not an owner
Signing already executed tx
What else? (Unexpected error)"
app version: 2.4.0-1332rc

PR #1005, opened 2020-10-20
Submit transaction confirmation integration with backend

jpalvarezl:
Handles #958 

- Added callbacks to submit a transaction confirmation to the gateway
- Added tests to `TransactionDetailsViewModel` and `TransactionRepository` where possible
- Added `readFromJson` utility method for tests

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
@@ -251,9 +251,7 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
                             message = R.string.confirm_transaction_dialog_message,
                             confirm = R.string.confirm,
                             confirmColor = R.color.safe_green
-                        ) {
-                            //TODO implement network callback
-                        }
+                        ) { viewModel.submitConfirmation(txDetails, executionInfo) }
                     }
                 } else {
                     binding.txConfirmButton.visible(false)
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
@@ -79,11 +79,16 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
                         updateUi(it)
                     }
                 }
+                is ConfirmationSubmitted -> {
+                    binding.txConfirmButtonContainer.visible(false)
+                    viewAction.txDetails?.let(::updateUi)
+                }
                 is Loading -> {
                     showLoading(viewAction.isLoading)
                 }
                 is ShowError -> {
                     binding.refresh.isRefreshing = false
+                    binding.txConfirmButton.isEnabled = true
                     when (viewAction.error) {
                         is Offline -> {
                             snackbar(requireView(), R.string.error_no_internet)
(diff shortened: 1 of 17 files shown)


app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
```diff
@@ -321,6 +328,7 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
 
         binding.content.visible(true)
         binding.contentNoData.root.visible(false)
```
elgatovital, line 330: better hanlde loading visibility in viewModel.state.observe (UpdateDetails viewAction in this case) and let updateUi deal only with content elements of the page

jpalvarezl: If you are loading content into the page, shouldn't the loading spinner always be hidden when the content is received? Besides, remember we had issues with `liveData` swallowing some state changes if you emit too frequently? I think this is a bit safer and makes sense in terms of how we use the app components. As in,

if loading -> show loading
if error -> hide content, hide loading, show error
if content -> hide loading, hide error, show content

elgatovital: we only call showContent in UpdateDetails

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
```diff
@@ -79,11 +79,17 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
...
+                is ConfirmationSubmitted -> {
+                    binding.txConfirmButtonContainer.visible(false)
+                    viewAction.txDetails?.let(::updateUi)
+                    snackbar(requireView(), R.string.confirmation_successfully_submitted)
+                }
                 is Loading -> {
                     showLoading(viewAction.isLoading)
                 }
                 is ShowError -> {
                     binding.refresh.isRefreshing = false
```
elgatovital, line 91: this should be replaced with showLoading(false)

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
elgatovital: yes, what you said is right, but having it separated and being called explicitly makes code more readable I think

elgatovital: even in your description you have separated those steps (hide loading, show content)

jpalvarezl: yeah but what I am talking about is the liveData that swallows events. That's what I think is a bigger problem.

elgatovital: to be clear: I am asking to sepparately call showLoading(false) in UpdateDetails and remove code that deails with spinners from updateUi function.

elgatovital (approved):

app/src/main/res/xml/network_security_config.xml
```diff
@@ -2,6 +2,7 @@
 <network-security-config>
     <domain-config cleartextTrafficPermitted="true">
         <domain includeSubdomains="true">127.0.0.1</domain>
+        <domain includeSubdomains="true">10.0.2.2</domain>
```
elgatovital, line 5: this should be removed

elgatovital (changes requested):
please update network_securit_config

jpalvarezl:
Just for reference: https://developer.android.com/studio/run/emulator-networking#networkaddresses

elgatovital (approved):

Issue #1010, opened 2020-10-21
[Android] "Awaiting your confirmation" state is not used on the tx details page 


liliya-soroka:
**Describe the bug**
When the tx is required the imported owner key confirmation the state "Awaiting you confirmation" appears on the tx list but "Awaiting confirmation" is still used on the tx details

**To Reproduce**
1. Import owner key
2. Create tx that will require a confirmation from the owner with the imported key
3. Go to the tx list and open tx in the "Awaiting your confirmation" state

**Expected behavior**
"Awaiting your confirmation" state should be used on the tx details page too together with confirm button

**Smartphone (please complete the following information):**
 - Device: Pixel 4
 - Android Version: 11
 - App Version 2.4.0 -1326

**Environment**
staging, rinkeby
**Additional context**

liliya-soroka:
Verified
app version 2.4.0 -1329rc

PR #1012, opened 2020-10-22
Show awaiting your confirmation on tx details page

elgatovital:
Handles #1010

@gnosis/mobile-devs


jpalvarezl (approved):
I would love to have this logic pushed a bit further down to avoid duplication but for now I think it's ok.

Issue #996, opened 2020-10-15
[Android] Invalid safe error is displayed when the valid safe address is loaded for a long time


liliya-soroka:
**Describe the bug**
Seems that instead of timeout error from back-end the "Invalid address" error is displayed under the safe address field on load safe page.
It's reproducable on the prod mainnet as there is an issue with loading safes, addresses, lists - sometimes it takes a lot of time and timeout error is displayed

**To Reproduce**
1. Add safe using QR scanning ( I have used QR code of the safe on web)
Current result: loading icon is displayed for a very long time on the add safe address page

**Expected behavior**
If the back-end error is the reason of failure during address adding the snack bar with the error should be displayed instead of wrong red error under the field

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/e0fbfc0d-199c-4348-8215-0c54de1ee3e5) - the error appears for the safe 0xB260dD872c57205C9f308ddbd1C3B038c5c69a5F 

**Smartphone (please complete the following information):**
 - Device: Pixel 4
 - Android Version: 11
 - App Version 2.3.0 - 1282rc

**Environment**
Prod mainnet
**Additional context**

tschubotz:
I marked this as `major` since this is about the "add Safe" flow. Once we know the reason for the slow loading time, we can change the bug prio also.

liliya-soroka:
Verified
app version 2.4.0 -1374

PR #1016, opened 2020-10-23
Fix error handling when adding safe

elgatovital:
Handles #996

- Prevent errors from being swallowed in the view model
- Show correct error in snack bar

@gnosis/mobile-devs


biafra23 (approved):
👍

Issue #1020, opened 2020-10-26
Tracking of num_keys_imported user property


DmitryBespalov:
num_keys_imported is an Integer user property that tracks the current number of imported keys.

For v1 of signing feature, it will be 0 or 1. Number is updated on adding/removing key.

- When key is added or removed, the value will change.

liliya-soroka:
one issue to check - https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-android/1045

PR #1023, opened 2020-10-27
Track num_keys_imported

biafra23:
Handles #1020

- Add user property
- set user property on storage and removal of owner 

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/settings/app/AppSettingsViewModel.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/settings/app/AppSettingsViewModel.kt
@@ -2,12 +2,14 @@ package io.gnosis.safe.ui.settings.app
 
 import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewModel
+import io.gnosis.safe.Tracker
 import io.gnosis.safe.utils.OwnerCredentialsRepository
 import pm.gnosis.model.Solidity
 import javax.inject.Inject
 
 class AppSettingsViewModel @Inject constructor(
-    private val ownerCredentialsRepository: OwnerCredentialsRepository
+    private val ownerCredentialsRepository: OwnerCredentialsRepository,
+    private val tracker: Tracker
 ) : ViewModel() {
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/io/gnosis/safe/ui/settings/app/AppSettingsViewModel.kt
```diff
@@ -23,6 +25,7 @@ class AppSettingsViewModel @Inject constructor(
 
     fun removeSigningOwner() {
         ownerCredentialsRepository.removeCredentials()
+        tracker.setNumKeysImported(0)
```
biafra23, line 28: Should I add a test that verifies that `tracker.setNumKeysImported(0)` was called?

app/src/main/java/io/gnosis/safe/ui/settings/owner/list/OwnerSelectionViewModel.kt
```diff
@@ -61,6 +63,7 @@ class OwnerSelectionViewModel
             val key = derivator.keyForIndex(ownerIndex)
             val addresses = derivator.addressesForPage(ownerIndex, 1)
             ownerCredentialsVault.storeCredentials(OwnerCredentials(address = addresses[0], key = key))
+            tracker.setNumKeysImported(1)
```
biafra23, line 66: Should I add a test that verifies that `tracker.setNumKeysImported(1)` was called?

jpalvarezl: If there is a test in place, adding a `verify` method call is sufficient IMO.

app/src/main/java/io/gnosis/safe/ui/settings/app/AppSettingsViewModel.kt
jpalvarezl: If there is already a test setup just add a `verify` call otherwise, I would say that you don't need to bother.

jpalvarezl (approved):
LGTM!

Issue #936, opened 2020-09-23
Tx creation page layout adjustemnts (LS: full regression)


elgatovital:
MasterCopyItem has its own separator that is merged with separator on creation tx page creating a thick line: ![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/55c76df4-c231-48ee-9478-a1e6deb77415)

add xml attributes to specify visibility of the innate separator (e.g. like in AddressItem)

fragment_transaction_details_creation.xml layout readability improvement:
Use ViewSwitcher to group implementation_item & no_implementation_item. Same for factory_item. This will allow switching views by simply calling switcher.showNext(). Also, grouping elements in ViewSwitcher will make xml more readable as it will be more clear that elements belong together. 

- Device:
- Android OS version:
- Safe app version:

**Notes for QA**: This change might break the layout of tx list, tx details,  app settings, collectible details and safe settings.

liliya-soroka:
@elgatovital , seem that the same issue on the tx details page for ERC721 - the line under Asset Contract is too thick. Could it be also fixed in the issue or it's better to create a new one?

biafra23:
@liliya-soroka I tried to find and fix this double line bug everywhere. That means you probably have to check the whole app for errors. Sorry!

liliya-soroka:
The issue should be carefully checked during the full regression for the release candidate 2.5.0

liliya-soroka:
Verified
app version 2.5.0-1421rc

PR #1029, opened 2020-10-28
Creation layout

biafra23:
Handles #936 

**I think we need to make a decision on whether the last separator should be part of a custom view. And then make this consistent throughout the app.**

- Remove additional separator by extracting it from the mastercopy custom view
- Changing color of item separator in tx list (according to the zeplin layouts)
- Fixing height of several separators
- Fixing bottom margin for tx hash
- Fixing bottom margin for created date
- Fixing separator color and height for detail views
- Added ids to several separators to make layout debugging easier
- Added custom attribut to NamedAddressItem to disable/enable separator

**Notes for QA**: This might break the layout of tx list, tx details,  app settings, collectible details and safe settings. 

Design: https://zpl.io/amAw7dK

Before | After 
------- | -------
Screenshot_20201028-134751 | Screenshot_20201028-134541

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/res/layout/fragment_settings_safe.xml
+++ b/app/src/main/res/layout/fragment_settings_safe.xml
@@ -109,7 +109,7 @@
                 <View
                     android:id="@+id/master_copy_separator"
                     android:layout_width="match_parent"
-                    android:layout_height="2dp"
+                    android:layout_height="@dimen/item_separator_height"
                     android:background="@color/item_separator"
                     app:layout_constraintBottom_toBottomOf="parent" />
 
--- a/app/src/main/res/layout/fragment_settings_safe.xml
+++ b/app/src/main/res/layout/fragment_settings_safe.xml
@@ -106,6 +106,13 @@
                     android:background="@drawable/background_selectable_white"
                     app:setting_openable="false" />
 
+                <View
+                    android:id="@+id/master_copy_separator"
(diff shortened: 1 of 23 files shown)


app/src/main/res/layout/fragment_settings_safe.xml
```diff
@@ -106,6 +106,13 @@
                     android:background="@drawable/background_selectable_white"
                     app:setting_openable="false" />
 
+                <View
+                    android:id="@+id/master_copy_separator"
+                    android:layout_width="match_parent"
+                    android:layout_height="2dp"
```
jpalvarezl, line 112: Is this double size? Should we use `@dimen/item_separator_height`?

jpalvarezl (approved):
The after screenshot looks great :D Really happy to see stuff parametrised!

app/src/main/res/layout/fragment_settings_safe.xml
biafra23: I missed one. This should also be `@dimen/item_separator_height` which is 2dp

biafra23: Fixed in dfe45c2bdd12532d7f1091d8eb32a29fde1ef724

elgatovital (approved):

Issue #1022, opened 2020-10-26
Transaction list reloading logic when going back from transaction details


DmitryBespalov:
- Always reload transaction list if transaction details were opened from the transaction in the 1st "screen" - visible portion of the list;
- Do not reload the screen when transactions were scrolled (even a little bit) and then transaction details opened and the user got back.

biafra23:
This might be helpful: https://stackoverflow.com/a/38248034

liliya-soroka:
Reopened:
Always reload transaction list if transaction details were opened from the transaction in the 1st "screen" - visible portion of the list;
New created transaction together with Queue section doesn't appear if the tx from the first page was opened
Steps:
1. Open tx list - first page with History section only 
2. Open tx details for the tx on the first page
3. create a new transaction that will be in awaiting confirmation state
4. go back from the tx details to the tx list
Current result: No new created tx with Queue section 

Env:
staging, rinkeby

Version 
2.5.0 -1421rc

elgatovital:
From the epic:
> Always reload if transaction details were opened from the transaction in the 1st "screen" - visible portion of transaction;
> Do not reload the screen when transactions were scrolled and then transaction details opened and user got back.

This means tx list won't be reloaded if you scrolled the list down. If the first tx element is visible and you open some tx item to see details, tx list will be reloaded when you come back.

liliya-soroka:
Verified 
app version 2.5.0 -1421rc
one additional issue for the edge case is created

PR #1031, opened 2020-10-29
Adjust reloading logic for tx list

elgatovital:
Handles #1022

- Reload tx list only if scrolled past first tx item

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListFragment.kt
@@ -13,6 +13,7 @@ import androidx.paging.LoadState
 import androidx.paging.PagingData
 import androidx.recyclerview.widget.DividerItemDecoration
 import androidx.recyclerview.widget.LinearLayoutManager
+import androidx.recyclerview.widget.RecyclerView
 import io.gnosis.data.models.Safe
 import io.gnosis.safe.R
 import io.gnosis.safe.ScreenId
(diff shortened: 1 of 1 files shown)


elgatovital:
preventing reload if the list was scrolled `just a bit` is impractical as such condition interferes with swipe to refresh. It would be very cumbersome to bring the list into the initial unscrolled position without triggering it.

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListFragment.kt
```diff
@@ -73,6 +76,13 @@ class TransactionListFragment : SafeOverviewBaseFragment<FragmentTransactionList
             }
         }
 
+        binding.transactions.addOnScrollListener(object : RecyclerView.OnScrollListener() {
+            override fun onScrolled(recyclerView: RecyclerView, dx: Int, dy: Int) {
+                val firstVisibleItem = (binding.transactions.layoutManager as LinearLayoutManager).findFirstVisibleItemPosition()
+                reload = firstVisibleItem <= 1
```
biafra23, line 82: Just curious. Can this be negative? What would that mean if it is negative?

biafra23 (approved):
This solution is so simple! 👍

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListFragment.kt
elgatovital: it would mean recycler view has no children and in this case, scrolling event won't be triggered at all

Issue #1027, opened 2020-10-28
[Android] Awaiting your confirmation is not seen fully on some devices on the tx details


liliya-soroka:
**Describe the bug**
As the type of the tx + status is displayed in one line on the tx details the Awaiting for your confirmation status is not seen on some devices

**To Reproduce**
1. I used Huawei P30
2. Go to the tx list and click on the tx in the Awaiting your confirmation state
3. Open tx details

**Expected behavior**
Display type and status in two lines if the state is too long 

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/ea21f88c-46a8-4530-99b6-78d1d867f784)

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 11
 - App Version 2.4.0 -1332rc

**Environment**
prod, mainnet
**Additional context**

liliya-soroka:
Verified
app version 2.4.0 -1374

PR #1033, opened 2020-10-29
Fix tx status layout

elgatovital:
Handles #1027

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -155,6 +155,7 @@
     <string name="tx_status_type_creation">Safe created</string>
 
     <string name="tx_status_status_awaiting_confirmations">\u2022 Awaiting confirmations</string>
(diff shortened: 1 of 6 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -155,6 +155,7 @@
     <string name="tx_status_type_creation">Safe created</string>
 
     <string name="tx_status_status_awaiting_confirmations">\u2022 Awaiting confirmations</string>
+    <string name="tx_status_status_awaiting_your_confirmation">\u2022 Awaiting your confirmation</string>
```
biafra23, line 158: Do we need the duplication of these Strings? I probably introduced it but I don't remember whether I had a good reason :-)

biafra23 (approved):
LGTM except for the string duplication question

app/src/main/res/values/strings.xml
elgatovital: no, we don't.  I'll clean it up by leaving only status labels

Issue #840, opened 2020-08-27
Error handling: incorrect error under ENS name field when no internet connection or chars was entered


liliya-soroka:
**Describe the bug**
There is not userfriendly and not easily readable error under ENs name field ( loading safe flow) in case no internet connection 
**To Reproduce**
1st situation:
1. Turn off internet connection
2. Go to Load safe flow
3.Enter valid ens name
Current result: ![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/a328dbe4-5d04-44fb-95fa-f1a603d973d2)

2nd situation"
1. Go to Load safe flow
3.Open enter ENS name
4. Turn off internet connection
5. Enter valid ens name
6. See error - the error from the attachment
Current result: ![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/0228c1eb-27fd-4d94-88a1-5a39d9a610fa)
3rd situation
1. Internet connection is on all the time
2. Go to the enter ENS
3. Put space before the ens name or any char
Current result: The error "The input doesn't conform to the STD 3 ASCII rules
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/5775fe3c-9b31-4eb3-b404-aa98598aeef1)

**Expected behavior**
No internet connection error should be displayed under ens name field

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/0a6e43b5-cee7-4dc8-afaa-8a18996aa7ff)

**Smartphone (please complete the following information):**
 - Device: Pixel 4
 - Android Version: 11
 - App Version 2.4.0 -1332rc
**Additional context**

liliya-soroka:
Reopened:
Steps to reproduce the behavior:
1.  Go to Load safe flow
3.Open enter ENS name
4. Turn off internet connection
5. Enter valid ens name
6. See error - the error from the attachment

app version 1.12.1 -110

biafra23:
@tschubotz 

Regarding 3rd case "Illegal characters in name". Since this can be validated locally IMHO the error should not be "No internet connection" but something to tell the user they are using illegal characters. I agree that the current error (The inut does not conform to the STD 3 ASCII rules) is not very friendly or easy to understand. Most users probably don't even know what ASCII means.

Who can give me a better error message if characters are used that cannot be in an ENS name?

liliya-soroka:
Verified
app version 2.4.0 - 1389

PR #1034, opened 2020-10-29
ENS invalid characters

biafra23:
Handles #840

- Replace IllegalArgumentException with "Name contains invalid characters"
- Add test

@gnosis/mobile-devs


elgatovital:
if I enter something and then clear the field the error 'no resolver set for the record' is displayed. We probably should not display errors if the input field is empty. It's basically back to the starting point.

elgatovital (approved):

Issue #1019, opened 2020-10-26
[Android]No any error on tx list in case of any error from the client-gateway( version without pinning is required)


liliya-soroka:
**Describe the bug**
Using the mitmproxy the response with code 422 and 500 were simulated from the client-gateway on the tx list. The tx list either is not loaded at all without any error if it was not opened or the page contains old data without any error on UI on the pull to refresh

**To Reproduce**
1st situation:
1. Turn on proxy 
2. send error as a response from the client-gateway
3. go to the tx list

2nd situation:
1. Turn on proxy 
2. go to the tx list
3. try to open tx details and send error as a response from the client-gateway
4. go back to the tx list and refresh the tx list
Current result: nothing happens

**Expected behavior**
The error should be displayed on UI in case of error from the client-gateway for the tx list as for the tx details

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 11
 - App Version : internal build with turned off pinning

**Environment**
staging, rinkeby

**Additional context**

liliya-soroka:
Verified
app version 2.5.0- 1421rc

PR #1039, opened 2020-11-02
Handle errors when loading tx list

elgatovital:
Handles #1019

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListFragment.kt
@@ -59,6 +59,11 @@ class TransactionListFragment : SafeOverviewBaseFragment<FragmentTransactionList
 
                 binding.progress.isVisible = loadState.refresh is LoadState.Loading && adapter.itemCount == 0
                 binding.refresh.isRefreshing = loadState.refresh is LoadState.Loading && adapter.itemCount != 0
+
+                val refresh = loadState.refresh
+                if (refresh is LoadState.Error) {
+                    handleError(refresh.error)
+                }
                 val append = loadState.append
                 if (append is LoadState.Error) {
                     handleError(append.error)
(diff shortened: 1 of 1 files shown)


biafra23 (approved):
LGTM 👍

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListFragment.kt
```diff
@@ -59,6 +59,11 @@ class TransactionListFragment : SafeOverviewBaseFragment<FragmentTransactionList
 
                 binding.progress.isVisible = loadState.refresh is LoadState.Loading && adapter.itemCount == 0
                 binding.refresh.isRefreshing = loadState.refresh is LoadState.Loading && adapter.itemCount != 0
+
+                val refresh = loadState.refresh
+                if (refresh is LoadState.Error) {
+                    handleError(refresh.error)
+                }
```
jpalvarezl, line 66: Actually this is a perfect use case for `let`. I think everyone uses it as a `null` checker but I think it really shines when you need an immutable reference to an object brought into scope. 

Not that there is anything wrong with doing this, the way you did. Just sharing a thought :D

jpalvarezl (approved):

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListFragment.kt
elgatovital: how would you use `let` here? ```loadState.refresh.let { if(it is LoadState.Error) handle(it) } ```?

jpalvarezl: yeah, just like that you would only have to change `handleError(it.error)`

jpalvarezl: it would be perfect to you use `takeIf{ it is LoadState.Error}?.let{...}` but it sucks that smart casts don't work in that scenario :/

Issue #991, opened 2020-10-14
[Android] The add address page is used while the correct address is loaded ( full regression on prod mainnet)


liliya-soroka:
**Describe the bug**
The add address page is used while the correct address is loaded. It courses the confusing for the user situation when the error for the previous wrong address is displayed for the correct safe address while it's loaded.
Previously the issue was not so actual as the data loading was much quicker on prod mainnet. Now we have some delay in data loading ( sometimes 5-15  sec for both clients), so the screen with error is displayed for quite a long time.

**To Reproduce**
1. Go to add safe
2. Past /scan invalid safe address
3. Clic in the address field and select enter ENS
4. Enter valid ens
Current result: The user is returned back to the Add address page with error for previous address and loading icon. Only in some time the page with recognised correct address appears. The user can be confused and can make wrong conclusion that the entered ens was wrong

**Expected behavior**
The page with "Enter your safe.." without error, link on web app  and with entered safe address should be displayed while the address is validated 

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/b7bcd912-88a3-437c-be67-274737267499)

**Smartphone (please complete the following information):**
 - Device: Pixel 4
 - Android Version: 11
 - App Version 2.3.0 - 1278rc

**Environment**
Prod mainnet
**Additional context**

liliya-soroka:
Reopened: Could the same fix be added for other add safe address options?
1. Go to add safe
2. Past /scan invalid safe address
3. Scan QR code or past from the  clipboard ( please try to add safe=0x583788f490a278DB2a8d6D7226264b4985f75678 on prod during testing. To load the safe more time are required than for the more "light" safes)
Current result: The user is returned back to the Add address page with error for previous address and loading icon. Only in some time the page with recognised correct address appears. The user can be confused and can make wrong conclusion that the entered ens was wrong
Screenshot:
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/dc4bd603-0d21-438a-9625-38087d8fd12f)

Tested env:
Prod , mainnet
App version:
2.4.0 -1374

liliya-soroka:
Verified
app version 2.5.0 -1421rc

Issue #990, opened 2020-10-14
[Android] Show wrong pasted safe address when the past from clipboard is used


liliya-soroka:
**Describe the bug**
When the user pasts the wrong address in the safe address field during the load safe the address field is empty and the error appears under the field. The user doesn't have the possibility to see and check  the passed address without visualization it 

**To Reproduce**
1. Open add safe flow
2. Select "Copy from clipboard"
3. Past wrong safe address( owner address, not existed safe address)

**Expected behavior**
Show the pasted data in the safe address

**Smartphone (please complete the following information):**
 - Device: Pixel 4
 - Android Version: 11
 - App Version 2.3.0 - 1278rc

**Environment**
prod mainnet
**Additional context**

liliya-soroka:
Verified
app version 2.4.0 - 1404

PR #1040, opened 2020-11-02
Prevent previous error being shown for new address; keep entered invalid safe address visible for reference

elgatovital:
Handles #991 
Handles #990

@gnosis/mobile-devs


biafra23 (approved):
I cannot reproduce the issue, but the code looks like it fixes the described bug :-/

Issue #608, opened 2020-05-20
Optmize test coverage to cover only intended files


biafra23:
### Expected behavior
According to our OKRs, test coverage is supposed to cover the following:
- Only logic code needs to be covered 100%
- Android: view models, repository, manager, services, ...
- Generated code is not required (should be tested elsewhere normally)

### Actual behavior
- Code coverage covers all code in all enabled modules

rmeissner:
The okr doesn't say that only logic code should be measured. It says that logic code should have 100% test coverage. This doesn't meant that you shouldn't be aware of the test coverage of other parts of the app. 

If you use the test config from the legacy app generated code is already excluded.

biafra23:
I changed the description regarding logic code coverage

rmeissner:
Could you add an example of generated code that is included?

biafra23:
Our code coverage setup is a bit confusing.

At the moment we have to different jacoco setups. One with version 0.7.9 and one with version 0.8.5. The configation for 0.8.5 is taken from `buildsystem/coverage_report.gradle`

`createDebugCoverageReport`  -> 0.7.9
`createDebugTestCoverage` -> 0.8.5

I will try to disable v0.7.9. as it just adds confusion and it is unclear how upgrade the version of configure it.

liliya-soroka:
closed, not qa required

PR #1043, opened 2020-11-04
Test coverage verification

biafra23:
Handles #608

- Add check for certain test coverage ATM: 0% for qrscanner, 30% for app and 70% for data

**What other file patterns should/could we add to `jacocoFileFilter` to increase the required coverage for the rest of the files?**

@gnosis/mobile-devs


buildsystem/coverage_report.gradle
```diff
@@ -42,24 +90,136 @@ if (variants != null) {
...
+    jacocoCoverageThresholdDefault = 0.60
+
+    jacocoIgnoreList = [] // add module names to ignore
+    // Exclude file by names, packages or types. Such files will be ignored during test coverage
+    // calculation
+    jacocoFileFilter = [
+            '**/*App.*',
+            '**/*Application.*',
+            '**/*Activity.*',
+            '**/*Fragment.*',
```
rmeissner, line 171: I am not sure if I agree with this. It is nice that the number then is only based on the coverage, but requirement that only logic code is tested is a compromise in the beginning. In the long run all code written by the team should be tested (including UI code) and it should be visible on codecov for which code this is not the case yet.

buildsystem/coverage_report.gradle
```diff
@@ -13,15 +13,63 @@ jacoco {
...
+        reports {
+            csv.enabled false
+            xml.enabled true
+            html {
+                enabled true
+                destination file("${buildDir}/coverage-report")
+            }
+        }
+        // Change as needed
         def fileFilter = [
```
jpalvarezl, line 73: Maybe it is also worth including `**/*Binding*.*` ?

buildsystem/coverage_report.gradle
biafra23: We have two lists. One for doing coverage and one for failing the build if a threshold s not met. I will change the names of the filters to make this more clear. I think it is ok to exclude Fragments and Activities for the threshold calculation.

rmeissner: Ahh yes, for the threshold go for it, totally makes sense.

buildsystem/coverage_report.gradle
biafra23: I'd exclude it for the threshold check, but not for the coverage list.

biafra23: Maybe we should exclude all files in `intermediate` and `generated`?

ci/run_ui_tests.sh
```diff
@@ -28,6 +28,6 @@ adb shell settings put global animator_duration_scale 0
 adb shell am start -n io.gnosis.safe.debug/io.gnosis.safe.ui.splash.SplashActivity
 
 echo "Execute UI tests"
-./gradlew createDebugTestCoverage --stacktrace
+./gradlew jacocoTestCoverageVerification --stacktrace
 
 bash <(curl -s https://codecov.io/bash) -f '*TestCoverage.xml'
```
biafra23, line 33: Why was this only done in the ui tests files? It seems jacocoTestReport has to run instrumentation tests since createDebugCoverageReport seems to depend on them. Should I try to make this run without instrumentation, or are we fine with running instrumntation on the ci. Apparently it worked in the past.

biafra23:
ATM the codecov plugin can't find the reports:

```
No coverage report found.
```

buildsystem/coverage_report.gradle
biafra23: Unfortunately, we cannot exclude by directory. The files are taken from the compiled classes directory ($buildDir/tmp/kotlin-classes/debug). We can only exclude based on package or file name patterns :-(

buildsystem/coverage_report.gradle
```diff
@@ -42,24 +90,137 @@ if (variants != null) {
...
-                    "jacoco/test${variantName}UnitTest.exec",
-                    "outputs/code-coverage/connected/*coverage.ec"
-            ])
+private static boolean isAndroidModule(Project project) {
+    def isAndroidLibrary = project.plugins.hasPlugin('com.android.library')
+    def isAndroidApp = project.plugins.hasPlugin('com.android.application')
+    return isAndroidLibrary || isAndroidApp
+}
+
+// Modules without any unit tests should also fail validation
```
elgatovital, line 187: would it fail build if validation fails? We don't have tests for :contracts module (an android module)

elgatovital:
shouldn't we see coverage reports for this pr? Currently, only buidkite check is visible

buildsystem/coverage_report.gradle
biafra23: If its not included in the modules build.gradle, then it doesn't fail.

biafra23:
> shouldn't we see coverage reports for this pr? Currently, only buildkite check is visible

No additional reports are generated for the PR. But we keep the buildkite checks.

jpalvarezl (approved):

Issue #1035, opened 2020-10-29
[Android] Balances total list item


jpalvarezl:
The client gateway will serve this structure under the `/v1/safes/{safe_address}/balances/{fiat}?{trusted}&{exclude_spam}`

```rust 
pub enum Balances {
    fiat_total: String, // total balance in the requested fiat
    Items: Vec<Balance>
}

pub struct Balance {
    pub token_info: TokenInfo,
    pub balance: String,
    pub fiat_balance: String,
    pub fiat_conversion: String,
}

pub struct TokenInfo {
    pub token_type: TokenType,
    pub address: String,
    pub decimals: u64,
    pub symbol: String,
    pub name: String,
    pub logo_uri: Option<String>,
}
```

`trusted` and `exclude_spam` have default values for the apps so they don't need to be changed. 

The UI should be like `Option 2` specified here: https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe/360undefined

liliya-soroka:
Verified
app version 2.5.0-1421rc

PR #1044, opened 2020-11-04
Show total balance

elgatovital:
Handles #1035

- Use client gateway for loading balances
- Add new models for balances
- Cleanup obsolete models
- Remove unused erc20 table
- Add migration

@gnosis/mobile-devs


data/src/main/java/io/gnosis/data/db/HeimdallDatabase.kt
```diff
@@ -3,28 +3,33 @@ package io.gnosis.data.db
...
-
     companion object {
         const val DB_NAME = "safe_db"
-        const val LATEST_DB_VERSION = 1
+        const val LATEST_DB_VERSION = 2
+
+        val MIGRATION_1_2 = object : Migration(1, 2) {
+            override fun migrate(database: SupportSQLiteDatabase) {
+                database.execSQL(
+                    """DROP TABLE IF EXISTS `erc20_tokens`"""
```
jpalvarezl, line 30: You don't need the multi line string delimiters here. But that's not important.

data/src/main/java/io/gnosis/data/repositories/TokenRepository.kt
```diff
@@ -1,28 +1,43 @@
...
-import io.gnosis.data.models.Erc20Token
+import io.gnosis.data.models.*
 import pm.gnosis.crypto.utils.asEthereumAddressChecksumString
 import pm.gnosis.model.Solidity
 import java.math.BigInteger
 
 class TokenRepository(
-    private val erc20TokenDao: Erc20TokenDao,
-    private val transactionServiceApi: TransactionServiceApi
+    private val transactionServiceApi: TransactionServiceApi,
```
jpalvarezl, line 11: Seems to me that this dependency is no longer used here.

data/src/main/java/io/gnosis/data/models/CoinBalances.kt
```diff
@@ -0,0 +1,33 @@
...
+data class TokenInfo(
+    @Json(name = "type") val tokenType: TokenType,
+    @Json(name = "address") val address: Solidity.Address,
+    @Json(name = "decimals") val decimals: Int = 0,
+    @Json(name = "symbol") val symbol: String = "",
+    @Json(name = "name") val name: String = "",
+    @Json(name = "logoUri") val logoUri: String?
+)
+
+enum class TokenType { ETHER, ERC20, ERC721 }
```
jpalvarezl, line 33: If the models from the app look exactly the same as the ones from the gateway DTO , does it make sense to redefine them?

data/src/main/java/io/gnosis/data/repositories/TokenRepository.kt
```diff
@@ -1,28 +1,43 @@
...
-            .map {
-                val token = it.tokenAsErc20Token()
-                erc20TokenDao.insertToken(token)
-                Balance(token, it.balance, it.balanceUsd)
-            }
+    suspend fun loadBalanceOf(safe: Solidity.Address): CoinBalances {
+        val response = gatewayApi.loadBalances(safe.asEthereumAddressChecksumString())
+        return CoinBalances(response.fiatTotal, response.items.map {
+
+            if (it.tokenInfo.address == null)
```
jpalvarezl, line 19: It's good that you don't rely on the `TokenInfo::type` here, there is been some discussion about that and there might be changes in the API for the type (because of xDai) this will definitely make things easier. I hope iOS also doesn't use `TokenType::Ether`

data/src/main/java/io/gnosis/data/repositories/TokenRepository.kt
elgatovital: it used for collectibles

data/src/main/java/io/gnosis/data/models/CoinBalances.kt
elgatovital: it is cleaner to split dto's and app models. Currently, they are practically identical but this will allow more flexibility later if something has to be adjusted

app/src/main/java/io/gnosis/safe/ui/assets/coins/CoinBalanceAdapter.kt
```diff
@@ -6,41 +6,80 @@ import androidx.recyclerview.widget.RecyclerView
...
+                holder.bind(total, balanceFormatter)
+            }
+            is CoinBalanceViewHolder -> {
+                holder.bind(balances[position - 1], balanceFormatter)
+            }
+        }
+    }
 
-    override fun getItemCount(): Int = balances.size
+    override fun getItemCount(): Int = if (balances.isNotEmpty()) balances.size + 1 else 0
```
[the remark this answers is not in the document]
elgatovital: refactored

biafra23:
This PR breaks the ETH balance display. It is wrongfully `>999T` for the following safe: 0x7cd310A8AeBf268bF78ea16C601F201ca81e84Cc 

It should be `8.8818`

elgatovital:
fixed

data/src/main/java/io/gnosis/data/repositories/TokenRepository.kt
jpalvarezl: Actually we do have an endpoint for collectibles in the gateway: `/v1/safes/<safe_address>/collectibles` it just forwards the request to the backend and caches the result. Just letting you know in case you want to implement it. The data models are the same so you would just have to add the method to the `GatewayApi` interface only, AFAICT.

data/src/main/java/io/gnosis/data/models/CoinBalances.kt
jpalvarezl: But the gateway provides objects based on the mocks done for the app's UI. This is actually redundant and confusing. If you need to add a layer of mapping it could be done at a later point. I think we also have the same issue with the transaction list where we are mapping structures that were made essentially looking at the apps UI anyway.

data/src/main/java/io/gnosis/data/repositories/TokenRepository.kt
elgatovital: not in the scope of this pr

data/src/main/java/io/gnosis/data/models/CoinBalances.kt
elgatovital: I would move this discussion out of this pr. We currently do have a mapping from our dto's to models as a general approach. I personally think it makes architecture more flexible, cleaner and ready for possible future changes without introducing too much overhead. We can discuss it later in more detail and create refactoring task if needed.

data/src/main/java/io/gnosis/data/db/HeimdallDatabase.kt
```diff
@@ -3,28 +3,33 @@ package io.gnosis.data.db
...
     abstract fun safeDao(): SafeDao
 
-    abstract fun erc20TokenDao(): Erc20TokenDao
-
     companion object {
         const val DB_NAME = "safe_db"
-        const val LATEST_DB_VERSION = 1
+        const val LATEST_DB_VERSION = 2
+
+        val MIGRATION_1_2 = object : Migration(1, 2) {
```
biafra23, line 27: Can we have a migration from 2 -> 1. So we can easily keep our safes when going back a db version. I had to delete all my safes because the app would crash, when I installed a build from master (same versionId.).

elgatovital: it's very unusual. Migrations are applied in an ascending order. I don't think it is possible

jpalvarezl (approved):

data/src/main/java/io/gnosis/data/models/CoinBalances.kt
jpalvarezl: I would argue that the sole purpose of the gateway is to prevent having to do this sort of mapping. We are introducing complexity for the sake of a potential future that is not even possible as far as the purpose of the gateway is concerned.

Issue #1042, opened 2020-11-03
[Android]Crash  AndroidKeyStorage.java line 61


liliya-soroka:
**Describe the bug**
The app is crashed for the Nexus 5X - https://console.firebase.google.com/u/2/project/safe-firebase-mainnet/crashlytics/app/android:io.gnosis.safe/issues/7e1ac6a40ec5eb1a109d4f2f9671e35c?time=last-seven-days&sessionEventKey=5F9FD98F0170000135868FAB917DC526_1468911219262101145

https://issuetracker.google.com/issues/37095309

**Smartphone (please complete the following information):**
 - Device: Nexus 5X
 - Android Version: 6.0.1
 - App Version 2.4.0(1384)

**Environment**
prod mainnet

**Additional context**

biafra23:
Similar problem: https://stackoverflow.com/questions/35955604/android-keystore-failed-to-generate-self-signed-certificate-invalid-date-str

Seems to be a rtl locale problem.

biafra23:
These might be the steps to reproduce (Need Android M device to verify):
- Switch phone to RTL (Change language to Persian or Hebrew make sure, you can go back!) via developer options -> Force RTL layout direction
- Delete all app data (or remove the app)
- Start app
- Add safe
- Go to tx list
- Observe crash

biafra23:
Related? https://issuetracker.google.com/issues/37095309

liliya-soroka:
Verified
app version 2.4.0-1475

PR #1056, opened 2020-11-09
AndroidKeyStorage crash

biafra23:
Handles #1042 

- Remove RTL "support" from app
- Temporarily change locale to allow key creation on Android M and earlier

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/di/modules/ApplicationModule.kt
```diff
@@ -231,7 +235,24 @@ class ApplicationModule(private val application: Application) {
...
     @Provides
     @Singleton
-    fun providesKeyStorage(@ApplicationContext context: Context): KeyStorage = AndroidKeyStorage(context)
+    fun providesKeyStorage(@ApplicationContext context: Context): KeyStorage =
+        if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.M) {
+            val localeCopy = Locale.getDefault()
+            setLocale(context, Locale.ENGLISH)
+            val androidKeyStorage = AndroidKeyStorage(context)
+            setLocale(context, localeCopy)
+            androidKeyStorage
```
jpalvarezl, line 244: Are we doing a quick locale swap here just for the key generation step?

app/src/main/java/io/gnosis/safe/di/modules/ApplicationModule.kt
```diff
@@ -231,7 +235,24 @@ class ApplicationModule(private val application: Application) {
...
+        } else {
+            AndroidKeyStorage(context)
+        }
+
+    private fun setLocale(context: Context, locale: Locale) {
```
jpalvarezl, line 249: maybe we could have a `private fun Context.setLocale()` method instead? Although it is not important.

app/src/main/java/io/gnosis/safe/di/modules/ApplicationModule.kt
```diff
@@ -231,7 +235,24 @@ class ApplicationModule(private val application: Application) {
 
     @Provides
     @Singleton
-    fun providesKeyStorage(@ApplicationContext context: Context): KeyStorage = AndroidKeyStorage(context)
+    fun providesKeyStorage(@ApplicationContext context: Context): KeyStorage =
+        if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.M) {
```
jpalvarezl, line 239: Do we need to do the API level check? I suppose we could just always use `Locale.ENGLISH` specifically for `KeyStorage` as well, then there would be fewer surprises.

app/src/main/java/io/gnosis/safe/di/modules/ApplicationModule.kt
biafra23: Yes. That is the solution I fond on the internet :-/

app/src/main/java/io/gnosis/safe/di/modules/ApplicationModule.kt
biafra23: Yes. We could. My idea was to remove this check and the branch, when have minSdkVersion > M.

app/src/main/java/io/gnosis/safe/di/modules/ApplicationModule.kt
elgatovital: this should be marked with //FIXME followed by a description of the reason for this workaround

biafra23: Added in fd5fbe51fb743768962bdba6212d18fe08007c17

app/src/main/java/io/gnosis/safe/di/modules/ApplicationModule.kt
jpalvarezl: Would it be better to always use `Locale.English` independently of the SDK version for `bouncy_castle`? Having this small fork in logic here could be problematic as we would have small changes in behaviour depending on the device.

elgatovital (approved):

app/src/main/java/io/gnosis/safe/di/modules/ApplicationModule.kt
biafra23: My risk assessment is like this: 
- If the locale switching turns out to be a problem, then it is only a problem on a minority of devices )if we have this branch logic)
- Google considered this a bug and fixed it, so it should not occur in devices after M

jpalvarezl: That makes sense. I understand now how we are better covered with this approach. Thanks!

jpalvarezl (approved):

DmitryBespalov:
sweet! 🍬

Issue #1046, opened 2020-11-05
Use Collectibles endpoint from the client gateway


jpalvarezl:
The client gateway relays the collectibles endpoint from the transaction service adding a layer of caching to the request and alleviating the transaction service.

There are no changes to be made to the models returned by the endpoint. Only the service to which the request is sent.
The gateway endpoint: `/v1/safes/<safe_address>/collectibles`

May need to adjust the tests

liliya-soroka:
Reopened:
"An error occurred" message appears instead of collectibles list for the safe 0xD2b3D480Db9Fb08Ea55c349C0299b81892F31544 , prod , mainnet
Note: the safe contains spam erc721 tokens that are marked as Unknow
app version 2.4.0 -1475

liliya-soroka:
I think that the issue https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-android/1073 and https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-android/1046 may be connected, as I can not add the same safes that have problems with collectables

biafra23:
@liliya-soroka I cannot reproduce this on 2.5.0-1531 (prod mainnet)

liliya-soroka:
Verified
the issue was with the urls for staging and prod mainnet
app version 2.6.0 -1534rc

PR #1059, opened 2020-11-10
Using gateway endpoint for collectibles

jpalvarezl:
Handles #1046 

- Using endpoint from gateway for `collectibles`
- Adjusted mock calls verifications

@gnosis/mobile-devs


data/src/main/java/io/gnosis/data/repositories/TokenRepository.kt
```diff
@@ -39,9 +38,8 @@ class TokenRepository(
         })
     }
 
-    //FIXME: use client gateway (grouping and sorting will be done on the backend side)
     suspend fun loadCollectiblesOf(safe: Solidity.Address): List<Collectible> =
-        transactionServiceApi.loadCollectibles(safe.asEthereumAddressChecksumString())
+        gatewayApi.loadCollectibles(safe.asEthereumAddressChecksumString())
             .asSequence()
             .groupBy {
                 it.address
```
elgatovital, line 45: Does client gw perform grouping?

elgatovital:
content of ressources/load_collectibles.json should be replaced with correct response data

jpalvarezl:
> The client gateway relays the collectibles endpoint from the transaction service adding a layer of caching to the request and alleviating the transaction service.
> There are no changes to be made to the models returned by the endpoint. 

There is no change in the structure. The json is still valid.

elgatovital:
I propose to put assets related models to models.assets package

jpalvarezl:
Re-structuring the project is beyond the scope of the issue #1046 . If you would like to do that let's open a new issue for it.

jpalvarezl:
I would also like to remind you that my time in this sprint is very limited :)

data/src/main/java/io/gnosis/data/repositories/TokenRepository.kt
jpalvarezl: Nope. Just caching. There is no change in the `DTO` classes. The response is the same as from the TX service, as specified in the ticket.

elgatovital (approved):

Issue #1060, opened 2020-11-10
Add staging mainnet to the continuous delivery


DmitryBespalov:
As we discussed in the last retro, Liliya would like to have Staging Mainnet configuration available on both platforms so that it allows her to better track down issues - whether they're mobile client related or backend related.

liliya-soroka:
Verified
app version 2.4.0 -1468

PR #1061, opened 2020-11-11
Add flavor for mainnet staging

elgatovital:
Handles #1060

@gnosis/mobile-devs

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -121,6 +121,21 @@ android {
             }
         }
 
+        mainnet_staging {
+            initWith debug
+            applicationIdSuffix ".mainnet_staging"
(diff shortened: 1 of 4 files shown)


app/build.gradle
```diff
@@ -121,6 +121,21 @@ android {
             }
         }
 
+        mainnet_staging {
```
jpalvarezl, line 124: Should we rename this to `mainnet_internal` or alternatively, rename `internal` to `staging` ?

elgatovital: mainnet_staging is more descriptive. internal version could be renamed to rinkeby_staging, but I would not do it now. Currently, everyone is aware of what internal version is composed of.

jpalvarezl (approved):

Issue #972, opened 2020-10-05
[Anroid]Crash : AddressInputHelper.java line 48 io.gnosis.safe.helpers.AddressInputHelper$$special$$inlined$apply$lambda$1.onClick


liliya-soroka:
AddressInputHelper.java line 48
io.gnosis.safe.helpers.AddressInputHelper$$special$$inlined$apply$lambda$1.onClick
https://console.firebase.google.com/u/2/project/safe-firebase-mainnet/crashlytics/app/android:io.gnosis.safe/issues/02c5703e8496bea46f1af1f5a788ccd9?time=last-seven-days&sessionEventKey=5F7A127A01F800013BF122BF88EE2575_1458277850370572383
**To Reproduce**
in progress..... 

**Expected behavior**

**Screenshots**

**Smartphone (please complete the following information):**
 - Device: Pixel 2 XL
 - Android Version: 11
 - App Version 2.1.0 (1058)

**Environment**
Prod, mainnet

**Additional context**

liliya-soroka:
Verfied
app version 2.4.0 -1487

Issue #971, opened 2020-10-05
[Android] Crash: QRCodeScanActivity.java line 77 io.gnosis.safe.qrscanner.QRCodeScanActivity$Companion.startForResult ( researching for the steps)


liliya-soroka:
**Describe the bug**
https://console.firebase.google.com/u/2/project/safe-firebase-mainnet/crashlytics/app/android:io.gnosis.safe/issues/e5ce9d75d48b9d8bc2a34c14d4bd1df2?time=last-seven-days&sessionEventKey=5F7AA6AC0316000122AEB6A0F8708344_1458515071215076269

**To Reproduce**
1. in progress of researching

**Expected behavior**
No crash

**Smartphone (please complete the following information):**
 - Device: Galaxy Note10
 - Android Version: 10
 - App Version 2.1.0 (1058)

**Environment**
Prod , mainnet
**Additional context**

DmitryBespalov:
Might be caused by a change in screen orientation while using the camera

liliya-soroka:
Verified
app version 2.4.0 -1487

PR #1062, opened 2020-11-11
Hide input dialog if fragment is detached from context

elgatovital:
Handles #971 
Handles #972

After thorough investigation I was able to reproduce both crashes. The reason for both crashlytics issues is that fragment is not attached to context as callbacks are being invoked. This can happen if the input dialog is open and the user clicks on an incoming notification that navigates away from AddSafeFragment, not leaving the app. The input bottom sheet remains open though, thus any click on bottom sheet items results in a crash. 
Hiding input dialog solves the problem.

@gnosis/mobile-devs


DmitryBespalov:
Nice finding! 🕵️ 👏

jpalvarezl (approved):
Nice simple solution :)

Issue #1045, opened 2020-11-05
[Android]num_keys_imported is not tracked if there is no loaded safe in the app


liliya-soroka:
**Describe the bug**
num_keys_imported is not tracked if there is no loaded safe in the app

**To Reproduce**
1. Clean app install
2. Go to the app settings
3. Import owner key
4. go over the app screens

**Expected behavior**
num_keys_imported should be tracked when there is no loaded safe.
Make sure that it's also tracked if the all safes are removed 

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.4.0 -debug version 

**Environment**
staging, rinkeby
**Additional context**

liliya-soroka:
Verified
app version 2.5.0 -1595

PR #1063, opened 2020-11-11
Set num_keys_imported on startup

biafra23:
Handles #1045 

This doesn't exactly cover 1045. I was unable to reproduce 1045. But what I found was that we do not set the value if it doesn't change. Now we set the value on startup. 

Maybe using user properties is not the right way to track imported and deleted keys. It seems the update is only sent on app start. And also only if it changed to the last value. I am currently trying to find out if that is the expected behavior.

@gnosis/mobile-devs


app/src/test/java/io/gnosis/safe/ui/splash/SplashViewModelTest.kt
```diff
@@ -119,9 +126,12 @@ class SplashViewModelTest {
         coEvery { notificationRepository.checkPermissions() } returns true
```
elgatovital, line 126: better to use verifySequence here and also check that invocations happen in the correct order

elgatovital (approved):

Issue #993, opened 2020-10-14
[Android]Crash: ContextUtilsKt.java line 38( take attention during full regression for safe settings)


liliya-soroka:
**Describe the bug**
rare case:
14 Oct the issue happened on QA testing device during removing the safes from the app. I had about 5 safes and started to remove one by one. At the end the safe settings with null as values for data on the safe settings appears and on the swipe from right to the left the app was crashed 
https://console.firebase.google.com/u/2/project/safe-firebase-mainnet/crashlytics/app/android:io.gnosis.safe/issues/00a8bc9c942ad376446b77f6973b5ac6?time=last-seven-days&sessionEventKey=5F86E3160050000102E2992A1DD1A772_1461927774085411863
**To Reproduce**
1. Add about 5 safes
2. Go to safe settings
3. Remove safes without waiting till the safe settings is updated with info about the next active safe
4. searching for the last step..

**Expected behavior**
no crash

**Screenshots**

**Smartphone (please complete the following information):**
 - Device: Pixel 4
 - Android Version: 11
 - App Version 2.3.0 -1278

**Environment**
Prod mainnet
**Additional context**

elgatovital:
not reproducible

liliya-soroka:
as It was agreed during the weekly sync the issue is moved to the New issues for replanning and research it more

elgatovital:
this one is actually fixed with https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-android/1064

liliya-soroka:
Verified
app version 2.4.0 -1487

PR #1064, opened 2020-11-11
Fix error snackbar in share dialog

elgatovital:
Handles #993

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/safe/share/ShareSafeDialog.kt
```diff
@@ -111,7 +110,7 @@ class ShareSafeDialog : BaseViewBindingDialogFragment<DialogShareSafeBinding>()
             safeFields.visible(false)
         }
         Timber.e(error)
-        snackbar(requireView(), Snackbar.LENGTH_LONG, R.string.error_invalid_safe)
```
biafra23, line 114: Why did this compile? Is Snackbar.LENGTH_LONG a valid StringRes?

biafra23 (approved):
Good catch! 🚀

app/src/main/java/io/gnosis/safe/ui/safe/share/ShareSafeDialog.kt
elgatovital: its an int, so yes

PR #1066, opened 2020-11-11
Add launcher icons for mainnet staging

elgatovital:
@gnosis/mobile-devs


biafra23 (approved):
LGTM 👍

app/src/mainnet_staging/res/values/static_strings.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="app_name" translatable="false">Safe Multisig - Mainnet Staging</string>
```
biafra23, line 3: Do you see the end of this name in your home screen?

Maybe we could rename it to Mainnet Staging - Safe Multisig?

elgatovital: or just Mainnet Staging. It is clear anyway that the app is Safe Multisig

elgatovital: but this name is mainly for app distribution screen. App is identifiable by the icon and nobody should use it, but us. So I think it's fine to just leave it as is and continue using our established app naming schema :)

elgatovital: So you vote for Mainnet Staging?

Issue #597, opened 2020-05-11
Add Timber log trees per build type


biafra23:
Different environments might need different log levels and reporting of errors:

DEBUG -> log level high
RELEASE -> log level minimal and maybe sending error logs to crashlytics?

jpalvarezl:
@biafra23 Hey, is this done? I think timber is working correctly already, isn't it?

jpalvarezl:
If that is the case, let's close this.

biafra23:
It is not done.

liliya-soroka:
@elgatovital , @biafra23 , is the testing required for the task?
If yes, could you add info on how it can be checked, please?

liliya-soroka:
closed, no QA required.
The firebase will be checked time to time during the regression testing

PR #1068, opened 2020-11-12
Add crash reporting timber tree

elgatovital:
Handles #597

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/HeimdallApplication.kt
+++ b/app/src/main/java/io/gnosis/safe/HeimdallApplication.kt
@@ -1,7 +1,9 @@
 package io.gnosis.safe
 
 import android.content.Context
+import android.util.Log
 import androidx.multidex.MultiDexApplication
+import com.google.firebase.crashlytics.FirebaseCrashlytics
 import io.gnosis.safe.di.ComponentProvider
 import io.gnosis.safe.di.components.ApplicationComponent
 import io.gnosis.safe.di.components.DaggerApplicationComponent
@@ -23,9 +25,13 @@ class HeimdallApplication : MultiDexApplication(), ComponentProvider {
 
     override fun onCreate() {
         super.onCreate()
+
         if (BuildConfig.DEBUG) {
             Timber.plant(DebugTree())
+        } else {
+            Timber.plant(CrashReportingTree())
         }
+
         component.appInitManager().init()
 
         try {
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/gnosis/safe/HeimdallApplication.kt
```diff
@@ -42,3 +49,26 @@ class HeimdallApplication : MultiDexApplication(), ComponentProvider {
         }
     }
 }
+
+private class CrashReportingTree : Timber.Tree() {
```
biafra23, line 53: Is this really a crash reporting tree? Isn't it more an exception reporting tree? Crashes are already reported. This is about reporting otherwise swallowed exceptions?

app/src/main/java/io/gnosis/safe/HeimdallApplication.kt
```diff
@@ -42,3 +49,26 @@ class HeimdallApplication : MultiDexApplication(), ComponentProvider {
...
+    override fun log(priority: Int, tag: String?, message: String, t: Throwable?) {
+
+        if (priority == Log.VERBOSE || priority == Log.DEBUG) {
+            return
+        }
+        
+        t?.let {
+            if (priority == Log.ERROR) {
+                with(FirebaseCrashlytics.getInstance()) {
+                    setCustomKey(CRASHLYTICS_KEY_LOCALE, Locale.getDefault().language)
```
biafra23, line 64: Why only report the language? Why not language and region/country? en_US is different from en_GB or pt_PT/pt_BR. 
I'd go for: 
```
                    val locale = Locale.getDefault()
                    setCustomKey(CRASHLYTICS_KEY_LOCALE, "${locale.isO3Language}_${locale.isO3Country}")
```

app/src/main/java/io/gnosis/safe/HeimdallApplication.kt
```diff
@@ -42,3 +49,26 @@ class HeimdallApplication : MultiDexApplication(), ComponentProvider {
...
+                    recordException(it)
```
biafra23, line 65: If we log in this way, do we see in the Crashlytics dashboard that it was a non crashing exception?

biafra23 (approved):
Please consider my questions/suggestions. But they are non blocking.

app/src/main/java/io/gnosis/safe/HeimdallApplication.kt
elgatovital: yes, Should I rename it to ErrorReportingTree?

app/src/main/java/io/gnosis/safe/HeimdallApplication.kt
elgatovital: sure, why not

app/src/main/java/io/gnosis/safe/HeimdallApplication.kt
elgatovital: They should be reported as Non-fatal exceptions

elgatovital: image
you can set this filter

Issue #1055, opened 2020-11-09
Combine dto's & models: tx details


elgatovital:
io.gnosis.data.backend.dto.GateTransactionDetails
io.gnosis.data.models.TransactionDetails

Transaction related models should be placed to io.gnosis.data.models.transaction.details
Explicitly annotate json attributes

liliya-soroka:
Verified
app version 2.5..0 -1531

Issue #814, opened 2020-08-22
SafeInfo fetch during Transaction List loading + DTOs


jpalvarezl:
**TODOs**
- [x] Remove fetching of `SafeInfo` as it is no longer necessary
- [x] Use `SettingsInfo` object instead of `dataDecoded` for `SettingsChange` (In the list)
- [x] Re-use `TransactionInfo` from details into the list
- [x] Match the structure from the gateway and only do types mapping (for `timestamp` for instance, `Long.toDate()`) 
- ~~[ ] Try reducing the amount of types required by the `RecyclerView`~~ **This will change with v2 of the tx list endpoint anyway**

io.gnosis.data.backend.dto.GateTransaction
io.gnosis.data.models.Transaction

Transaction related models should be placed to io.gnosis.data.models.transaction
Explicitly annotate json attributes

**Is your feature request related to a problem? Please describe.**
After integrating the client gateway, there is no need to fetch the `SafeInfo` at the time of loading the transaction list, as its only purpose was to gather the data to calculate the transaction `status`. This is no longer the case, as the `status` is calculated in the gateway; this request has been rendered unnecessary.

`SettingsInfo` contains parameters fetched from `DataDecoded` using the index in the `abi` which is considered the preferred way. The app uses the parameter name adding a failing point for contracts with parameters with the same name for a different type, for example.

`TransactionInfo` is the same class both in the `Summary` and `Detailed` view of the data. This should be leveraged by the app, to avoid duplication.

~~The recyclerView for the TX list consumes 9 different types. Potentially this can be reduced by adding small decision logic in the viewHolder/viewModel. Having so many types makes the code a bit error prone as there is a lot of duplication for data mapping into view types. Making it hard to identify which type specifically may be causing an issue.~~ **This will change with the tx list v2 anyway**
 
**Describe the solution you'd like**
See above section

liliya-soroka:
Verified
app version 2.5.0-1495

PR #1069, opened 2020-11-12
Refactor TX list and details

jpalvarezl:
Handles #814 #1055 

- Flattened models in the app
- Tx list and tx details use models directly from the gateway
- Tests were left as untouched as possible

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewModel.kt
```diff
@@ -591,5 +568,8 @@ fun TransactionView.isHistory() = when (status) {
 fun Solidity.Address.getVersionForAddress(): String = masterCopyVersion(this) ?: SAFE_MASTER_COPY_UNKNOWN_DISPLAY_STRING
 
 fun Transaction.canBeSignedByOwner(ownerAddress: Solidity.Address?): Boolean {
-    return missingSigners?.contains(ownerAddress) == true
+    return executionInfo?.missingSigners?.contains(ownerAddress) == true
 }
+
+//TODO add unit tests
```
jpalvarezl, line 574: Add tests

elgatovital (approved):
LGTM! 👍 Please create an issue for moving decoded param classes from backend.dto.TransactionDto to models.transaction

jpalvarezl:
I opened the issue for Android: https://github.com/gnosis/safe-android/issues/1070

and for the gateway: https://github.com/gnosis/safe-client-gateway/issues/190

elgatovital:
image
same for item_tx_change_mastercopy, etc. All tx item layouts should be checked and adjusted

jpalvarezl:
What is wrong with that one?

elgatovital:
it has to be right aligned

jpalvarezl:
You are right, the `viewEnd` was the problem just like with `textEnd`. It was only for the master copy as far as I could see.

Issue #1070, opened 2020-11-12
Move DataDecoded related classes to the models.transaction package


jpalvarezl:
`DataDecoded` related classes should be moved to the `models.transactions` package.

liliya-soroka:
Verified
app version 2.5.0 -1495

PR #1075, opened 2020-11-16
Refactor data decoded classes

elgatovital:
Handles #1070

@gnosis/mobile-devs


data/src/main/java/io/gnosis/data/backend/dto/SafeInfoDto.kt
```diff
@@ -5,6 +5,7 @@ import com.squareup.moshi.JsonClass
 import pm.gnosis.model.Solidity
 import java.math.BigInteger
 
+@Deprecated("use client gw")
```
biafra23, line 8: 👍

biafra23 (approved):
LGTM 👍

PR #1077, opened 2020-11-16
Adjust proguard rules

elgatovital:
@gnosis/mobile-devs

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -73,8 +73,8 @@
 -keep @interface pm.gnosis.heimdall.data.adapters.DecimalNumber
 -keep @interface pm.gnosis.heimdall.data.adapters.BigDecimalNumber
 
--keep interface io.gnosis.data.backend.dto.** { *; }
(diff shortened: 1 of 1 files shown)


app/proguard-rules.pro
```diff
@@ -73,8 +73,8 @@
 -keep @interface pm.gnosis.heimdall.data.adapters.DecimalNumber
 -keep @interface pm.gnosis.heimdall.data.adapters.BigDecimalNumber
 
--keep interface io.gnosis.data.backend.dto.** { *; }
--keep class io.gnosis.data.backend.dto.** { *; }
```
biafra23, line 77: We still have CoinBalancesDto. Shouldn't we keep this until CoinBalancesDto and SafeInfoDto have been removed?

elgatovital: SafeInfoDto we don't need, and assets still work, because we have this rule

elgatovital:
```

biafra23 (approved):

Issue #964, opened 2020-09-30
Extract harcoded strings to strings.xml


elgatovital:
Strings used to display information should never be hardcoded. Use the Android resource framework to separate the localized aspects of the app.

Identified violations:

SafeRepository
```
const val DEFAULT_FALLBACK_HANDLER_DISPLAY_STRING = "DefaultFallbackHandler"
const val SAFE_MASTER_COPY_UNKNOWN_DISPLAY_STRING = "Unknown"
```

liliya-soroka:
Verfieid
app version 2.6.0 -1534rc

PR #1078, opened 2020-11-16
Resource strings

biafra23:
Handles #964 

- Extract DefaultFallbackhandler into string resource
- Extract implementation version into string resource
- mastercopy -> implementation

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/settings/safe/AdvancedSafeSettingsFragment.kt
```diff
@@ -91,14 +91,11 @@ class AdvancedSafeSettingsFragment : BaseViewBindingFragment<FragmentSettingsSaf
                 openable = false
                 name = getString(R.string.safe_settings_not_set)
             }
-            viewModel.isDefaultFallbackHandler(fallbackHandler) -> labeledAddress(fallbackHandler, R.string.safe_settings_default_fallback_handler)
-            else -> labeledAddress(fallbackHandler, R.string.safe_settings_unknown)
+            viewModel.isDefaultFallbackHandler(fallbackHandler) -> labeledAddress(fallbackHandler, io.gnosis.data.R.string.default_fallback_handler)
```
rmeissner, line 94: why full package path?

app/src/main/java/io/gnosis/safe/ui/settings/view/NamedAddressItem.kt
```diff
@@ -52,13 +52,13 @@ class NamedAddressItem @JvmOverloads constructor(
             field = value
         }
 
-    var name: String? = null
+    var name: Int? = null
```
rmeissner, line 55: If this is a res value I would annotate it as such

app/src/main/java/io/gnosis/safe/ui/settings/view/NamedAddressItem.kt
```diff
@@ -52,13 +52,13 @@ class NamedAddressItem @JvmOverloads constructor(
...
         set(value) {
-            if (value.isNullOrBlank()) {
+            if (value == null) {
```
rmeissner, line 57: wouldn't the elvis operator be nicer here?
```
value?.let{
  binding.name.visible(true)
  binding.name.setText(it)
} ?: run {
  binding.name.visible(false)
}
```

data/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="default_fallback_handler">DefaultFallbackHandler</string>
+    <string name="unknown_fallback_handler">Unknown</string>
+    <string name="empty_string"></string>
```
rmeissner, line 5: Does it really make sense to have an empty string in a translatable file? I would at least mark it as not translatable.

Also it feels weird to have string resources in the data package imo as these are more "app" (UI) specific

app/src/main/java/io/gnosis/safe/ui/settings/safe/AdvancedSafeSettingsFragment.kt
biafra23: I didn't know its not necessary for module. But it is not necessary.

biafra23: I assumed I would need it in any file that also uses apps R. But apparently only SafeRepository needs this import: 
`import io.gnosis.data.R`

app/src/main/java/io/gnosis/safe/ui/settings/view/NamedAddressItem.kt
biafra23: Is `run` necessary?

data/src/main/res/values/strings.xml
biafra23: You're right. Those string resources do not have to be there.

app/src/main/res/values/strings.xml
```diff
@@ -240,4 +235,13 @@
     <string name="confirm_transaction_dialog_message">You are about to confirm the transaction with your currently imported owner key. This confirmation is off-chain. The transaction should be executed separately in the web interface.</string>
     <string name="confirmation_successfully_submitted">Confirmation successfully submitted</string>
     <string name="ens_name_contains_illegal_character">Name contains invalid characters</string>
+
+    <string name="default_fallback_handler">DefaultFallbackHandler</string>
+    <string name="unknown_fallback_handler">Unknown</string>
+    <string name="empty_string" translatable="false" />
```
elgatovital, line 241: I think that's overkill. Why not just use ""

app/src/main/res/values/strings.xml
```diff
@@ -240,4 +235,13 @@
...
+    <string name="unknown_implementation_version">Unknown</string>
+    <string name="implementation_version_0_0_2">0.0.2</string>
```
elgatovital, line 243: these version strings could be moved to strings_static as they are not translatable

app/src/main/java/io/gnosis/safe/utils/StringUtils.kt
```diff
@@ -67,6 +73,25 @@ fun Solidity.Address.formatEthAddress(context: Context, prefixLength: Int = 6, s
...
+@StringRes
+fun Solidity.Address?.fallBackHandlerLabel(): Int =
+    if (this == DEFAULT_FALLBACK_HANDLER) {
+        R.string.default_fallback_handler
+    } else {
+        R.string.unknown_fallback_handler
+    }
+
+@StringRes
+fun Solidity.Address.implementationVersion(): Int {
```
elgatovital, line 85: it would be cleaner to move this to SafeRepository. Then knowledge of the safe, contracts, and versions would be encapsulated in one place

app/src/main/res/values/strings.xml
biafra23: "" is not an Int

biafra23: We could use null as in `Int?`

elgatovital (approved):

Issue #1053, opened 2020-11-09
Combine dto & models: Balances


elgatovital:
io.gnosis.data.backend.dto.CoinBalancesDto
io.gnosis.data.models.CoinBalances

Explicitly annotate json attributes for parsed parameters
Balance related models should be placed under io.gnosis.data.models.assets

liliya-soroka:
Verified
app version 2.6.0 -1534rc

PR #1083, opened 2020-11-17
Combine DTO and model layer

biafra23:
Handles #1053 

- Remove DTO layer
- adjust tests

@gnosis/mobile-devs


data/src/main/java/io/gnosis/data/repositories/TokenRepository.kt
```diff
@@ -23,7 +23,7 @@ class TokenRepository(
             else
```
biafra23, line 23: This logic (replacing null token address with etherum info) should be implemented by the client gateway, shouldn't it?

data/src/main/java/io/gnosis/data/repositories/TokenRepository.kt
```diff
@@ -23,7 +23,7 @@ class TokenRepository(
...
                 Balance(
                     TokenInfo(
```
elgatovital, line 25: TokenType parsing should be done by adapter. Also this mapping step is not needed at all, as everything can be parsed directly from client gw response

biafra23: It shouldn't be necessary. But we would need a custom adapter to implement this logic. I would rather have the client gateway send a response that doesn't need this logic. 

If you remove this code, tests will fail.

data/src/main/java/io/gnosis/data/repositories/TokenRepository.kt
```diff
@@ -13,27 +16,10 @@ class TokenRepository(
...
         return CoinBalances(response.fiatTotal, response.items.map {
-
-            if (it.tokenInfo.address == null || it.tokenInfo.address == ZERO_ADDRESS)
-                Balance(
-                    ETH_TOKEN_INFO,
-                    it.balance,
-                    it.fiatBalance
-                )
+            if (it.tokenInfo.address == ZERO_ADDRESS)
+                it.copy(tokenInfo = it.tokenInfo.copy(logoUri = "local::ethereum"))
```
biafra23, line 20: @jpalvarezl We should probably have this (logoUri for ETH transfers) in the client gateway?

elgatovital (approved):

PR #1090, opened 2020-11-19
Push notifications for signing transactions

elgatovital:
Handles #1079

- Add notification type for a confirmation request 
- Calculate signatures for owners
- Generate uuid for device registration

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/notifications/NotificationRepository.kt
+++ b/app/src/main/java/io/gnosis/safe/notifications/NotificationRepository.kt
@@ -38,18 +38,18 @@ class NotificationRepository(
         }
 
     private var deviceUuid: String?
-        get() = preferencesManager.prefs.getString(DEVICE_UUID, null)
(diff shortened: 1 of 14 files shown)


biafra23 (approved):
LGTM 
We should add error handling for certain cases, when the user needs to do something (in a future PR). For example when the users devices clock is off more than five minutes.

app/src/main/java/io/gnosis/safe/notifications/NotificationRepository.kt
```diff
@@ -241,6 +263,7 @@ class NotificationRepository(
 
     companion object {
         private const val DEVICE_UUID = "prefs.string.device_uuid"
+        private const val REGISTRATION_UPDATE_FAILED = "prefs.boolean.registration_update_failed"
```
biafra23, line 266: Should this variable have a KEY_ prefix?

elgatovital: sure

PR #1095, opened 2020-11-20
xDai

biafra23:
Handles #

- Introduces xdai flavour
- removes several unused BuildConfig variables
- Introduces BuildConfig.NATIVE_CURRENCY_SYMBOL to be used instead of hard coded "ETH"

This client shows ETH if sent by the backend for example in balances: https://safe-client.xdai.gnosis.io/v1/safes/0xcba12315cc838375F0e1E9a9f5b2aFE0196B07B6/balances/usd 

@gnosis/mobile-devs


data/src/main/java/io/gnosis/data/repositories/TokenRepository.kt
```diff
@@ -65,7 +66,7 @@ class TokenRepository(
             TokenType.ETHER,
             ZERO_ADDRESS,
             18,
-            "ETH",
+            BuildConfig.NATIVE_CURRENCY_SYMBOL,
             "Ether",
```
rmeissner, line 70: Should this and the icon below be adjusted too?

data/src/main/java/io/gnosis/data/repositories/TokenRepository.kt
```diff
@@ -65,7 +66,7 @@ class TokenRepository(
             TokenType.ETHER,
```
rmeissner, line 66: I would rename this constant to NATIVE_TOKEN_INFO

data/src/main/java/io/gnosis/data/repositories/TokenRepository.kt
```diff
@@ -65,7 +66,7 @@ class TokenRepository(
...
             ZERO_ADDRESS,
             18,
-            "ETH",
+            BuildConfig.NATIVE_CURRENCY_SYMBOL,
             "Ether",
             "local::ethereum"
```
rmeissner, line 71: Maybe changing this string to `local::native_token` and adding a different resource for the xdai build would make sense

data/src/main/java/io/gnosis/data/repositories/TokenRepository.kt
biafra23: Yeah. But to what? :-)

.buildkite/deployment.yml
```diff
@@ -9,6 +9,11 @@ steps:
     key: 'upload-firebase-unsafe'
     command: 'ci/upload_app_distribution.sh internal-beta unsafe'
 
+  - label: 'Upload App Distribution xDai'
+    if: build.branch == "master" || build.branch == "release" || build.branch =~ /^v[0-9]+.*/
```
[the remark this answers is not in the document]
biafra23: If we do testing (of the xdai client) at some point,  this would allow to test non-releases earlier.

data/src/main/java/io/gnosis/data/repositories/TokenRepository.kt
rmeissner: xDai I would say

.buildkite/deployment.yml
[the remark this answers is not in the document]
rmeissner: But we do not do this right now, and therefore it just adds additional build time without being really useful. I we require this in the future we can add it

data/src/main/java/io/gnosis/data/repositories/TokenRepository.kt
rmeissner: Shouldn't you use `NATIVE_CURRENCY_NAME` from the BuildConfig here :D

data/src/test/java/io/gnosis/data/repositories/TokenRepositoryTest.kt
```diff
@@ -60,9 +61,9 @@ class TokenRepositoryTest {
                                 tokenType = TokenType.ETHER,
```
elgatovital, line 61: TokenType.ETHER should be renamed to TokenType.NATIVE_CURRENCY

data/src/main/java/io/gnosis/data/adapters/TokenTypeAdapter.kt
```diff
@@ -13,7 +13,7 @@ class TokenTypeAdapter {
     @FromJson
     fun fromJson(tokenType: String): TokenType =
         when (tokenType) {
-            "ETHER" -> TokenType.ETHER
+            "ETHER" -> TokenType.NATIVE_CURRENCY
```
elgatovital, line 16: We need a FIXME here and on the client gw side it has to be adjusted also. Something like "NATIVE" would be nice. Could you create an issue for client gw please?

data/src/main/java/io/gnosis/data/repositories/TokenRepository.kt
```diff
@@ -61,13 +62,13 @@ class TokenRepository(
 
     companion object {
         val ZERO_ADDRESS = Solidity.Address(BigInteger.ZERO)
-        val ETH_TOKEN_INFO = TokenInfo(
-            TokenType.ETHER,
+        val NATIVE_CURRENCY_INFO = TokenInfo(
```
elgatovital, line 65: I think we should remove this NATIVE_CURRENCY_INFO in the near future as not every native currency has 18 decimals and this info should be delivered by client gw

data/src/main/java/io/gnosis/data/adapters/TokenTypeAdapter.kt
biafra23: I think there is already a ticket. But I am not sure it covers this: https://github.com/gnosis/safe-client-gateway/issues/179

data/src/main/java/io/gnosis/data/repositories/TokenRepository.kt
biafra23: This seems to be related: https://github.com/gnosis/safe-client-gateway/issues/179

data/src/main/java/io/gnosis/data/adapters/TokenTypeAdapter.kt
biafra23: I added a FIXME

elgatovital (approved):

Issue #1089, opened 2020-11-19
[Android] Data can't be loaded is missed on the tx list /transaction details pages


liliya-soroka:
**Describe the bug**
There is no "Data can't be loaded.." on the tx list page in case of errors

**To Reproduce**
1. Turn the internet connection off
2. Open transaction list /pull to refresh tx list
Current result: Snackbar with the error message appears but there is no " Data cannot be loaded". The same for the transaction details

**Expected behavior**
"Data cannot be loaded" should be displayed together with snack bar as on the balance and safe settings pages
**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/eda72725-56fe-4a4e-b48b-da59b9f919f3)
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/a5d8e84d-d02a-484c-aa8d-54cac22b52c8)
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/593e9847-ade6-4f5f-8208-f9a60f8be7c5)

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.5.0-1521

**Environment**
staging, rinkeby
**Additional context**

biafra23:
Just to clarify:
>  "Data cannot be loaded" should be displayed together with snack bar as on the balance and safe settings pages

If there was already data shown and then the internet is turned off. Then the data should be still shown, if pull to refresh is used to refresh data.

liliya-soroka:
Verified
app version 2.5.0 -1595

liliya-soroka:
Verified
app version 2.5.0 -1595

PR #1097, opened 2020-11-23
Show data cannot be loaded for tx list & details placeholder when data could not be loaded

elgatovital:
Handles #1089

@gnosis/mobile-devs


biafra23 (approved):
Looking good! 👍

PR #1098, opened 2020-11-23
Add flavour profile and activate leak canary for it

biafra23:
Handles #868

- adds flavor profile
- Activate leak canary for this flavor

**Remark: You need to add a "profile" client  config in google-services.json to build with this flavor**

@gnosis/mobile-devs

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -75,6 +75,13 @@ android {
             buildConfigField javaTypes.STRING, "SAFE_CREATION_FUNDER", asString(getKey("SAFE_CREATION_FUNDER", "0xAb8C18E66135561676f0781555D05CF6B22024A3"))
         }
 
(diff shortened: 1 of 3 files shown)


elgatovital:
I don't think we need it. Profiling can be done locally with any available tools of choice and not necessarily leak canary. Having profiling flavor available to others via app distribution has little value, because only android devs will analyse it. Thus it would be better to include regular profiling task in our workflow instead of having this flavor. I think we should close this pr.

biafra23:
I agree, that we do not need any distribution channel for this profile. And this PR does not add any. But I think we should have minimal friction on building a version with leak canary. And this PR does exactly this. I can also imagine having other profiling libraries added in the future.

elgatovital (approved):

Issue #842, opened 2020-08-27
New format for errors


DmitryBespalov:
Spec: https://docs.google.com/document/d/1foQ9DSMgO2iN-6u4JsJIDJWB7mUHX-D8ApZWvx6oKAo/edit#

liliya-soroka:
https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-android/1109

liliya-soroka:
The task is closed as only one bug is left :
https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-android/1120
app version :  2.7.0 -1634rc

PR #1105, opened 2020-11-26
Use new format for errors

elgatovital:
Handles #842

https://docs.google.com/document/d/1foQ9DSMgO2iN-6u4JsJIDJWB7mUHX-D8ApZWvx6oKAo/edit#

- Introduce new error classes
- Map exceptions to errors
- Show error messages using the new format

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ErrorHandler.kt
+++ b/app/src/main/java/io/gnosis/safe/ErrorHandler.kt
@@ -1,0 +1,103 @@
+package io.gnosis.safe
+
+import android.content.Context
+import androidx.annotation.StringRes
+import io.gnosis.data.repositories.EnsInvalidError
+import io.gnosis.data.repositories.EnsResolutionError
+import io.gnosis.data.repositories.EnsReverseRecordNotSetError
+import io.gnosis.safe.helpers.Offline
+import io.gnosis.safe.ui.safe.add.InvalidName
+import io.gnosis.safe.ui.safe.add.UsedSafeAddress
+import pm.gnosis.utils.HttpCodes
+import pm.gnosis.utils.exceptions.InvalidAddressException
+import retrofit2.HttpException
+import java.net.ConnectException
+import java.net.SocketTimeoutException
+import java.net.UnknownHostException
+import javax.net.ssl.SSLHandshakeException
+
+enum class Error(
+    val id: Int,
+    @StringRes
+    val reason: Int,
+    @StringRes
+    val howToFix: Int,
+    val trackingRequired: Boolean = false
+
+) {
+    // Network-related errors
+    ERROR_101(101, R.string.error_network_no_internet_reason, R.string.error_network_no_internet_fix),
+    ERROR_102(102, R.string.error_network_ssl_reason, R.string.error_network_ssl_fix),
+    ERROR_103(103, R.string.error_network_timeout_reason, R.string.error_network_timeout_fix) {
+        override fun message(context: Context): String {
+            return "${context.getString(reason, 30)}. ${context.getString(howToFix)}. (${context.getString(R.string.error_id, id)})"
+        }
+    },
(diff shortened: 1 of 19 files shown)


app/src/main/java/io/gnosis/safe/ui/assets/coins/CoinsFragment.kt
```diff
@@ -69,14 +68,8 @@ class CoinsFragment : BaseViewBindingFragment<FragmentCoinsBinding>() {
...
-                                when (action.error) {
-                                    is Offline -> {
-                                        snackbar(requireView(), R.string.error_no_internet)
-                                    }
-                                    else -> {
-                                        snackbar(requireView(), action.error.getErrorResForException())
-                                    }
-                                }
+                                val error = action.error.toError()
+                                snackbar(requireView(), error.message(requireContext(), R.string.error_description_assets_coins))
```
biafra23, line 72: The error messages are now so long it is hard to read them completely even with Snackbar.LENGTH_LONG. Could we go with LENGTH_INDEFINITE?

elgatovital: I would leave it like this for now and see what users will say. We can adjust it later.

biafra23: Have you tried it? Were you able to read the full message before it disappeared?
Setting this to 6000ms would make it disappear on its own. The user can swipe it to the right if he wants to get rid of it earlier.

elgatovital: let's do this as a separate task https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-android/1107

app/src/main/java/io/gnosis/safe/ErrorHandler.kt
```diff
@@ -0,0 +1,106 @@
...
+    ERROR_401(401, R.string.error_network_not_authorized_reason, R.string.error_network_not_authorized_fix),
+    ERROR_403(403, R.string.error_network_not_authorized_reason, R.string.error_network_not_authorized_fix),
+    ERROR_500(500, R.string.error_network_server_side_reason, R.string.error_network_server_side_fix),
+
+    // Common client errors
+    ERROR_1101(1101, R.string.error_client_safe_duplicate_reason, R.string.error_client_safe_duplicate_fix),
+    ERROR_1102(1102, R.string.error_client_address_invalid_reason, R.string.error_client_address_invalid_fix),
+    ERROR_1103(1103, R.string.error_client_seed_phrase_reason, R.string.error_client_seed_phrase_fix),
+    ERROR_1104(1104, R.string.error_client_tx_hash_validation_reason, R.string.error_client_tx_hash_validation_fix),
+    ERROR_1105(1105, R.string.error_client_unsupported_implementation_copy_reason, R.string.error_client_unsupported_implementation_copy_fix),
```
biafra23, line 49: This is not used. Do we need it later, or should it be used already?

biafra23 (approved):
👍 
I really like the unification of al errors is one class and also having custom exception for everything. 
I would increase the time an error is shown, but that's not a blocker. Liliya will tell us if she doesn't approve :-)

app/src/main/java/io/gnosis/safe/ErrorHandler.kt
elgatovital: we might want to make safe validation more strict and allow adding safes that have min required version of the implementation copy.

elgatovital: I'll add this to the next refinement. We have to adjust our safe validation

Issue #1119, opened 2020-12-02
[Android]Wrong error message if the safe with old master copy is added


liliya-soroka:
**Describe the bug**
Wrong error message if the safe with old master copy is added. Safe = 0xdb5dfe548e79441e709edd46f21d80cbd245b1be

**To Reproduce**
1. Go to add safe flow
2. Try to pass the safe address =0xdb5dfe548e79441e709edd46f21d80cbd245b1be

**Expected behavior**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/41357fcc-65b5-42b8-b86e-b5a73645bc08)

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/db0c9b30-00c1-4004-a495-353f368beb3a)

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.5.0 -1595

**Environment**
staging, rinkeby
**Additional context**

liliya-soroka:
Reopened:
wrong error for the next safes with mastercopy = 0.1.0
0x743ff9422189544370b9df561aa5901e91d16c90
0xa4487e1abc196e80f35d0ce8639e97bcd96f3ea9
Current result:
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/58ff1ee2-48d7-408c-a966-01f013df5645)
mainnet, prod
app version 2.7.0 -1634rc

Issue #1118, opened 2020-12-02
[Android]Wrong error message for invalid seed phrase 


liliya-soroka:
**Describe the bug**
Wrong error message for invalid seed phrase

**To Reproduce**
1. Go to the safe settings 
2. Enter invalid seed phrase

**Expected behavior**
Can’t use this seed phrase: this is not a valid seed phrase for an Ethereum account. Please correct the error or use another seed phrase.

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/47bebe69-294c-459b-9cdb-86a06414c71d)

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.5.0 -1595

**Environment**
staging, rinkeby
**Additional context**

liliya-soroka:
Verified
app version 2.7.0-1634rc

Issue #1117, opened 2020-12-02
[Android] Wrong error in case when failed confirmation because of the no internet connection 


liliya-soroka:
**Describe the bug**
Wrong error in case when failed confirmation because of the no internet connection

**To Reproduce**
1. Import seed phrase
2. create tx that requires confirmation from the imported owner
3. turn the internet connection off
4. click confirm

**Expected behavior**
Failed to confirm transaction: device is not connected to the Internet. Please try again when Internet is available. (Error 101)

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/896fd8ae-6ab5-4aac-bc4c-81db5926aca0)

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.5.0 -1595

**Environment**
staging, rinkeby
**Additional context**

liliya-soroka:
Verified
app version 2.7.0 -1634rc

Issue #1116, opened 2020-12-02
[Android] Wrong error message for the wrong qr code scanning


liliya-soroka:
**Describe the bug**
Incorrect QR Code scanning  is displayed

**To Reproduce**
1. Go to add safe
2. Scan non-safe qr code

**Expected behavior**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/2cb8d49e-3a6a-4b22-b784-c32f7ab4784f)

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/c75e99cd-5d3e-4993-bdf7-c5698f095416)

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.5.0 -1595

**Environment**
staging, rinkeby
**Additional context**

liliya-soroka:
Verified
app version 2.7.0 -1634rc

PR #1121, opened 2020-12-02
Handle safe not found error; check implementation copy before adding safe; minor layout fixes

elgatovital:
Handles #1109 
Handles #1110 
Handles #1116
Handles #1117
Handles #1118
Handles #1119

@gnosis/mobile-devs

### DIFF
--- a/data/src/main/java/io/gnosis/data/repositories/SafeRepository.kt
+++ b/data/src/main/java/io/gnosis/data/repositories/SafeRepository.kt
@@ -45,10 +45,6 @@ class SafeRepository(
 
     suspend fun removeSafe(safe: Safe) = safeDao.delete(safe)
 
-    suspend fun isValidSafe(safeAddress: Solidity.Address): Boolean {
-        return transactionServiceApi.getSafeInfo(safeAddress.asEthereumAddressChecksumString()) != null
-    }
-
     suspend fun clearActiveSafe() {
         preferencesManager.prefs.edit {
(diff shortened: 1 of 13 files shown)


data/src/main/java/io/gnosis/data/repositories/SafeRepository.kt
```diff
@@ -81,6 +72,27 @@ class SafeRepository(
 
     suspend fun saveSafeMeta(safeMeta: SafeMetaData) = safeDao.saveMeta(safeMeta)
 
+    suspend fun getSafeStatus(safeAddress: Solidity.Address): SafeStatus {
+
+        val safeInfo = transactionServiceApi.getSafeInfo(safeAddress.asEthereumAddressChecksumString())
+
+        val supportedContracts = setOf(
```
biafra23, line 79: Just to clarify: We don't support earlier contracts anymore?

What about 1.2.0? Do we need changes on the app to support it? Could we allow this now, so we don't have to release a version when it is used?

biafra23 (approved):
Looks good and fixes everything Liliya mentioned 👍

data/src/main/java/io/gnosis/data/repositories/SafeRepository.kt
elgatovital: we will need to add 1.2.0 eth address at some point. Implementation copy values are not incrementally increasing, so we need to list actual values of the supported contracts

Issue #1107, opened 2020-11-26
Check if snackbar display duration has to be adjusted due to increased length of error messages


elgatovital:
Users might not be able to read an error message with the default snackbar display duration. 
Increase duration to 6 seconds

- Device:
- Android OS version:
- Safe app version:

# Notes
If it takes more than 1 hour than please not put it to the release

liliya-soroka:
Verified
app version 2.7.0 - 1655rc

PR #1126, opened 2020-12-09
Use custom snackbar builder with increased display duration for errors

elgatovital:
Handles #1107

@gnosis/mobile-devs


biafra23 (approved):
👍

Issue #1112, opened 2020-12-02
Unified tx list: Load data


elgatovital:
- [x] Load queued and next tx's using client gw endpoint for queued tx's
- [x] Load history tx's using client gw endpoint for history tx's
- [x] Support pagination
- [x] Remove queued & history header insertion (insertSeparators in tx list view model)
- [x] Add new tx item types 
  - [x] separator header with text label (for next tx and for queued tx's list, for history tx's grouping)
  - [x] item type for conflict header 
    - [x] nonce
- [x] tx object's should be extended with conflictType field (values e.g. None, WithNext, WithoutNext)

Endpoints:
- https://safe-client.rinkeby.staging.gnosisdev.com/v1/safes/0x1230B3d59858296A31053C1b8562Ecf89A2f888b/transactions/history/
- https://safe-client.rinkeby.staging.gnosisdev.com/v1/safes/0x1230B3d59858296A31053C1b8562Ecf89A2f888b/transactions/queued/

biafra23:
@liliya-soroka 
Notes for QA. It doesn't make sense to QA this part. I strongly recommend only testing this, when all tickets related to Unified TX list are done.

liliya-soroka:
Loading issue on the transaction tab and queue tab
Reopened:
2.7.0 -1729
https://drive.google.com/file/d/1CGuoUgOCA-wA17c5vWJlBb8S46ZmYJhd/view?usp=sharing

liliya-soroka:
Verified
The separate issue for the data loading in case of imported owner key was created - https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-android/1160
app version 2.7.0 - 1751

Issue #1124, opened 2020-12-07
Don't rely on reflection for property names for JSON parsing


rmeissner:
Currently we rely on reflection to determine property names of json responses. This is error prone as refactorings (e.g. field name changes) or usage of proguard would break this.

- Annotate all classes that are used by moshi with `Json` and set a name
  - e.g. all classes in https://github.com/gnosis/safe-android/tree/master/data/src/main/java/io/gnosis/data/models/transaction
- Don't use reflection to read field names
  - e.g. https://github.com/gnosis/safe-android/blob/master/data/src/main/java/io/gnosis/data/adapters/MoshiBuilder.kt#L10

liliya-soroka:
closed , the full regression will be done before the 2.8.0 release

PR #1133, opened 2020-12-11
Unified tx list load data

biafra23:
Handles #1112

- Add Unified TX endoints for queue and history
- add UnifiedEntry
- adjusts tests
- Adding timezone offset to query

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/transactions/TransactionView.kt
```diff
@@ -154,7 +154,7 @@ sealed class TransactionView(
         val transactionHash: String
     )
 
-    data class SectionHeader(@StringRes val title: Int, override val id: String = "<unused>") : TransactionView(TransactionStatus.PENDING, id)
+    data class SectionHeader(val title: String, override val id: String = "<unused>") : TransactionView(TransactionStatus.PENDING, id)
```
elgatovital, line 157: we should use localization on the client-side. I suggest ignoring the label field and continue using local string resources. You could communicate to client-gw that the label ist not needed.

data/src/main/java/io/gnosis/data/models/transaction/UnifiedEntry.kt
```diff
@@ -0,0 +1,46 @@
+package io.gnosis.data.models.transaction
+
+import com.squareup.moshi.Json
+import java.util.*
+import io.gnosis.data.models.transaction.Transaction as TransactionSummary
+
+enum class UnifiedEntryType {
```
elgatovital, line 7: UnifiedEntryType sounds cool, but it's not specific enough. I think TxListEntryType would be more specific. Unified is a very generic term. Lots of things could be unified and we don't know what might be unified in the future.

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
```diff
@@ -15,11 +18,15 @@ class TransactionRepository(
     private val gatewayApi: GatewayApi
 ) {
 
-    suspend fun getTransactions(safeAddress: Solidity.Address): Page<Transaction> =
-        gatewayApi.loadTransactions(safeAddress.asEthereumAddressChecksumString())
+    suspend fun getTransactions(safeAddress: Solidity.Address): Page<UnifiedEntry> {
```
elgatovital, line 21: We should have separate calls for getting queued and history transactions.

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionView.kt
biafra23: The date (long) needs to be localized and "timezoned". But is that part of this PR of data loading? SectionHeader should probably receive a date instance and not a string here.

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
biafra23: Sure. But should that be part of this PR?

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionView.kt
jpalvarezl: Isn't `SectionHeader` for `Next` and `Queued`? Either way, maybe it would be easier to store the `@StringRes Int`. As for dates, I agree with Dirk as in, either providing the `viewHolder` with the date formatter or format the date in the `viewModel` where the list is created. I would recommend having an additional type here `DateHeader` or something like that, then you don't have to coerce everything into a `String`

data/src/main/java/io/gnosis/data/adapters/MoshiBuilder.kt
```diff
@@ -36,6 +36,14 @@ internal val settingsInfoAdapter =
         .withSubtype(SettingsInfo.EnableModule::class.java, SettingsInfoType.ENABLE_MODULE.name)
         .withSubtype(SettingsInfo.DisableModule::class.java, SettingsInfoType.DISABLE_MODULE.name)
 
+internal val unifiedEntryAdapter =
+    PolymorphicJsonAdapterFactory.of(UnifiedEntry::class.java, "type")
+        .withSubtype(UnifiedEntry.Transaction::class.java, UnifiedEntryType.TRANSACTION.name)
+        .withSubtype(UnifiedEntry.DateLabel::class.java, UnifiedEntryType.DATE_LABEL.name)
+        .withSubtype(UnifiedEntry.Label::class.java, UnifiedEntryType.LABEL.name)
+        .withSubtype(UnifiedEntry.ConflictHeader::class.java, UnifiedEntryType.CONFLICT_HEADER.name)
```
jpalvarezl, line 44: Should we still use `UnifiedEntryType.EnumVariant.name`? would that survive a proguard code obfuscation?

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
```diff
@@ -15,11 +18,15 @@ class TransactionRepository(
...
+        val queued = gatewayApi.loadTransactionsQueue(safeAddress.asEthereumAddressChecksumString())
+        val history = gatewayApi.loadTransactionsHistory(safeAddress.asEthereumAddressChecksumString())
 
-    suspend fun loadTransactionsPage(pageLink: String): Page<Transaction> =
-        gatewayApi.loadTransactionsPage(pageLink)
+        return Page((queued.count ?: 0) + (history.count ?: 0), history.next, history.previous, queued.results + history.results)
```
jpalvarezl, line 25: Why are you merging both lists? I think we should have 2 separate calls for each endpoint.

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionView.kt
biafra23: Labels are supposed to tell us what the `next` tx is. Don't we need that information? We should not show these Lables to the user. But we need to know which is the next tx, don't we?

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
biafra23: Because the UI is not prt of this PR and I still wanted to see all txs.

data/src/main/java/io/gnosis/data/adapters/MoshiBuilder.kt
biafra23: This will be fixed when we work on #1124

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionView.kt
jpalvarezl: Yes we do. I wouldn't get rid of them. Actually they are part of the mocks AFAIK: https://projects.invisionapp.com/share/W4ZBLJ2SHMB#/screens

The `Label` coming from the backend is just a marker within the list for `next` and `queued` sections, only returned in the `/transactions/queued` endpoint. The clients are expected to handle localisation. 

That's why my suggestion was to use `@StringRes Int` . But I am not sure specifically for what this `TransactionView` subclass will be used for. The backend returns different types for `Label`, `DateLabel` and `ConflictHeader` . I wouldn't recommend using a single `TransactionView` subclass for all 3 types, but that's up to you.

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
jpalvarezl: Actually, if this PR is meant to provide the data layer, I would expect for all the methods in `TransactionRespository` be ready to use by the `viewModel`. 

If you need to see the aggregated result, then you could easily mix it in a unit test, as this is not a use case consumed by production code. Or if you just need it in the UI, then maybe the `viewModel` is a better place to mix this up.

jpalvarezl: I think this is related to Vitaly's comment here: https://github.com/gnosis/safe-android/pull/1133#discussion_r542156672

To which my answer would be "yes, this should be part of the PR, IMO"

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
biafra23: Done in 1a1106d143e60c2c652dd3b48988ec1bcba1e75b

data/src/main/java/io/gnosis/data/models/transaction/TxListEntry.kt
```diff
@@ -0,0 +1,46 @@
+package io.gnosis.data.models.transaction
+
+import com.squareup.moshi.Json
+import java.util.*
+import io.gnosis.data.models.transaction.Transaction as TransactionSummary
+
+enum class TxListEntryType {
+    @Json(name = "TRANSACTION")  TRANSACTION,
```
elgatovital, line 8: why so many spaces btw annotation and enum value?

elgatovital (approved):

Issue #1114, opened 2020-12-02
Unified tx list: history tab


elgatovital:
- [ ] Load data when opening tab (using client gw endpoint for history tx data)
- [x] Support pagination
- [x] Show empty placeholder when no history tx's are there
- [x] Use established error handling logic (display snackbar with error codes)
https://docs.google.com/document/d/1foQ9DSMgO2iN-6u4JsJIDJWB7mUHX-D8ApZWvx6oKAo/edit#
- [x] Load on pull to refresh
- [x] Use cell types for header for group separators 
- [x] Format & display date using device locale for headers
- [x] Modify tx cell to display time instead of date & time

liliya-soroka:
the task is verified 
app version 2.7.0 - 1751

PR #1137, opened 2020-12-15
Unified TX List: History tab

biafra23:
Handles #1114 

Design: https://zpl.io/aNQGYve 

- Use right endpoint for queued and history txs
- Show only time (wihout seconds) in items (not the date)
- Unify all settings changes into one layout

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListFragment.kt
```diff
@@ -109,7 +114,7 @@ class TransactionListFragment : BaseViewBindingFragment<FragmentTransactionListB
             dividerItemDecoration.setDrawable(AppCompatResources.getDrawable(requireContext(), R.drawable.divider)!!)
             addItemDecoration(dividerItemDecoration)
         }
-        binding.refresh.setOnRefreshListener { viewModel.load() }
+        binding.refresh.setOnRefreshListener { viewModel.load(type = type) }
```
elgatovital, line 117: here we also don't need to pass type to viewModel

biafra23: Then we need to store the Type in the viewModel when we loadHistoryAndSubscribeToSafeChanges() or loadQueueAndSubscribeToSafeChanges() are called initially.

biafra23: I implemented it in 459b0c14659548cda49e1b6202909898c471f2cd but I am not sre this is really better. Now we have to make sure that we never call load() before calling loadHistory... or loadQueued...

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewModel.kt
```diff
@@ -38,9 +32,19 @@ class TransactionListViewModel
     appDispatchers: AppDispatchers
 ) : BaseStateViewModel<TransactionsViewState>(appDispatchers) {
 
-    init {
+    private lateinit var listType: TransactionListFragment.Type
+
+    fun loadHistoryAndSubscribeToSafeChanges() {
```
elgatovital, line 37: you could call viewModel.load in fragment's state observer (is ActiveStateChanged) and use type flag of the fragment. In init block of the viewModel you just inform about safe change:
```kotlin
init {
        safeLaunch {
            safeRepository.activeSafeFlow().collect { 
 updateState { TransactionsViewState(isLoading = true, viewAction = iActiveSafeChanged(safe)  }
 }
        }
    }
```

app/src/main/java/io/gnosis/safe/ui/transactions/paging/TransactionPagingSource.kt
```diff
@@ -3,20 +3,23 @@ package io.gnosis.safe.ui.transactions.paging
...
 import io.gnosis.data.repositories.TransactionRepository
+import io.gnosis.safe.ui.transactions.TransactionListFragment
 import pm.gnosis.model.Solidity
 import timber.log.Timber
 
 class TransactionPagingSource(
     private val safe: Solidity.Address,
-    private val txRepo: TransactionRepository
+    private val txRepo: TransactionRepository,
+    private val type: TransactionListFragment.Type
```
elgatovital, line 13: this Type enum is used outside of fragment now. I propose to put it into TransactionPagingSource. This way it will read itself quite nicely: TransactionPagingSource.Type (HISTORY | QUEUE)

biafra23:
Maybe getting rid of reflection is more involved than we thought. Maybe it is enough to annotate everything for now to make it less brittle and migrate to codegen in a future PR?

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewModel.kt
biafra23: I'd prefer to have `load(type, safeChange)` over using 
```
if (type == TransactionPagingSource.Type.HISTORY) {
                viewModel.loadHistory()
            } else {
                viewModel.loadQueue()
            }
```

biafra23: See: fd361c4b17d3c7ce44d9407b2fcc07cdffdd3401

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewModel.kt
```diff
@@ -38,21 +33,23 @@ class TransactionListViewModel
...
 
-    override fun initialState(): TransactionsViewState = TransactionsViewState(null, true)
-
-    fun load(safeChange: Boolean = false) {
+    fun load(type: TransactionPagingSource.Type, safeChange: Boolean = false) {
         safeLaunch {
             val safe = safeRepository.getActiveSafe()
             updateState { TransactionsViewState(isLoading = true, viewAction = if (safeChange) ActiveSafeChanged(safe) else ViewAction.None) }
             if (safe != null) {
                 val owner = ownerCredentialsRepository.retrieveCredentials()?.address
```
elgatovital, line 51: we don't have to check owner for history tx's

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewModel.kt
elgatovital: looks good

elgatovital (approved):

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewModel.kt
biafra23: IMHO the performance gain (one call of .contains() in a very short list) is not worth the code duplication. Let me know if you disagree, then I'll add it to the next PR.

PR #1138, opened 2020-12-17
Don't rely on reflection for property names for json parsing

elgatovital:
Handles #1124

@gnosis/mobile-devs

### DIFF
--- a/data/src/main/java/io/gnosis/data/adapters/MoshiBuilder.kt
+++ b/data/src/main/java/io/gnosis/data/adapters/MoshiBuilder.kt
@@ -8,40 +8,40 @@ import pm.gnosis.common.adapters.moshi.*
 
 internal val transferInfoAdapter =
     PolymorphicJsonAdapterFactory.of(TransferInfo::class.java, "type")
-        .withSubtype(TransferInfo.Erc20Transfer::class.java, TransferType.ERC20.name)
-        .withSubtype(TransferInfo.Erc721Transfer::class.java, TransferType.ERC721.name)
-        .withSubtype(TransferInfo.EtherTransfer::class.java, TransferType.ETHER.name)
+        .withSubtype(TransferInfo.Erc20Transfer::class.java, "ERC20")
+        .withSubtype(TransferInfo.Erc721Transfer::class.java, "ERC721")
+        .withSubtype(TransferInfo.EtherTransfer::class.java, "ETHER")
 
 internal val transactionInfoAdapter =
     PolymorphicJsonAdapterFactory.of(TransactionInfo::class.java, "type")
-        .withSubtype(TransactionInfo.Transfer::class.java, TransactionType.Transfer.name)
-        .withSubtype(TransactionInfo.SettingsChange::class.java, TransactionType.SettingsChange.name)
-        .withSubtype(TransactionInfo.Custom::class.java, TransactionType.Custom.name)
-        .withSubtype(TransactionInfo.Creation::class.java, TransactionType.Creation.name)
+        .withSubtype(TransactionInfo.Transfer::class.java, "Transfer")
+        .withSubtype(TransactionInfo.SettingsChange::class.java, "SettingsChange")
+        .withSubtype(TransactionInfo.Custom::class.java, "Custom")
+        .withSubtype(TransactionInfo.Creation::class.java, "Creation")
         .withDefaultValue(TransactionInfo.Unknown)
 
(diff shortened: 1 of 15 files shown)


data/src/main/java/io/gnosis/data/models/transaction/TransferInfo.kt
```diff
@@ -1,33 +1,50 @@
 package io.gnosis.data.models.transaction
 
+import com.squareup.moshi.Json
 import io.gnosis.data.repositories.TokenRepository
 import pm.gnosis.model.Solidity
 import java.math.BigInteger
 
 enum class TransferType {
-    ERC20, ERC721, ETHER
+    ERC20,
```
biafra23, line 9: Should we have a \@Json annotation for the enum values too?

data/src/main/java/io/gnosis/data/models/transaction/SettingsInfo.kt
```diff
@@ -1,19 +1,55 @@
 package io.gnosis.data.models.transaction
 
+import com.squareup.moshi.Json
 import pm.gnosis.model.Solidity
 
 enum class SettingsInfoType {
-    SET_FALLBACK_HANDLER, ADD_OWNER, REMOVE_OWNER, SWAP_OWNER, CHANGE_THRESHOLD, CHANGE_IMPLEMENTATION, ENABLE_MODULE, DISABLE_MODULE
+    SET_FALLBACK_HANDLER,
```
biafra23, line 7: Should we have a \@Json annotation for the enum values too?

biafra23:
Is it possible to change our formatting config so that all annotations can be on the same line as the type definition and are not wrapped. It looks as if they are wrapped even if lines aren't too long. 

Not a blocker. Just a question :-)

biafra23:
Don't we need Annotations for the io.gnosis.data.models.transaction.Param class and its sub classes?

biafra23:
When we got rid of all reflection based parsing we should be able to remove KotlinJsonAdapterFactory from the adapters list, right?

data/src/main/java/io/gnosis/data/models/transaction/TransferInfo.kt
elgatovital: These types are defined by client gw. So I would just reuse them. Defining annotations for these enums will make it less readable and we will have 2 sets of strings to maintain: one here and another one in MoshiBuilder

biafra23:
Maybe this can help us get rid of reflection: https://github.com/ZacSweers/MoshiX

biafra23:
Maybe getting rid of reflection is more involved than we thought. Maybe it is enough to annotate everything for now to make it less brittle and migrate to codegen in a future PR?

data/src/main/java/io/gnosis/data/models/transaction/TransferInfo.kt
biafra23: Then why do we annotate anything? All types and names are defined by the client gateway. 

As I understand it we annotate, so:
1. We can refactor everything without breaking the parsing 
2. To use moshi codegen instead of reflection

elgatovital: convincing argument

biafra23 (approved):
I guess, we're not trying to get rid of reflection completely. So 👍

data/src/main/java/io/gnosis/data/adapters/MoshiBuilder.kt
```diff
@@ -61,5 +60,4 @@ val dataMoshi =
         .add(transactionExecutionDetailsAdapter)
         .add(txListEntryAdapter)
         .add(ParamAdapter())
-        .add(KotlinJsonAdapterFactory())
```
biafra23, line 64: 🚀

data/src/test/java/io/gnosis/data/repositories/TokenRepositoryTest.kt
```diff
@@ -28,7 +29,7 @@ class TokenRepositoryTest {
     private val gatewayApi = mockk<GatewayApi>()
     private val tokenRepository = TokenRepository(gatewayApi)
 
-    private val moshi = dataMoshi
+    private val moshi = dataMoshi.newBuilder().add(KotlinJsonAdapterFactory()).build()
```
biafra23, line 32: The KotlinJsonAdapterFactory doesnt seem to be necessary to run the tests.

Issue #1115, opened 2020-12-02
Unified tx list: Time formatter for queued & next tx's


elgatovital:
Show time elapsed from tx creation for next & queued tx's
![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/bc0dd495-0c11-4d82-bc3d-b5c3d7c2ef56)

Example:
- Just now
- X seconds ago (if possible)
- N minutes ago
- N days ago
- M months ago
- Y years ago

liliya-soroka:
Safe liltestsafe.eth
< 1 min difference → just now - confirmed
< 1 hour difference → x min ago - confirmed
< 24 hours difference → 1 hour ago | x hours ago  - confirmed

24 hours difference → 1 day ago | x days ago - confirmed


verified
app version 2.7.0 - 1812

PR #1140, opened 2020-12-17
Time formatter for queued & next tx's

elgatovital:
Handles #1115

- Add elapsed time interval calculation and strings

@gnosis/mobile-devs


biafra23 (approved):
IMHO this could be integrated now. Used in the queued tx list.

Issue #1113, opened 2020-12-02
Unified tx list: queued tab


elgatovital:
- [x] Load data when opening tab (using client gw endpoint for queued tx data)
- [x] Support pagination
- [x] Show empty placeholder when no queued & next tx's are there
- [x] Use established error handling logic (display snackbar with error codes)
https://docs.google.com/document/d/1foQ9DSMgO2iN-6u4JsJIDJWB7mUHX-D8ApZWvx6oKAo/edit#
- [x] Load on pull to refresh
- [x] Add new cell types / view holders
  - [x] conflict header
  - [x] conflict tx with next
  - [x] conflict tx without next
  - [x] list header with text (separator) 

Adjusting time format for list items is handled here
https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-android/1115

elgatovital:
![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/55c8d41c-163f-4835-997b-7b76654890ac)

liliya-soroka:
One critical issue - https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-android/1160
the task is verified ( formating and loading issue with imported owner key is known)
app version 2.7.0 - 1751

PR #1149, opened 2020-12-17
Unified TX List: Queued tab

biafra23:
Handles #1113 

- Render queued txs correctly
- Render conflicts 
 
@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewModel.kt
```diff
@@ -226,7 +249,8 @@ class TransactionListViewModel
     private fun Transaction.queuedCustomTransaction(
         txInfo: TransactionInfo.Custom,
         safeAddress: Solidity.Address,
-        awaitingYourConfirmation: Boolean
+        awaitingYourConfirmation: Boolean,
+        isConflict: Boolean
```
jpalvarezl, line 253: Were the unit tests not affected by the addition of this flag?

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewHolderFactory.kt
```diff
@@ -300,13 +315,43 @@ class SectionDateHeaderViewHolder(private val viewBinding: ItemTxSectionHeaderBi
...
+    }
+
+    private fun TextView.appendLink(url: String, urlText: String) {
+        val textColor = ForegroundColorSpan(textColors.defaultColor) //TODO get foreground color from TextView
+
+        val linkDrawable = ContextCompat.getDrawable(context, R.drawable.ic_link_green_24dp)!!
+        linkDrawable.setBounds(0, 0, linkDrawable.intrinsicWidth, linkDrawable.intrinsicHeight)
+        movementMethod = LinkMovementMethod.getInstance()
+        text = SpannableStringBuilder()
+            .appendText(Html.fromHtml(text.toString()), textColor)
```
jpalvarezl, line 341: I am not sure what happens here, are you appending the spannable to itself as a string? Why is this necessary?

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewHolderFactory.kt
```diff
@@ -300,13 +315,43 @@ class SectionDateHeaderViewHolder(private val viewBinding: ItemTxSectionHeaderBi
...
+            nonce.text = sectionDateHeader.nonce.toString()
+            sectionTitle.text = resources.getString(R.string.tx_list_conflict_header_explainer)
+            sectionTitle.appendLink(
+                resources.getString(R.string.tx_list_conflict_header_link),
+                resources.getString(R.string.tx_list_conflict_header_learn_more)
+            )
+        }
+    }
+
+    private fun TextView.appendLink(url: String, urlText: String) {
```
jpalvarezl, line 334: This method looks like it could be separated into 2: `prepareLinkSpannable` and `appendLink`. 

I think it would be nice to extract these extension functions into a `StringUtils` file. The same would for `SpannableStringBuild.appendTextWithSpans`

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewHolderFactory.kt
```diff
@@ -300,13 +315,43 @@ class SectionDateHeaderViewHolder(private val viewBinding: ItemTxSectionHeaderBi
...
+        text = SpannableStringBuilder()
+            .appendText(Html.fromHtml(text.toString()), textColor)
+            .append(" ")
+            .appendText(urlText, URLSpan(url))
+            .append(" ")
+            .appendTextWithSpans(" ", listOf(ImageSpan(linkDrawable, ImageSpan.ALIGN_BASELINE), URLSpan(url)))
+    }
+
+    private fun SpannableStringBuilder.appendTextWithSpans(text: CharSequence, what: List<Any>, flags: Int = 0): SpannableStringBuilder {
+        val start = length
```
jpalvarezl, line 349: Wouldn't you have to update `start` for each item in `what`?

jpalvarezl: I don't think I understand very well what this method does. Could you explain it to me?

biafra23: > Wouldn't you have to update start for each item in what?

No. Because the whats are supposed to be on top of each other. In this case the Image should also be clickable and open the url. Other cases could be a list of underline and color change. Both need to be applied to the same range.

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewHolderFactory.kt
biafra23: Thank you for pointing this out!  🚀  
No it's not necessary. In fact it is harmful. I changed it to use `TextView.append()` instead. And now several links can be appended without problems. And now I can make it public.

app/src/main/res/values/strings.xml
```diff
@@ -160,8 +160,8 @@
     <string name="tx_list_creation">Safe created</string>
     <string name="tx_list_receive">Receive</string>
     <string name="tx_list_send">Send</string>
-    <string name="tx_list_label_type_next">Next</string>
-    <string name="tx_list_label_type_queued">Queued</string>
+    <string name="tx_list_label_type_next">NEXT TRANSACTION</string>
```
jpalvarezl, line 163: There could be a conflict section under `NEXT` I think 🤔 , if there are many transactions proposed with the current safe nonce. I think that is beyond the scope of this PR so product can look at the app string values with real life data and decide.

My point being, is that there may be more than 1 transaction here, so it would be `Next transactions` maybe? Anyhow, not something we should look into right now.

biafra23: I think the reasoning for singular is that only one of these will be executed?

I will clarify in slack with Tobias

jpalvarezl: Just saw that, good call 👍

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewHolderFactory.kt
biafra23: See new version in: 2a1cf57848f392a2611397c6554133af37f44517

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListFragment.kt
```diff
@@ -106,9 +106,17 @@ class TransactionListFragment : BaseViewBindingFragment<FragmentTransactionListB
...
-            addItemDecoration(dividerItemDecoration)
+
+            addItemDecoration(object : RecyclerView.ItemDecoration() {
+                override fun getItemOffsets(outRect: Rect, view: View, parent: RecyclerView, state: RecyclerView.State) {
+                    super.getItemOffsets(outRect, view, parent, state)
+                    when (val viewHolder = nullOnThrow { parent.getChildViewHolder(view) }) {
+                        is ConflictViewHolder -> if (viewHolder.hasNext()) return
+                        is SectionConflictHeaderViewHolder -> return
+                    }
+                    outRect[0, 0, 0] = context.resources.getDimension(R.dimen.item_separator_height).toInt()
```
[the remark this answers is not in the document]
biafra23: Done. See: f23981deaec3edba616b1cb5e9c4ccf4d8e93bdf

jpalvarezl (approved):
Looking good!

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewModel.kt
biafra23: I added a check for empty nonces in conflicting transfers. See: 40368ea242a1838cae95c03aaf07579a479e8586

app/src/test/java/io/gnosis/safe/ui/transactions/TransactionListViewModelTest.kt
```diff
@@ -171,24 +173,57 @@ class TransactionListViewModelTest {
...
+                transaction = transaction,
+                activeSafe = defaultSafeAddress,
+                awaitingYourConfirmation = false,
+                isConflict = true
+            )
+        }
+
+        assertEquals("", (transactionViews[0] as TransactionView.TransferQueued).nonce)
+        assertEquals("", (transactionViews[1] as TransactionView.TransferQueued).nonce)
+        assertEquals("", (transactionViews[2] as TransactionView.TransferQueued).nonce)
```
jpalvarezl, line 219: Nice tests

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewHolderFactory.kt
```diff
@@ -327,3 +342,23 @@ class SectionLabelHeaderViewHolder(private val viewBinding: ItemTxSectionHeaderB
     }
 }
 
+class ConflictViewHolder(private val viewBinding: ItemTxConflictTxBinding, private val factory: TransactionViewHolderFactory) :
+    BaseTransactionViewHolder<TransactionView.Conflict>(viewBinding) {
+
+    private lateinit var conflictView: TransactionView.Conflict
```
jpalvarezl, line 348: Having this internal state here is a bit risky, in case the viewHolder gets recycled on a quick scroll and the method `hasNext()` gets called before this is updated to the correct value. I think it is an edge case, but it could be a source of UI problems.

biafra23: I made a separate ticket to investigate and try to fix here: https://github.com/gnosis/safe-android/issues/1159

jpalvarezl: Sounds good!

Issue #1157, opened 2020-12-22
[Android] Load safe: Don't display text on enter safe address when the address is entered


liliya-soroka:
Bug description
According to design the info about not safe should not be displayed on the safe address page if the address is entered
Steps To Reproduce
1. Go to the load safe page
2. Past address from clipboard/scan/enter ens

Expected Result
https://app.zeplin.io/project/5e8f27b9963e55bb79b13634/screen/5e8f2d59228c3621ade1db30
Screenshots
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/4ab0c248-2d58-4ec6-b3db-7a99df524507)
Device & App version

Device/OS: Huawei. P30
App version: 2.8.0 -1745
Environment
staging, rinkeby
Additional info
(Optional)

liliya-soroka:
Verified
app version 2.7.0 -1862

PR #1173, opened 2021-01-07
Hide "create safe" on error

biafra23:
Handles #1157 

Add Safe | Error
--------- | -----
Screenshot 2021-01-07 at 15 43 25 | Screenshot 2021-01-07 at 15 43 53

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/safe/add/AddSafeFragment.kt
```diff
@@ -72,6 +73,7 @@ class AddSafeFragment : BaseViewBindingFragment<FragmentAddSafeBinding>() {
         with(binding) {
             progress.visible(false)
             nextButton.isEnabled = false
+            bottomLabels.isVisible = false
```
elgatovital, line 76: better to be consistent and use either isVisible or .visible(false) everywhere. We already use .visible(boolean) here, so I would continue using this extension

elgatovital (approved):

Issue #1147, opened 2020-12-17
[Android] The default icon for tokens is blinked ( from dark to light) if the device has dark theme and the app has light mode


liliya-soroka:
**Describe the bug**
The default icon for tokens is blinked ( from dark to light) if the device has dark theme and the app has light mode

**To Reproduce**
1. Set up Dark theme on the device
2. Make sure that the light mode is set up in the app settings
3. Open assets in the liltestsafe.eth or tx details where the default icon is used for token
Current result: Dark default icon is displayed for a few sec - ![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/1158b858-5651-4f5f-a35d-508c68e40f86)

**Expected behavior**
The correct icon should be displayed even if the opposite theme is selected in the device and in the app

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.7.0 -1686

**Environment**
staging, rinkeby
**Additional context**

liliya-soroka:
Verified
app version 2.7.0 (1862)

PR #1187, opened 2021-01-12
Fix coin icons placeholder

elgatovital:
Handles #1147

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/utils/ImageUtils.kt
+++ b/app/src/main/java/io/gnosis/safe/utils/ImageUtils.kt
@@ -15,25 +15,32 @@ fun ImageView.loadTokenLogo(
     @DrawableRes backgroundDrawable: Int? = R.drawable.circle
 ) {
     setPadding(0)
+    setImageDrawable(null)
+    colorFilter = null
     background = backgroundDrawable?.let {
         ContextCompat.getDrawable(context, backgroundDrawable)
     }
-    setImageDrawable(null)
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/gnosis/safe/utils/ImageUtils.kt
```diff
@@ -15,25 +15,32 @@ fun ImageView.loadTokenLogo(
...
-        icon?.startsWith("local::") == true -> setImageResource(placeHolderResource)
-
-        !icon.isNullOrBlank() ->
+        icon?.startsWith("local::") == true -> {
+            setImageResource(placeHolderResource)
+        }
+        !icon.isNullOrBlank() -> {
+            //FIXME: workaround, as Picasso uses a final ApplicationContext for its #getDrawable() action, which will not listen to theme's attributes.
+            // this would lead to day/night mode not handled correctly
+            val placeholderDrawable = ContextCompat.getDrawable(context, placeHolderResource)!!
```
jpalvarezl, line 33: Is this really a workaround? Seems like you are loading the resource correctly

jpalvarezl (approved):

app/src/main/java/io/gnosis/safe/utils/ImageUtils.kt
elgatovital: it would be nice to just pass resource id's directly to picasso, without loading them first

Issue #1141, opened 2020-12-17
[Android] White frame for the default icon present in the dark mode but not on the design 


liliya-soroka:
**Describe the bug**
White frame for the default icon present in the dark mode but not on the design 

**To Reproduce**
1. Go to assets page with the tokens without logo in the dark mode( default logo)

**Expected behavior**
Design- ![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/01fe62cc-4a32-4dc5-8168-49e783ae81c3)

**Screenshots**
In the app in dark mode - ![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/7c264422-7a99-4d1d-b9dc-cf6a5e37f733)

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.7.0 -1686

**Environment**
staging, rinkeby
**Additional context**

liliya-soroka:
Verified
app version 2.7.0 -1862

PR #1189, opened 2021-01-12
Use background with stroke

biafra23:
Handles #1141

- Using stroke to prevent background shining through on the edge
- Replacing custom `CircleTransformation` with `ShapeableImageView`

Before | After
-------|------
image| image
image | image

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/utils/ImageUtils.kt
```diff
@@ -34,42 +32,10 @@ fun ImageView.loadTokenLogo(
             Picasso.get()
                 .load(icon)
                 .placeholder(placeholderDrawable)
-                .error(placeholderDrawable)
```
elgatovital, line 37: what happens in case token logo can't be loaded?

app/src/main/res/layout/item_coin_balance.xml
```diff
@@ -9,15 +9,28 @@
     android:paddingVertical="16dp">
 
     <ImageView
+        android:id="@+id/logo_image_background"
```
elgatovital, line 12: can't we just set background for logo_image?

app/src/main/res/drawable/ic_coin_placeholder_background.xml
```diff
@@ -0,0 +1,13 @@
+<vector
```
elgatovital, line 1: why do we need this vector drawable just for the background?

app/src/main/java/io/gnosis/safe/utils/ImageUtils.kt
biafra23: The asset set with placehoder() is shown.

app/src/main/res/layout/item_coin_balance.xml
biafra23: That is exactly the problem with this ticket. When using background I found no way o get rid of the corona. Maybe you have an idea how to do this with a background?

biafra23: You can find an example with background  in the details screen.

app/src/main/java/io/gnosis/safe/utils/ImageUtils.kt
jpalvarezl: Do we get rid of the `error` asset? What happens if the URL in `icon` does not resolve into something Picasso can handle?

app/src/main/res/drawable/ic_coin_placeholder_background.xml
biafra23: We probably don't need it. I was using it instead of `circle`, because I thought  the explicit dimenions would allow me to scale it. But I was wrong.

app/src/main/res/layout/item_coin_balance.xml
```diff
@@ -9,15 +9,28 @@
...
+        android:layout_width="28dp"
+        android:layout_height="28dp"
+        android:layout_marginStart="16dp"
+        android:padding="0.001dp"
```
jpalvarezl, line 16: wow didn't know you could use decimals!

jpalvarezl: what is this padding for?

app/src/main/res/drawable/ic_coin_placeholder_background.xml
biafra23: Removed in 7de3ee455b7f277c960a626c830c3d7a1553afe2

app/src/main/res/layout/item_coin_balance.xml
biafra23: image

jpalvarezl: Have you tried: https://developer.android.com/reference/android/view/View#setClipToOutline(boolean) ?

jpalvarezl (approved):

Issue #1203, opened 2021-01-18
[Android] Only time is displayed on Safe created transaction details


liliya-soroka:
**Describe the bug**
Only time is displayed on Safe created transaction details as creation date instead of date+time

**To Reproduce**
1. Go to any safe 
2. Go to the History and scroll down to the first transaction - Safe created
3. Open transaction details and check creation date
Current result: only time is displayed without date

**Expected behavior**
date +time should be displayed as creation date for the Safe created as for another transactions

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/6c538820-dded-4952-b0c5-947a43abdaf5)If applicable, add screenshots to help explain your problem.

**Smartphone (please complete the following information):**
 - Device: Pixel P30
 - Android Version: 10
 - App Version 2.8.0 -1882rc

**Environment**
prod , mainnet
**Additional context**

liliya-soroka:
Verified
app version 2.9.0 - 1967rc

Issue #1191, opened 2021-01-13
[Android] Signing V2: Import owner key


biafra23:
### Expected behavior

- Text input fields allow entering secret key a s hex value now 
- Description changed (see design)
- If the user enters a pass phrase
  - Only one owner address is shown 
  - A button labeled "Show more derived keys" 
    - If the user presses that button, more addresses are shown
- If the user enters a private key
  - Only one owner address  is shown
  - Use KeyPair.fromPrivate().address to get the owner address on Android 

## Tracking 
(see: https://github.com/gnosis/safe/wiki/Tracking)
- New param for `user_key_imported` event
    - `import_type` (String): `seed` or `key`

Desgin:

Enter Key or Seed: https://zpl.io/amo0Md6
![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/4c8290bd-1f77-4ba6-82ae-db8c67f61621)

Owner Key found: https://zpl.io/anjAOqA
![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/49717485-aa41-456f-adc8-3165ac4eb81d)

Import owner - Default: https://zpl.io/adlW4OE
![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/4fcddd75-4a89-4aba-973c-36e6f156387a)

Import owner - Select: https://zpl.io/aBL8KKK
![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/d2957f23-7253-4665-b76a-4f4043c03059)

DmitryBespalov:
@biafra23 

>- Do we have an implementation to derive an address from a key (ios/android)?

Yes, it should be trivial: `Private Key -> Public Key -> Address`

liliya-soroka:
Reopened"
Update the text on the 1st page, please ,on the screen according to the latest design 

app version 2.9.0 -1967rc

liliya-soroka:
Verified
app version 2.9.0 -1991rc

The tracking will be checked separately

PR #1212, opened 2021-01-20
Import owner key

biafra23:
Handles #1191 and #1203

- Allows users to import secret key and pass phrase in one entry field (not at the same time)
- Owner address display changed for default key/address 
- Owner address display different for secret key import
- Fixes date in creation tx
- Correct error messages for malformed secret keys

https://user-images.githubusercontent.com/246473/105868698-bea4d880-5ff6-11eb-82cd-89b324910cdf.mp4

https://user-images.githubusercontent.com/246473/105869123-2eb35e80-5ff7-11eb-84de-726342a2c672.mp4

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/settings/owner/OwnerSeedPhraseViewModel.kt
```diff
@@ -12,7 +12,42 @@ class OwnerSeedPhraseViewModel
     appDispatchers: AppDispatchers
 ) : PublishViewModel<ImportOwnerKeyState>(appDispatchers) {
 
-    fun validate(seedPhrase: String) {
+    fun validate(seedPhraseOrKey: String) {
+        if (detectPrivateKey(seedPhraseOrKey)) {
```
elgatovital, line 16: if(isPrivateKey) would be better readable

app/src/main/java/io/gnosis/safe/ui/settings/owner/OwnerSeedPhraseViewModel.kt
```diff
@@ -12,7 +12,42 @@ class OwnerSeedPhraseViewModel
...
+
+        val pattern = "[0-9a-fA-F]{64}".toRegex()
+        pattern.find(input)?.let {
+            return true
+        }
+
+        return false
+    }
+
+    internal fun validatePrivateKey(key: String) {
```
elgatovital, line 34: there is no validation here except for whether it is a zero key. You should also check if it's a valid hex number. Something like this 0[xX][0-9a-fA-F]{64}

app/src/main/java/io/gnosis/safe/ui/settings/owner/list/OwnerSelectionFragment.kt
```diff
@@ -74,62 +76,98 @@ class OwnerSelectionFragment : BaseViewBindingFragment<FragmentOwnerSelectionBin
...
                 when (viewAction) {
-                    is LoadedOwners -> {
+                    is SingleOwner -> {
+                        lifecycleScope.launch {
+                            with(binding) {
+                                progress.visible(false)
+                                importButton.isEnabled = true
+
+                                if (viewAction.hasMore) {
+                                    binding.derivedOwners.visible(true)
```
elgatovital, line 100: binding. prefix can be removed in this section

app/src/main/java/io/gnosis/safe/ui/settings/owner/OwnerSeedPhraseViewModel.kt
biafra23: Ok. But then we do this check twice. But you are right. When we use this at some point in another part of the application a future versio of us might assume it really validates.

app/src/main/java/io/gnosis/safe/ui/settings/owner/OwnerSeedPhraseViewModel.kt
```diff
@@ -12,7 +12,42 @@ class OwnerSeedPhraseViewModel
...
+
+        val pattern = "[0-9a-fA-F]{64}".toRegex()
+        pattern.matchEntire(input)?.let {
+            return true
+        }
+
+        return false
+    }
+
+    internal fun validatePrivateKey(key: String) {
```
elgatovital, line 34: do we need to check if input is private key here or can we just assume it? validatePrivateKey should be called after isPrivateKey

app/src/main/java/io/gnosis/safe/ui/settings/owner/list/OwnerListAdapter.kt
```diff
@@ -51,52 +54,83 @@ class OwnerListAdapter() : PagingDataAdapter<Solidity.Address, RecyclerView.View
                 }
             }
         }
+        if (holder is DefaultOwnerViewHolder) {
```
elgatovital, line 57: can we just have 
```kotlin
 kotlin.runCatching {
                val uiModel = getItem(position)
                uiModel?.let {
                    holder.bind(it, position)
                }
            }
```
here and remove `if`. Both holders have same signature, so we shouldn't need this redundancy

app/src/main/java/io/gnosis/safe/ui/settings/owner/OwnerSeedPhraseViewModel.kt
```diff
@@ -12,7 +12,42 @@ class OwnerSeedPhraseViewModel
...
-    fun validate(seedPhrase: String) {
+    fun validate(seedPhraseOrKey: String) {
+        if (isPrivateKey(seedPhraseOrKey)) {
+            validatePrivateKey(seedPhraseOrKey)
+        } else {
+            validateSeedPhrase(seedPhraseOrKey)
+        }
+    }
+
+    internal fun isPrivateKey(seedPhraseOrKey: String): Boolean {
```
elgatovital, line 23: we could mark methods with relaxed visibility due to testing with 
@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)

elgatovital (changes requested):

elgatovital:
unit tests are failing

app/src/main/java/io/gnosis/safe/ui/settings/owner/OwnerSeedPhraseViewModel.kt
biafra23: I added this check so we could use validatePrivateKey in other contexts. But I can remove it. It is not necessary as we use it right now.

app/src/main/java/io/gnosis/safe/ui/settings/owner/list/OwnerListAdapter.kt
biafra23: The problem is that bind is only defined in these sub classes.

biafra23: Even this doesn't compile:
``` 
if (holder is OwnerViewHolder || holder is DefaultOwnerViewHolder) {
            kotlin.runCatching {
                val uiModel = getItem(position)
                uiModel?.let {
                    holder.bind(it, position)
                }
            }
        }
```

app/src/main/java/io/gnosis/safe/ui/settings/owner/OwnerSeedPhraseViewModel.kt
biafra23: It doesn't seem to break the build. But it highlights non private usage:
image

biafra23: Done in: 5225b2e8bc8ab0730e628a210c4435c101221dbd

app/src/main/java/io/gnosis/safe/ui/settings/owner/OwnerSeedPhraseViewModel.kt
biafra23: changed in: 9d9a67cfec986d158d5763a4d4392c302ec70e3a

elgatovital (approved):

Issue #1208, opened 2021-01-20
[Android] Tutorial (Signing V2: Import owner key)


biafra23:
### Expected behavior

- Additional interstitial screen explaining what it means to add a secret key
  - Next button leads to current import screen

Desgin:
Tutorial: https://zpl.io/VOlq5JQ
![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/fe470efe-e348-42af-a207-8088046ef707)

liliya-soroka:
test cases are added - https://gno.testrail.io/index.php?/suites/view/11&group_by=cases:section_id&group_id=995&group_order=asc

liliya-soroka:
@elgatovital , was the tracking implemented for this page?

liliya-soroka:
Verified
the tracking will be tested separately as soon as the app will be installed locally

liliya-soroka:
Tracking verified in 2.9.0 rc

Issue #1205, opened 2021-01-18
[Android] Signing V2: Discoverability: Banner


biafra23:
## Banner
- Display banner above assets 
   - **if** the following conditions are met:
     - For app update: No owner key currently installed && banner has never been dismissed && have a safe loaded
     - For new installs: No owner key was ever installed && banner has never been dismissed && have a safe loaded
  - **Import owner key**  We added signing support to the app! Now you can import your owner key and sign transactions on the go.
  - Buttons: 
    - Import owner key -> Onboarding
    - Cross in the upper left corner to dismiss banner
- Store whether the banner has been dismissed

## Tracking 
(see: https://github.com/gnosis/safe/wiki/Tracking)
- `user_banner_owner_skip` Triggered when clicking the x on the banner to ask users to import their owner key.
- `user_banner_owner_import` Triggered when clicking the Import on the banner to ask users to import their owner key.

## Design
Banner: https://zpl.io/b6XOZ8R

![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/76c9c64a-4a37-4a46-b3d9-8137a8760e1d)

**Updates**: 
When the user starts the import process but does not complete it, the banner should not be shown again (Dirk, 25.1.2021)

elgatovital:
@biafra23 does banner scroll with the list?

liliya-soroka:
Banner test cases are added - https://gno.testrail.io/index.php?/suites/view/11&group_by=cases:section_id&group_id=1583&group_order=asc

liliya-soroka:
Reopened :
The banner is dismissed if the use follows "Import owner key" link but an owner key is not added
Steps:
1. Install the app
2. Load safe
3. Follow "Import owner key" link
4. Cancel owner key importing
5. Come back to the balance page
Current result: No banner
Expected result: 
Only X button or adding owner key should dismiss the banner

App version 2.8.0 - 1920

elgatovital:
@liliya-soroka What is the source for this requirement? Clicking import link is same as dismissing the banner, except you start the import owner flow. @tschubotz, @posthnikova  what do you think?

liliya-soroka:
https://docs.google.com/document/d/1DqUNRCFpJQWS-HMM4gWzDN-ihNIA8-WoFT8ZfjPHJpw/edit#heading=h.eq8eoqliqxlt 
9. Allow to dismiss banner? How many times to show banner after it’s dismissed?
[...]

PR #1220, opened 2021-01-21
Display import owner banner above coin balances

elgatovital:
Handles #1205
Handles #1208

- Add banner
- React to banner actions
- Refactor coins adapter
- Save banner visibility flag in settings handler
- Track banner actions
- Add owner info screen
- Adjust nav graph
- Confirmation snackbar about imported owner key on AssetsFragment will be handled in a separate pr

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/Tracker.kt
```diff
@@ -61,6 +69,8 @@ class Tracker private constructor(context: Context) {
     object Event {
         val KEY_IMPORTED = "user_key_imported"
         val TRANSACTION_CONFIRMED = "user_transaction_confirmed"
+        val BANNER_OWNER_SKIPED = "user_banner_owner_skip"
```
biafra23, line 72: Typo, should be: BANNER_OWNER_SKIPPED

app/src/main/java/io/gnosis/safe/ui/assets/coins/CoinsViewModel.kt
```diff
@@ -45,17 +47,21 @@ class CoinsViewModel
                     )
                 }
                 val balanceInfo = tokenRepository.loadBalanceOf(safe.address, userDefaultFiat)
-                val balances = getBalanceViewData(balanceInfo)
-
-                updateState { CoinsState(loading = false, refreshing = false, viewAction = UpdateBalances(Adapter.Data(null, balances))) }
+                val showBanner = settingsHandler.showOwnerBanner && !ownerCredentialsRepository.hasCredentials()
```
biafra23, line 50: I think we need something separate to detect whether the user never imported (and possibly deleted) an owner. Since the button after adding a safe also depends on it, we could put it in that future PR.

biafra23: When using the SettingsHandler to keep track if an owner has been added in the past, the we don't need the ownerCredentialsRepository here.

elgatovital: after updating installed app with already imported owner key you won't be able to rely just on SettingsHandler and prefs. ownerCredentialsRepository has to be checked explicitly

biafra23 (approved):
LGTM 👍

PR #1232, opened 2021-01-28
Text and margin updated

biafra23:
Handles #1191
Text update according to design: https://zpl.io/VxNo8zW

image

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/settings/owner/OwnerSeedPhraseFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/settings/owner/OwnerSeedPhraseFragment.kt
@@ -96,7 +96,12 @@ class OwnerSeedPhraseFragment : BaseViewBindingFragment<FragmentOwnerSeedPhraseB
         val spannableString = SpannableString(resources.getString(stringResource));
         val matcher = Pattern.compile("\n\n").matcher(resources.getString(stringResource));
         while (matcher.find()) {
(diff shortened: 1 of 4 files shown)


elgatovital:
why don't we just use two textview's without doing some text manipulations at runtime?

app/src/main/java/io/gnosis/safe/ui/settings/owner/OwnerSeedPhraseFragment.kt
```diff
@@ -85,6 +92,15 @@ class OwnerSeedPhraseFragment : BaseViewBindingFragment<FragmentOwnerSeedPhraseB
         })
     }
 
+    private fun replaceDoubleNewlineWithParagraphLineSpacing(@StringRes stringResource: Int): SpannableString {
```
elgatovital, line 95: why don't we just use two TextView's without doing some text manipulations at runtime?

biafra23: Because this will work with future layouts when the text has paragraphs. "\n\n" can be replaced by a paragraph line spacing. From Sasha's answer I suspect that this will happen more often.

elgatovital: Ok, But the value in size span (8) should come from dimens

biafra23: done

elgatovital (approved):

Issue #1231, opened 2021-01-28
[Android] "Import owner key" title should be used for all pages in Import owner key flow 


liliya-soroka:
**Describe the bug**
According to design the title "Import owner key" title should be used for the whole import owner key flow 
Now two steps have title "Import Owner Key" and one step has the title "Import owner key 
**To Reproduce**
1. Go to app settings ->Import owner key 
2. Click Next on intro screen ->Enter seep phrase or private key 
Current result: Import Owner Key title is used instead of "Import Owner Key"

3. Go to the 3rd step of the flow 
Current result: "import owner key" title is used

**Expected behavior**
https://app.zeplin.io/project/5e9efc30726e106958cc43de/screen/6005cfe98f4724185f35ee8b

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.8.0 - 1962

**Environment**
staging, rinkeby
**Additional context**

liliya-soroka:
Verified
app version 2.9.0 -1991rc

PR #1233, opened 2021-01-29
Adjust toolbar title for import owner flow

elgatovital:
Handles #1231

@gnosis/mobile-devs


biafra23 (approved):

biafra23:
Do we need all four? 
image

elgatovital:
> Do we need all four?
> image

I would leave it for now and revisit it later. Those are different places in app and we use separate string resources for such case.

PR #1240, opened 2021-02-02
Add *Infos to transfers, setting changes and custom transactions

biafra23:
Handles #1152 

-
-
-

@gnosis/mobile-devs

### DIFF
--- a/data/src/main/java/io/gnosis/data/models/transaction/TransactionInfo.kt
+++ b/data/src/main/java/io/gnosis/data/models/transaction/TransactionInfo.kt
@@ -10,46 +10,35 @@ sealed class TransactionInfo(
 ) {
     @JsonClass(generateAdapter = true)
     data class Custom(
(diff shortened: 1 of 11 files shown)


data/src/main/java/io/gnosis/data/models/transaction/TransactionInfo.kt
```diff
@@ -58,8 +47,16 @@ sealed class TransactionInfo(
...
     INCOMING,
+
     @Json(name = "OUTGOING")
     OUTGOING,
+
     @Json(name = "UNKNOWN")
     UNKNOWN
 }
+
+@JsonClass(generateAdapter = true)
```
elgatovital, line 58: Will we use AddressInfo in other places like settings, advanced settings? If yes it would make sence to move it outside of the transaction-package and place it into data/models

elgatovital (approved):

Issue #1242, opened 2021-02-03
[Android]Crash: AddSafeOwnerFragment.java line 67


liliya-soroka:
**Describe the bug**
The app is crashed if  "Import owner key" is clicked during add safe when add safe flow was called from Transactions or Safe settings

**To Reproduce**
1. Install the app
2. Accept T&C
3. Go to the transactions list or safe settings
4. Add safe: copy & past safe address->enter safe name
5. Click "Import owner key" during onboarding 
**Current result** the app is crashed
https://console.firebase.google.com/u/0/project/safe-firebase-mainnet/crashlytics/app/android:io.gnosis.safe/issues/5653a43e39e391fa09028ce9207d7ff1?time=last-seven-days&sessionEventKey=601A89A20173000127234BA94C7F2B43_1503444454678598464

**Expected behavior**
The app should not be crashed and the Import owner key flow should be started

**Screenshots**

**Smartphone (please complete the following information):**
 - Device: [e.g. Pixel 2]
 - Android Version: [e.g. Android 9]
 - App Version [e.g. 1.0.1]

**Environment**

**Additional context**

liliya-soroka:
Verified
app version 2.9.0 -2006rc

PR #1244, opened 2021-02-03
Fix navigation to import owner flow

elgatovital:
Handles #1242

- Make owner import flow global action to allow navigation from different starting points (Assets, Tx's, Settings)

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/safe/empty/NoSafeFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/safe/empty/NoSafeFragment.kt
@@ -29,7 +29,7 @@ class NoSafeFragment : BaseViewBindingFragment<FragmentNoSafesBinding>() {
         super.onViewCreated(view, savedInstanceState)
 
         binding.loadSafeButton.setOnClickListener {
-            requireParentFragment().findNavController().navigate(R.id.action_to_add_safe_nav)
+            requireParentFragment().findNavController().navigate(R.id.action_to_add_safe)
         }
     }
 
--- a/app/src/main/res/layout/fragment_transactions.xml
+++ b/app/src/main/res/layout/fragment_transactions.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<androidx.constraintlayout.widget.ConstraintLayout
+<androidx.coordinatorlayout.widget.CoordinatorLayout
     xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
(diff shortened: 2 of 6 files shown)


app/src/main/java/io/gnosis/safe/ui/safe/empty/NoSafeFragment.kt
```diff
@@ -29,7 +29,7 @@ class NoSafeFragment : BaseViewBindingFragment<FragmentNoSafesBinding>() {
         super.onViewCreated(view, savedInstanceState)
 
         binding.loadSafeButton.setOnClickListener {
-            requireParentFragment().findNavController().navigate(R.id.action_to_add_safe_nav)
+            requireParentFragment().findNavController().navigate(R.id.action_to_add_safe)
```
jpalvarezl, line 32: Can't you here the typed safe nav args here?

app/src/main/res/layout/fragment_transactions.xml
```diff
@@ -23,4 +23,4 @@
 
     </LinearLayout>
 
-</androidx.constraintlayout.widget.ConstraintLayout>
+</androidx.coordinatorlayout.widget.CoordinatorLayout>
```
jpalvarezl, line 26: What issue was this triggering? Seems like you have 1 level of unnecessary nesting here.

jpalvarezl: As the `LinearLayout` has the same height and width as the parent.

app/src/main/java/io/gnosis/safe/ui/safe/empty/NoSafeFragment.kt
jpalvarezl: Also for the other ones. Like `AddSafeFragment.globalActionId` or something?

jpalvarezl (approved):
I am just approving as the changes I am proposing are more about style than functionality.

app/src/main/res/layout/fragment_transactions.xml
elgatovital: confirmation snackbar informing user about successful owner import has to be shown above the bottom bar

app/src/main/java/io/gnosis/safe/ui/safe/empty/NoSafeFragment.kt
elgatovital: we don't need any arguments here.  We just navigating to the point defined by the action id. In this specific line I have renamed the action for a better readability. It had nav suffix from the times it pointed to another subgraph.

jpalvarezl: In this case though, this no longer "refactor" proof, as this will fail in runtime instead of compile time. I agree with you, arguments are not necessary, but the destination classes that get generated are helpful in case of a refactor so we can fail in compile time instead of runtime.

Issue #1120, opened 2020-12-02
Not implemented errors from the New errors format


liliya-soroka:
**Describe the bug**
The error :  description: An error accured. please try again later. for the next error codes:
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/46495a04-1fe3-4b4a-9f64-832d6ff8dcb1)( except 500)
**To Reproduce**
1. Connect to proxy
2. run mitmproxy with 4xx , 3xx, 5xx errors
2. Go over the app pages and flows (balance, collectables.. with client.gateway and add safe, safe settings with transactions https://safe-transaction.staging.gnosisdev.com/api/v1/safes/0x8bc9Ab35a2A8b20ad8c23410C61db69F2e5d8164

**Expected behavior**
The error according to the requirments https://docs.google.com/document/d/1foQ9DSMgO2iN-6u4JsJIDJWB7mUHX-D8ApZWvx6oKAo/edit# should be displayed for any 3xx, 4xx( including 404, 422 with code 1), 5xx should be displayed instead of the default error

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/2c3bd90b-11f1-49f6-b4da-f5e8966c4fc2)

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.5.0 -1595

**Environment**
staging, rinkeby
(unsafe)
**Additional context**
```
import json
from mitmproxy import http

def request(flow: http.HTTPFlow) -> None:
    if flow.request.pretty_url.startswith("https://safe-transaction.staging.gnosisdev.com"):
        data = {"code": 1}
        flow.response = http.HTTPResponse.make(
            503,  # (optional) status code
            json.dumps(data, indent=4, sort_keys=False),  # (optional) content
            {"Content-Type": "application/json"}  # (optional) headers
        )
    elif flow.request.pretty_url.startswith("https://safe-client.rinkeby.staging.gnosisdev.com"):
        data = {"code": 1}
        flow.response = http.HTTPResponse.make(
            503,  # (optional) status code
            json.dumps(data, indent=4, sort_keys=False),  # (optional) content
            {"Content-Type": "application/json"}  # (optional) headers
        )
```

liliya-soroka:
Reopened:
The correct message was added only for 500 error
app version 2.8.0 -1882rc unsafe

liliya-soroka:
Verified
[...]

PR #1247, opened 2021-02-04
Missing error codes

biafra23:
Handles #1120 

- Parse body for error code for 422 response code
- Treat unmatched 3xx, 4xx and 5xx the same but preserve http response code as error code
- Tracking of errors
- Unit tests added

@gnosis/mobile-devs


data/src/main/java/io/gnosis/data/backend/TransactionSerivceApi.kt
```diff
@@ -8,7 +8,7 @@ import retrofit2.http.Path
 @Deprecated("use client gw")
 interface TransactionServiceApi {
 
-    @GET("v1/safes/{address}")
+    @GET("v1/safes/{address}/")
```
biafra23, line 11: Prevent redirect. We should check all endpoints for redirects.

app/src/main/java/io/gnosis/safe/ErrorHandler.kt
```diff
@@ -77,39 +105,73 @@ enum class Error(
...
     }
+
+@JsonClass(generateAdapter = true)
+data class ServerError(
+    @Json(name = "code") val code: Int,
+    @Json(name = "message") val message: String?,
+    @Json(name = "arguments") val arguments: List<String>?
+)
+
+private val serverErrorAdapter = dataMoshi.adapter(ServerError::class.java)
```
elgatovital, line 177: why do we need this adapter here?

app/src/main/java/io/gnosis/safe/ErrorHandler.kt
```diff
@@ -77,39 +105,73 @@ enum class Error(
...
+                        val errorBodyString = it.response()?.errorBody()?.string()
+                        val serverError = errorBodyString?.let {
+                            if (errorBodyString.isNotEmpty()) {
+                                serverErrorAdapter.fromJson(errorBodyString)
+                            } else {
+                                null
+                            }
+                        }
+
+                        if (Tracker.isInitialized()) {
```
elgatovital, line 139: as discussed let's have tracking logic at places where mapping is called. BaseFragment and BaseActivity already have tracker property.

biafra23: Done in: 2a2648867868a5b1437085a01e4898229f1b77cf

elgatovital (approved):

PR #1248, opened 2021-02-05
Show known address names in tx list

elgatovital:
Handles #1154

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewModel.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewModel.kt
@@ -6,6 +6,7 @@ import androidx.paging.PagingData
 import androidx.paging.cachedIn
 import androidx.paging.filter
 import androidx.paging.map
+import io.gnosis.data.models.AddressInfo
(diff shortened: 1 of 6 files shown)


app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewModel.kt
```diff
@@ -220,19 +222,28 @@ class TransactionListViewModel
...
+
+        val address = txInfo.to.asEthereumAddressString()
+        val localName = safes.find { it.address == txInfo.to }?.localName
+        val toInfo = txInfo.toInfo
+
+        val addressInfo = when {
+            localName != null -> AddressInfoData.Local(localName, address)
+            toInfo != null -> AddressInfoData.Remote(toInfo.name, toInfo.logoUri, address)
+            else -> AddressInfoData.Default
+        }
```
biafra23, line 245: Could this (line 237 - 245) be extracted into a method to avoid duplication? Looks like this is the same code as in line 271 - 279.

biafra23 (approved):
Looks very good! I like the mini blockies. 👍

Issue #1246, opened 2021-02-04
[Android] Tracking: missed tracking from signing v2


liliya-soroka:
**Describe the bug**
1. (see: https://github.com/gnosis/safe/wiki/Tracking)
New param for user_key_imported event
import_type (String): seed or key

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.9.0 -2006rc

**Environment**
prod mainnet
**Additional context**

tschubotz:
Marking this as `major` not because user experience is impaired but because we can't tell if now users add more keys due to due the pk import option. So I think this should be fixed in the next release right away.

liliya-soroka:
Reopened:
I still see user_key_imported for both key and seed 
I am expecting to see
user_key_imported for a private key and user_seed_imported for the seed prise

app version 2.9.0 - debug

elgatovital:
@liliya-soroka where do you get these requirements from

liliya-soroka:
Verified
correct import type is displayed for seed phrase and private key
app version 2.10.0 - 2058rc

PR #1251, opened 2021-02-08
Track owner key import type

elgatovital:
Handles #1246

@gnosis/mobile-devs


app/src/test/java/io/gnosis/safe/ui/settings/owner/OwnerSelectionViewModelTest.kt
```diff
@@ -61,7 +61,7 @@ class OwnerSelectionViewModelTest {
             ownerCredentialsVault.storeCredentials(any())
             settingsHandler.showOwnerBanner = false
             settingsHandler.showOwnerScreen = false
-            tracker.logKeyImported()
+            tracker.logKeyImported(any())
```
biafra23, line 64: Since we know `logKeyImported()` should be passed `true` here, why not very it is?

biafra23 (approved):
LGTM

Issue #1253, opened 2021-02-11
A queued transaction is not shown


vporton:
**Describe the bug**
When I click on a transaction in the queue, it is loaded infinitely long.

**To Reproduce**
- Create a transaction with `gnosis-safe.io/app/` in a multisig wallet with an Android phone.
- Click queue.
- Click an unconfirmed transaction in the queue.

**Expected behavior**
Show the transaction.

**Smartphone (please complete the following information):**
 - Device: Redmi 7A
 - Android Version: 10 QKQ1.191014.001
 - App Version: 2.9.0

tschubotz:
Thanks for the report @vporton .

@liliya-soroka reproduced this when interacting with 1inch. cf. https://gnosisinc.slack.com/archives/CQB729LLF/p1613059609305500?thread_ts=1613052192.290100&cid=CQB729LLF

vporton:
> Thanks for the report @vporton .
> 
> @liliya-soroka reproduced this when interacting with 1inch. cf. https://gnosisinc.slack.com/archives/CQB729LLF/p1613059609305500?thread_ts=1613052192.290100&cid=CQB729LLF

I can't view this message. When I try to sing up for slack, I see:
```
The email address must match one of the domains listed below. Please try another email.
```

tschubotz:
@vporton this wasn't meant for you but for us internally as a reference on how to reproduce this. We don't share info such as Safe addresses on Github if we get them via privately, e.g. via direct messages. Hence that non-public reference.

liliya-soroka:
sorry, I scrolled just a little bit and did  not see that I need to scroll more 
On MacOS 13" I need to scroll down a lot to see that something more is presented on the page
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/a7d48c5f-d253-456b-923a-b1b855caab03)
I will change the description and expectation

liliya-soroka:
Verified
app version 2.10.0 - 2063rc

PR #1258, opened 2021-02-12
Fix: A queued transaction is not shown

biafra23:
Handles #1253 

- Regression test added
-

@gnosis/mobile-devs


elgatovital (approved):
:+1:

Issue #1267, opened 2021-02-15
[Andoid] Parse rejection transaction


MouazAlzahabi:
We need to add new type of transaction called "rejection" and parse it from the response of queued/history transactions, the rejection transaction is custom transaction with new flag isCancellation = true in TxInfo. 
The following [link](https://github.com/gnosis/safe-client-gateway/pull/261/files) is for reference

liliya-soroka:
Verified
The cancelled txs are defined correctly in Queue /History 
app version 2.9.0 - 2099

PR #1279, opened 2021-02-16
Parse isCancellation flag 

biafra23:
Handles #1267

- Add `isCancellation` to `TransactionInfo.Custom`
- Populate `isCancellation` to `TransactionView.CustomTransaction` and `TransactionView.CustomTransactionQueued`
- Adjust tests

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/TransactionView.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/TransactionView.kt
@@ -85,8 +85,7 @@ sealed class TransactionView(
         val alpha: Float,
         val nonce: String,
         val methodName: String?,
-        val addressInfo: AddressInfoData,
-        val isCancellation: Boolean
+        val addressInfo: AddressInfoData
     ) : TransactionView(status, id)
 
     data class CustomTransactionQueued(
@@ -101,8 +100,7 @@ sealed class TransactionView(
         @DrawableRes val confirmationsIcon: Int,
         val nonce: String,
         val methodName: String?,
-        val addressInfo: AddressInfoData,
-        val isCancellation: Boolean
+        val addressInfo: AddressInfoData
     ) : TransactionView(status, id)
 
     data class Creation(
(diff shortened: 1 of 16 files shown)


app/src/test/java/io/gnosis/safe/ui/transactions/TransactionListViewModelTest.kt
```diff
@@ -489,7 +489,8 @@ class TransactionListViewModelTest {
                 confirmationsTextColor = R.color.primary,
                 threshold = 2,
                 confirmations = 2,
-                addressInfo = AddressInfoData.Default
+                addressInfo = AddressInfoData.Default,
+                isCancellation = false
```
MouazAlzahabi, line 493: Just small comment here, 

As we agreed later on there will be new type of transaction for rejection transaction, and in UI it's handled as new type so I recommend to create new model type instead of just flag.

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionView.kt
```diff
@@ -100,7 +101,8 @@ sealed class TransactionView(
         @DrawableRes val confirmationsIcon: Int,
         val nonce: String,
         val methodName: String?,
-        val addressInfo: AddressInfoData
+        val addressInfo: AddressInfoData,
+        val isCancellation: Boolean
```
MouazAlzahabi, line 105: Just small comment here,

As we agreed later on there will be new type of transaction for rejection transaction, and in UI it's handled as new type so I recommend to create new model type instead of just flag.

elgatovital (approved):

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionView.kt
elgatovital: @biafra23 you could add Rejection and RejectionQueued to io.gnosis.safe.ui.transactions.TransactionView

elgatovital (changes requested):

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionView.kt
biafra23: To be clear `Rejection` should look like this?

```
    @JsonClass(generateAdapter = true)
    data class Rejection(
        @Json(name = "to") val to: Solidity.Address,
        @Json(name = "toInfo") val toInfo: AddressInfo?,
        @Json(name = "dataSize") val dataSize: Int,
        @Json(name = "value") val value: BigInteger,
        @Json(name = "methodName") val methodName: String?
    ) : TransactionInfo(TransactionType.Rejection)
```

Why does it need to have `methodName`, `value` and `dataSize`?

biafra23: Should we really call it `Rejection` if the backend calls it `cancellation_transaction`?

biafra23: Oh wait. That json sample was probably the response from the transaction service. 

@jpalvarezl 
Where can I find documentation on the `Rejection` txInfo type?

elgatovital (approved):

elgatovital (approved):

Issue #1270, opened 2021-02-15
[Android] Submit rejection transaction


MouazAlzahabi:
We want to create rejection transaction and submit it. 

In order to submit any transaction to back end, a new [endpoint](https://github.com/gnosis/safe-client-gateway/wiki/transactions_proposal) already created. 

In order to submit a rejection transaction we have to use the previous endpoint with the flowing parameters: 
- **To:** safe address
- **Nonce:** the nonce of tx to be canceled.
- **Signature:** the hash of current tx signed by owner key
- **Safe tx gas:** must be zero
- **Sender:** Owner public address
- **All other params:** zero values

### **Notes**
- We can use same error codes and create signature method  used for submit confirmation

liliya-soroka:
Verified
app version 2.11.0 -2142rc

PR #1283, opened 2021-02-18
Create and submit rejection

biafra23:
Handles #1270

- Add new endpoint
- Add model for tx proposal
- create cancellation tx
- submit rejection to client gateway

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
```diff
@@ -158,7 +163,8 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
                             confirmColor = R.color.primary
                         ) {
                             binding.txConfirmButton.isEnabled = false
-                            viewModel.submitConfirmation(viewModel.txDetails!!)
+//                            viewModel.submitConfirmation(viewModel.txDetails!!)
+                            viewModel.submitRejection()
```
jpalvarezl, line 167: Are you just using the the UI for confirmation to test?

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
```diff
@@ -50,8 +51,47 @@ class TransactionDetailsViewModel
         @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
         set
 
+    fun submitRejection() {
+        val executionInfo = txDetails?.detailedExecutionInfo as? DetailedExecutionInfo.MultisigExecutionDetails
+            ?: throw RuntimeException("MissingCorrectExecutionDetailsException")
```
jpalvarezl, line 56: I think here you could have a `class MissingCorrectExecutionDetailsExpect : Exception` instead, and leverage type safety at the UI level to display specific error messages

data/src/main/java/io/gnosis/data/models/transaction/TransactionDetails.kt
```diff
@@ -47,27 +50,27 @@ sealed class DetailedExecutionInfo(
...
+        val safeTxGas: BigInteger = BigInteger.ZERO,
         @Json(name = "baseGas")
-        val baseGas: BigInteger,
+        val baseGas: BigInteger = BigInteger.ZERO,
         @Json(name = "gasPrice")
-        val gasPrice: BigInteger,
+        val gasPrice: BigInteger = BigInteger.ZERO,
         @Json(name = "gasToken")
-        val gasToken: Solidity.Address
+        val gasToken: Solidity.Address = "0".asEthereumAddress()!!
```
jpalvarezl, line 73: Why do you need default values for all he fields in `MultisigExecutionDetails` and `TransactionDetails`?

data/src/main/java/io/gnosis/data/models/transaction/TransactionInfo.kt
```diff
@@ -12,11 +12,11 @@ sealed class TransactionInfo(
...
-        @Json(name = "toInfo") val toInfo: AddressInfo?,
-        @Json(name = "dataSize") val dataSize: Int,
-        @Json(name = "value") val value: BigInteger,
-        @Json(name = "methodName") val methodName: String?,
-        @Json(name = "isCancellation") val isCancellation: Boolean
+        @Json(name = "toInfo") val toInfo: AddressInfo? = null,
+        @Json(name = "dataSize") val dataSize: Int = 0,
+        @Json(name = "value") val value: BigInteger = BigInteger.ZERO,
+        @Json(name = "methodName") val methodName: String? = null,
+        @Json(name = "isCancellation") val isCancellation: Boolean = false
```
jpalvarezl, line 19: Just wondering why default values are necessar 🤔

data/src/main/java/io/gnosis/data/repositories/EnsRepository.kt
```diff
@@ -2,7 +2,10 @@ package io.gnosis.data.repositories
 
 import io.gnosis.contracts.BuildConfig
 import pm.gnosis.crypto.utils.Sha3Utils
-import pm.gnosis.ethereum.*
+import pm.gnosis.ethereum.Block
+import pm.gnosis.ethereum.EthCall
+import pm.gnosis.ethereum.EthRequest
+import pm.gnosis.ethereum.EthereumRepository
```
jpalvarezl, line 8: Are these unused imports or the just got unfolded by the IDE?

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
```diff
@@ -33,18 +32,47 @@ class TransactionRepository(
...
+    suspend fun proposeTransaction(
+        safeAddress: Solidity.Address,
+        value: BigInteger = BigInteger.ZERO,
+        data: String? = null,
+        nonce: BigInteger,
+        operation: Operation = Operation.CALL,
+        safeTxGas: Long = 0,
+        baseGas: Long = 0,
+        gasPrice: Long = 0,
+        gasToken: Solidity.Address = "0x00".asEthereumAddress()!!,
```
jpalvarezl, line 44: I would use the more standard `0x0000000000000000000000000000000000000000`, as the web interface allows `0x0` which is already unconventional.

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
```diff
@@ -33,18 +32,47 @@ class TransactionRepository(
...
+        refundReceiver: Solidity.Address? = null,
+        safeTxHash: String,
+        sender: Solidity.Address,
+        signature: String,
+        origin: Solidity.Address? = null
+    ): Response<Unit> =
```
jpalvarezl, line 50: I think that you don't need a return value for responses that don't have a body. Unless, you are doing something with the http code.

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
elgatovital: why does it matter here? Type is Solidity.Address so it will be converted to the same value. I would prefer shorter notation here.

data/src/main/java/io/gnosis/data/models/transaction/TransactionInfo.kt
elgatovital: it will crash if values will be missing in the backend response because types are not nullable. I think it's good practice to stay on the safe side.

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
elgatovital: yes, ui part is in the separate pr. We should merge it first

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
biafra23: I tried. It doesn't work without a return value. Let me have another look.

data/src/main/java/io/gnosis/data/models/transaction/TransactionDetails.kt
biafra23: I used default values, so I could make the invocation shorter.

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
biafra23: AFAICT, we're not catching this anywhere. But the Button is hidden on non-eligible transactions. I don't mind making this a specific type.

data/src/main/java/io/gnosis/data/repositories/EnsRepository.kt
biafra23: These are not unused imports.

data/src/main/java/io/gnosis/data/repositories/TransactionRepository.kt
biafra23: I was mistaken. It works without a return type. So I removed it.

elgatovital (approved):

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
biafra23: This has now UI.

PR #1289, opened 2021-02-19
Notification channels

biafra23:
- Add Notification Channel Group per added safe
- Update Group when local safe name is changed
- Delete Notification Channel Group when a safe is removed from the app
- On app startup: Create Notification Channel Group per existing safe (this can be removed in a future update)

Home -> Safe Multisig -> App Info -> Notifications: 
Screenshot_20210219-234453

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/notifications/NotificationManager.kt
+++ b/app/src/main/java/io/gnosis/safe/notifications/NotificationManager.kt
@@ -9,6 +9,7 @@ import androidx.annotation.StringRes
 import androidx.core.app.NotificationCompat
 import androidx.core.app.NotificationManagerCompat
 import io.gnosis.data.models.Safe
+import io.gnosis.data.repositories.SafeRepository
 import io.gnosis.data.repositories.TokenRepository
 import io.gnosis.safe.R
 import io.gnosis.safe.notifications.models.PushNotification
@@ -16,6 +17,7 @@ import io.gnosis.safe.ui.StartActivity
 import io.gnosis.safe.utils.BalanceFormatter
 import io.gnosis.safe.utils.convertAmount
 import io.gnosis.safe.utils.formatForTxList
+import kotlinx.coroutines.runBlocking
 import pm.gnosis.crypto.utils.asEthereumAddressChecksumString
 import pm.gnosis.svalinn.common.PreferencesManager
 import pm.gnosis.svalinn.common.utils.edit
@@ -24,10 +26,22 @@ import timber.log.Timber
 class NotificationManager(
     private val context: Context,
     private val preferencesManager: PreferencesManager,
-    private val balanceFormatter: BalanceFormatter
+    private val balanceFormatter: BalanceFormatter,
+    private val safeRepository: SafeRepository,
+    private val notificationManager: NotificationManagerCompat = NotificationManagerCompat.from(context)
 ) {
 
-    private val notificationManager = NotificationManagerCompat.from(context)
+    init {
+        if (Build.VERSION.SDK_INT >= 26) {
+            // For upgrading users. Can be removed in a future release?
+            runBlocking {
+                safeRepository.getSafes().forEach { safe ->
+                    createNotificationChannelGroup(safe)
+                }
+            }
+        }
+    }
+
 
     private var latestNotificationId: Int
         get() = preferencesManager.prefs.getInt(LATEST_NOTIFICATION_ID, -1)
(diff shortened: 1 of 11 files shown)


app/src/main/java/io/gnosis/safe/ui/safe/add/AddSafeNameViewModel.kt
```diff
@@ -31,7 +33,8 @@ class AddSafeNameViewModel
             runCatching {
                 val safe = Safe(address, localName.trim())
                 safeRepository.saveSafe(safe)
-                notificationRepository.registerSafe(safe.address)
+                notificationRepository.registerSafe(safe)
+                notificationManager.createNotificationChannelGroup(safe)
```
elgatovital, line 37: the channel should also be removed when removing a safe, right?

app/src/main/java/io/gnosis/safe/notifications/NotificationManager.kt
```diff
@@ -40,21 +50,57 @@ class NotificationManager(
             }
         }
 
+    fun createNotificationChannelGroup(safe: Safe) {
+        val id = safe.address.asEthereumAddressChecksumString()
+        val name = safe.localName
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            notificationManager.createNotificationChannelGroup(NotificationChannelGroup(id, name))
+            createNotificationChannel(id)
```
elgatovital, line 58: should notification channel id's have also some common prefix?

app/src/main/java/io/gnosis/safe/ui/safe/add/AddSafeNameViewModel.kt
biafra23: `deleteNotificationChannelGroup()` removes the all groups channels and the group itself. 
```
    fun deleteNotificationChannelGroup(safe: Safe) {
        val subChannels = ChannelType.values()

        subChannels.forEach { subChannel ->
            notificationManager.deleteNotificationChannel(safe.address.asEthereumAddressChecksumString() + "." + context.getString(subChannel.resId))
        }
        notificationManager.deleteNotificationChannelGroup(safe.address.asEthereumAddressChecksumString())
    }
´´´

Called in `SafeSettingsViewModel.removeSafe()`

File: app/src/main/java/io/gnosis/safe/notifications/NotificationManager.kt (hunk 2 above)
Comment (biafra23) [reply app/src/main/java/io/gnosis/safe/notifications/NotificationManager.kt:58]: It's not necessary. What would be an advantage?

File: app/src/main/java/io/gnosis/safe/notifications/NotificationManager.kt [hunk 3]
```diff
@@ -1,35 +1,45 @@
...
+    private val balanceFormatter: BalanceFormatter,
+    private val safeRepository: SafeRepository,
+    private val notificationManager: NotificationManagerCompat = NotificationManagerCompat.from(context)
 ) {
 
-    private val notificationManager = NotificationManagerCompat.from(context)
-
     init {
-        createNotificationChannel(CHANNEL_ID)
+        if (Build.VERSION.SDK_INT >= 26) {
```
elgatovital, line 34: where is the else branch? What about skd levels < 26? Right not not channel will be created at all

app/src/main/java/io/gnosis/safe/notifications/NotificationManager.kt
```diff
@@ -1,35 +1,45 @@
...
+            // For upgrading users. Can be removed in a future release?
```
elgatovital, line 35: can we check if channel group already exists?

app/src/main/java/io/gnosis/safe/notifications/NotificationManager.kt
biafra23: The previous code also didn't create a channel for api level < 26.

app/src/main/java/io/gnosis/safe/notifications/NotificationManager.kt
biafra23: Apparently, we can. Will rewrite to check first.

app/src/main/java/io/gnosis/safe/notifications/NotificationManager.kt
elgatovital: you are right, there simply were no different channels at those api levels

elgatovital (approved):

Issue #1272, opened 2021-02-15
[Android] Enable reject transaction in transaction details


MouazAlzahabi:
We need to enable reject transaction in transaction details. 

In order to do so, we want to add "Reject" button as show in the [design](https://zpl.io/amjpDNe).

Click on reject button will navigate to [Confirm rejection screen](https://github.com/gnosis/safe-android/issues/1271)

[Backend should return](https://github.com/gnosis/safe-client-gateway/issues/258) if this transaction was already rejected, in that case the button is disabled.

liliya-soroka:
Verified
everything except " Backend should return if this transaction was already rejected, in that case, the button is disabled."
Back-end is not ready.
It was agreed that Reject will be displayed in the 2.11.0 release till the back-end is implemented
app version 2.11.0 -2124rc

PR #1291, opened 2021-02-23
De/Activate confirm and reject button

biafra23:
Handles #1272

- Enable and disable buttons given the information we have.
- Show snackbar on tx list (queue)
- parse and check rejections list in `MultisigExecutionDetails`

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/Tracker.kt
+++ b/app/src/main/java/io/gnosis/safe/Tracker.kt
@@ -78,8 +78,10 @@ class Tracker private constructor(context: Context) {
         }
     }
 
-    fun logException(exception: Throwable) {
-        FirebaseCrashlytics.getInstance().recordException(exception)
+    fun logException(exception: Throwable, trackIt : Boolean = true) {
+        if (trackIt) {
+            FirebaseCrashlytics.getInstance().recordException(exception)
+        }
     }
 
     object Event {
--- a/app/src/main/java/io/gnosis/safe/Tracker.kt
+++ b/app/src/main/java/io/gnosis/safe/Tracker.kt
@@ -78,10 +78,8 @@ class Tracker private constructor(context: Context) {
         }
     }
 
-    fun logException(exception: Throwable, trackIt : Boolean = true) {
-        if (trackIt) {
-            FirebaseCrashlytics.getInstance().recordException(exception)
-        }
+    fun logException(exception: Throwable) {
+        FirebaseCrashlytics.getInstance().recordException(exception)
     }
 
     object Event {
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/details/ButtonStateHelper.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/details/ButtonStateHelper.kt
@@ -42,7 +42,6 @@ class ButtonStateHelper(
             else -> false
         }
 
-
     fun rejectButtonIsEnabled(): Boolean =
         when {
             !rejectButtonIsVisible() -> {
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/details/ButtonStateHelper.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/details/ButtonStateHelper.kt
@@ -34,6 +34,9 @@ class ButtonStateHelper(
         if (!rejectButtonIsVisible()) {
             return false
         }
+        if (!isRejection && needsYourConfirmation && hasBeenRejected && !needsExecution) {
+            return false
+        }
         return true
     }
 
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/details/ButtonStateHelper.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/details/ButtonStateHelper.kt
@@ -4,11 +4,17 @@ class ButtonStateHelper(
     val hasBeenRejected: Boolean,
     val needsYourConfirmation: Boolean,
     val isRejection: Boolean,
-    val needsExecution: Boolean
+    val needsExecution: Boolean,
+    val canReject: Boolean,
+    val isOwner: Boolean,
+    val completed: Boolean
 ) {
 
     fun rejectButtonIsVisible(): Boolean {
-        if (isRejection) {
+        if (!isOwner || completed) {
+            return false
+        }
+        if (isRejection || !canReject) {
             return false
         }
         if (!isRejection && !needsYourConfirmation && hasBeenRejected && needsExecution) {
@@ -18,6 +24,9 @@ class ButtonStateHelper(
     }
 
     fun confirmButtonIsVisible(): Boolean {
+        if (!isOwner || completed) {
+            return false
+        }
         if (needsYourConfirmation) {
             return true
         }
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/details/ButtonStateHelper.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/details/ButtonStateHelper.kt
@@ -10,53 +10,58 @@ class ButtonStateHelper(
     val completed: Boolean
 ) {
 
-    fun rejectButtonIsVisible(): Boolean {
-        if (!isOwner || completed) {
-            return false
-        }
-        if (isRejection || !canReject) {
-            return false
-        }
-        if (!isRejection && !needsYourConfirmation && hasBeenRejected && needsExecution) {
-            return false
-        }
-        return true
-    }
+    fun rejectButtonIsVisible(): Boolean =
+        when {
+            !isOwner || completed -> {
+                false
+            }
+            isRejection || !canReject -> {
+                false
+            }
+            !isRejection && !needsYourConfirmation && hasBeenRejected && needsExecution -> {
+                false
+            }
+            else -> true
 
-    fun confirmButtonIsVisible(): Boolean {
-        if (!isOwner || completed) {
-            return false
         }
-        if (needsYourConfirmation) {
-            return true
-        }
-        if (!isRejection && !hasBeenRejected && !needsExecution) {
-            return true
-        }
-        if (!isRejection && !needsYourConfirmation && hasBeenRejected && needsExecution) {
-            return true
-        }
-        return false
-    }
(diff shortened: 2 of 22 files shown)


app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
```diff
@@ -88,6 +88,7 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
                         updateUi(it)
                     }
                     snackbar(requireView(), R.string.rejection_successfully_submitted)
```
elgatovital, line 90: will we see this confirmation snackbar? Also will it appear on top of the bottom bar?

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
```diff
@@ -422,6 +423,14 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
                         getStringResForStatus(txDetails.txStatus, needsYourConfirmation),
                         getColorForStatus(txDetails.txStatus)
                     )
+                    if (needsYourConfirmation) {
```
elgatovital, line 426: we have check for this at the top outside of the view stub layout, right? As button container is not part of the view stub I don't think this is the right place to set visibility of the buttons.

app/src/main/res/layout/fragment_transaction_details.xml
```diff
@@ -183,7 +183,7 @@
             android:layout_width="match_parent"
             android:layout_weight="1"
             android:layout_marginLeft="@dimen/default_margin"
-            android:layout_marginRight="20dp"
+            android:layout_marginRight="4dp"
```
elgatovital, line 186: I don't know if this makes it more readable when you have to add different margins instead of explicitly define one value.

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
biafra23: There is two problems with the current implementation. 
1. Sometime the snackbar doesn't appear
2. The Rejection confirmation screen goes to the details first, submits the rejection (sometimes shows the snackbar) and then jumps to the tx list

app/src/main/res/layout/fragment_transaction_details.xml
biafra23: Otherwise the confirm button doesn't have the right margin when the reject button is gone.

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
biafra23: The rejections have different rules on whether to display the button or not.

app/src/main/res/layout/fragment_transaction_details.xml
biafra23: This needed to be changed because the other button can also be gone :-(

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
biafra23: This has been resolved.

app/src/main/java/io/gnosis/safe/ui/transactions/details/ConfirmRejectionFragment.kt
```diff
@@ -39,13 +56,49 @@ class ConfirmRejectionFragment : BaseViewBindingFragment<FragmentConfirmRejectio
...
+                        when (it) {
+                            is TxRejectionFailed -> {
+                                val error = it.cause.toError()
+                                if (error.trackingRequired) {
+                                    tracker.logException(it.cause)
+                                }
+                                errorSnackbar(requireView(), error.message(requireContext(), R.string.error_description_tx_rejection))
+                            }
+                            else -> {
+                                val error = it.toError()
```
elgatovital, line 92:
```val error = it.cause.toError()
                                if (error.trackingRequired) {
                                    tracker.logException(it.cause)
                                }
```
this part is same for all branches and can be moved outside when

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
elgatovital: sure, but can we do this check at the top in one place. Otherwise we are doing this work twice

app/src/main/res/navigation/main_nav.xml
```diff
@@ -259,7 +259,17 @@
         android:id="@+id/confirmRejectionFragment"
         android:name="io.gnosis.safe.ui.transactions.details.ConfirmRejectionFragment"
         android:label="ConfirmRejectionFragment"
-        tools:layout="@layout/fragment_confirm_rejection" />
+        tools:layout="@layout/fragment_confirm_rejection" >
+
+        <!-- TODO: Add action to go to tx list -->
+        <action
+            android:id="@+id/action_confirmRejectionFragment_to_transactionsFragment"
+            app:destination="@id/transactionsFragment" />
```
elgatovital, line 267: we need to pop tx details and rejection confirmation fragments from stack. Either directly in code or by adding popUpTo to action

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
biafra23: Done in 2720218293f4bb404309d24d030988c9373f9e6e

app/src/main/res/navigation/main_nav.xml
biafra23: Done in  5fc0c7289b0bb3958f5b047015e4d2d3c9ebd93a

app/src/main/java/io/gnosis/safe/ui/transactions/details/ConfirmRejectionFragment.kt
biafra23: For `TxRejectionFailed` error is in `cause` while for the other two it is the Throwable itself. Isn't that enough of a difference to have when  or if-else?

biafra23: What do you think about changing `logException()` to take a Boolean parameter and always pass `error.trackingRequired`. 

If that's ok for you, I'd use it everywhere.

elgatovital (approved):

app/src/main/java/io/gnosis/safe/ui/transactions/details/ConfirmRejectionFragment.kt
elgatovital: I think logException should do what it says - log exception without additional checks

elgatovital: can we then propagate cause directly in TxRejectionFailed, so we don't have this extra logic in fragment?

biafra23:
I am currently extracting the button-logic into a separate and testable class.

app/src/main/java/io/gnosis/safe/Tracker.kt
```diff
@@ -78,8 +78,10 @@ class Tracker private constructor(context: Context) {
         }
     }
 
-    fun logException(exception: Throwable) {
-        FirebaseCrashlytics.getInstance().recordException(exception)
+    fun logException(exception: Throwable, trackIt : Boolean = true) {
```
elgatovital, line 81: I think this change should be reverted. If exception should not be trackable it is better not to call logException at all. logException(e, false) is missleading.

app/src/main/java/io/gnosis/safe/ui/transactions/details/ConfirmRejectionFragment.kt
biafra23: You mean `cause` instead of `TxRejectionFailed`? Is that a good idea? It is not that much logic, is it?

app/src/main/java/io/gnosis/safe/Tracker.kt
biafra23: Done

app/src/main/java/io/gnosis/safe/ui/transactions/details/ButtonStateHelper.kt
```diff
@@ -0,0 +1,66 @@
...
+    val needsYourConfirmation: Boolean,
+    val isRejection: Boolean,
+    val needsExecution: Boolean,
+    val canReject: Boolean,
+    val isOwner: Boolean,
+    val completed: Boolean
+) {
+
+    fun rejectButtonIsVisible(): Boolean {
+        if (!isOwner || completed) {
```
elgatovital, line 14: not critical: it would be compacter with when {...

app/src/main/java/io/gnosis/safe/ui/transactions/details/ConfirmRejectionFragment.kt
elgatovital: no, I mean passing cause directly to TxRejectionFailed as a parameter

app/src/main/java/io/gnosis/safe/ui/transactions/details/ButtonStateHelper.kt
biafra23: Done

elgatovital:
:+1:

app/src/main/java/io/gnosis/safe/ui/transactions/details/ConfirmRejectionFragment.kt
biafra23: Like this? ef0e73fe4e3a6608c9cd6984327a2467ea5881dd

Issue #1274, opened 2021-02-16
[Android] Passcode settings


elgatovital:
- [x] Add Passcode item to app settings
- [x] On click navigate to Passcode settings screen
- [x] Security details screen content
   - [x] Before passcode was set up / passcode is disabled show following items:
   - [x] Enable passcode
    - [x]  On click start setup flow if there was no passcode before
     - [x] Update screen ui and set passcode to enabled (no passcode confirmation is needed)
   - [x] After passcode was set up / passcode is enabled show following items:
       - [x] Change passcode
           - [ ] On click start change passcode flow
       - [x] Global disable passcode
         - [x] On click show passcode confirmation screen
         - [x] After successful confirmation set passcode to disabled 
         - [x] Show error message if entered passcode is not correct???
            - E.g. shake passcode input field
         - [x] Return to security details & update ui 

## Design
- [Security Settings screen (Zeplin)](https://zpl.io/2En3Gxv) (Passcode enabled) 
![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/b366acf4-878c-4f26-adb7-041f8c71af18)

- [Security Settings Screen (Zeplin)](https://zpl.io/VO63Enk) (Passcode disabled)
![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/e63fe610-8ecd-4b51-9640-b298ed834c63)

liliya-soroka:
Reopened:

The digital keyboard should be used during the passcode set up/change/enter instead of the usual keyboard

app version 2.13.0 -2282

biafra23:
>     The digital keyboard should be used during the passcode set up/change/enter instead of the usual keyboard

1. What do you mean by digital keyboard? Where is this requirement from? What is a digital keyboard? All screens should show a numeric keyboard. According to the designs: https://zpl.io/VDP3Qo3
2. The change passcode part has been moved to a separate ticket: #1340

liliya-soroka:
@biafra23 , yes I mean numeric keyboard, but now we have usual keyboard-
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/6c500dd0-3876-4f0d-a73e-f6e06ae070c4)

biafra23:
@liliya-soroka What phone are you using (that shows the wrong keyboard)?

Do you have another phone or emulator to reproduce this?

liliya-soroka:
Verified
[...]

PR #1309, opened 2021-03-09
Passcode settings

biafra23:
Handles #1274 

App Settings | Passcode disabled | Passcode enabled
------------- | ------------------- | ----------
image | image | image

The entry point has been disabled in the app settings layout.

**App bar title** has been fixed. It is wrong in the screen shots. 

@gnosis/mobile-devs


app/src/main/res/values/styles.xml
```diff
@@ -154,7 +154,7 @@
         <item name="android:paddingEnd">@dimen/default_margin</item>
         <item name="android:textSize">14sp</item>
         <item name="android:fontFamily">sans-serif</item>
-        <item name="android:textColor">@color/text_emphasis_low</item>
+        <item name="android:textColor">@color/text_emphasis_medium</item>
```
biafra23, line 157: I checked the the design for  the other usage of HelpText and it also says text_emphasis_medium. So this change fixes both usages.

elgatovital (approved):

Issue #1299, opened 2021-03-01
[Android] Data Model / Functionality for the Multiple Owner Keys


elgatovital:
- [x] There is no hard limit on the number of keys to be imported (as much as the device allows)

See specification: https://docs.google.com/document/d/13GirjSwieQaqUPz-8HXzJP_rfgmobSp7E4TpbnLlb0M/edit#bookmark=id.nazzjeun92wz

- [x] If user already imported a key in a previous app version, the key’s name will be “Key 0x1234...2345” where the “0x…” part is the ellipsized address of the key

liliya-soroka:
Reopened:

- [ ] If user already imported a key in a previous app version, the key’s name will be “Key 0x1234...2345” where the “0x…” part is the ellipsized address of the key -  not implemented 
Current result :" Imported owner key " is used as a name of the owner key imported in the previous version with single owner key

app version 2.12.0 -2382

liliya-soroka:
Verified
app version 2.14.0 -2417rc

Issue #1273, opened 2021-02-16
[Android] Persist passcode data


sche:
- [x] Store passcode hash in preferences
- [x] Refactor owner key storage and use android KeyStore insead of SharedPreferences
    - [x] Owner keys should be accessible in the background. E.g. notification service registration. This operation should happen in the background without any input from the user.
- [x] When changing passcode store hash derived from new passcode after successful comparison of current passcode hash and hash derived from confirmed passcode
- [x] Store passcode baner related flags:
    - [x] Baner was dismissed
    - [x] Store passcode enabled/disabled state

liliya-soroka:
Verified, but all issues related to 
 Store passcode baner related flags:
 Baner was dismissed
 Store passcode enabled/disabled state
are listed in the separate task ( related to the banner)
app version 2.13.0 - 2337rc

PR #1313, opened 2021-03-11
Persistency for multiple owners and passcode

elgatovital:
Handles #1273
Handles #1299

- Add db table for multiple owners
- Add custom encryption manager 
- Add credentials repository with possibility to access owner credentials for background work
- Deprecate owner credentials vault
- Import existing owner from vault to db

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
@@ -124,9 +124,7 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
     }
 
     private fun updateUi(txDetails: TransactionDetailsViewData, needsYourConfirmation: Boolean, canReject: Boolean, isOwner: Boolean) {
-
-        var needsYourConfirmation = false
-
+        
         var nonce: BigInteger? = null
         when (val executionInfo = txDetails.detailedExecutionInfo) {
             is DetailedExecutionInfo.MultisigExecutionDetails -> {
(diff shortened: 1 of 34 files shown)


app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
```diff
@@ -123,7 +123,7 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
         binding.refresh.isRefreshing = loading
     }
 
-    private fun updateUi(txDetails: TransactionDetailsViewData) {
+    private fun updateUi(txDetails: TransactionDetailsViewData, needsYourConfirmation: Boolean, canReject: Boolean, isOwner: Boolean) {
 
         var needsYourConfirmation = false
```
biafra23, line 128: The parameter `needsYourConfirmation` is shadowed by this.

data/src/main/java/io/gnosis/data/repositories/CredentialsRepository.kt
```diff
@@ -0,0 +1,74 @@
...
+    suspend fun saveOwner(owner: Owner) {
+        ownerDao.save(owner)
+    }
+
+    suspend fun removeOwner(owner: Owner) {
+        ownerDao.delete(owner)
+    }
+
+    fun signWithOwner(owner: Owner, data: ByteArray): String {
+        val cryptoData = EncryptionManager.CryptoData.fromString(owner.privateKey.toString())
```
biafra23, line 64: We need to use:
``` 
val converter = EncryptedByteArray.Converter()
val cryptoData = EncryptionManager.CryptoData.fromString(converter.toStorage(owner.privateKey))
```

Instead of `toString()`

It looks like the converter can be reused.

biafra23: Maybe we can even get rid of this to String -> from String conversion?

biafra23 (approved):
It seems we need to add some tests. `HeimdallEncryptionManager` would be a candidate, nor?

Issue #1276, opened 2021-02-16
[Android] Reset passcode


elgatovital:
- [x] Show forgot confirm passcode screen in navigation stack
- [x] On click show reset passcode screen with one button and disclaimer
    - [ ] On button click remove all owner keys and reset secure passcode
    - [ ] Return to the screen where reset was started and show snackbar with passcode turned off message

![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/34acf79e-eced-4964-bc48-ecc11becdf1d)

screen: https://zpl.io/VqdM456

liliya-soroka:
Reopened:

Steps:
1. Set up password
2. Import owner key
3. Go to the Change password option
4. Click forgot password->confirm
Current result: "Could not delete all owner keys" error message is displayed but the imported owner key was deleted. Passcode is not disabled
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/7cc5a7db-596d-4a6e-9ba1-b7dfe147da1b)
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/f35d8987-3e8d-45f1-85cc-86739e474879)
Expected result: The key should be deleted without the error and the password should be disabled 


app version 2.13.0 - 2315rc

liliya-soroka:
Verified but the https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-android/1349 exists and should be fixed

app version 2.13.0 - 2337rc

Issue #1275, opened 2021-02-16
[Android] Passcode set up flow after owner import


elgatovital:
- [x] Add success screen after owner key is imported with proposal to setup app passcode
- [x] Screen can be skipped & user returns to the place where import owner flow was started
    - [x]  Don’t show an import confirmation snackbar as confirmation was already depicted on the previous screen.
    - [x] If passcode was already set up, import owner key flow remains unchanged: return to the place where flow was started and show snackbar with successful confirmation.
- [x] When entering new passcode navigate automatically to confirm passcode screen
- [x] On skip cancel the whole flow
- [x] User should be able to cancel and return back from the confirmation screen to the previous screen for entering another passcode
- [x] After successful confirmation of the previously entered passcode navigate to passcode success screen

![image](https://user-images.githubusercontent.com/246473/111466746-f50ff180-8723-11eb-9ba2-7358cf5b2736.png)

biafra23:
>  After successful confirmation of the previously entered passcode navigate to passcode success screen

What is the passcode success screen? 

I assumed from teh designs:
1. When entering from passcode settings we go back to passcode setting and show a success snackbar
2. When entering fro owner selection screen we go back to app settings showing a success snackbar

liliya-soroka:
Verified
app version 2.13.0 -2337rc

PR #1318, opened 2021-03-15
Passcode setup

biafra23:
Handles:
- #1275 
- #1276

**Missing**: 
- When passcode is disabled, App Settings are not updated and still show an imported key. 
- Tests for ViewModel

- Entry points: 
   - After safe import
   - When enabling passcode in settings
- Disable passcode (ask for current passcode) 
- Reset passcode without checking passcode (deletes all owner keys)

Import owner key | Create passcode | Confirm passcode 
------------------ | ---------------- | -------------------
image | image | image

Settings | Passcode settings | Create passcode | Confirm passcode
-------- | ------------------- | ----------------- | -------------------
image | image| image |image

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/CreatePasscodeFragment.kt
```diff
@@ -0,0 +1,116 @@
...
+class CreatePasscodeFragment : BaseViewBindingFragment<FragmentPasscodeBinding>() {
+
+    override fun screenId() = ScreenId.CREATE_PASSCODE
+    private val navArgs by navArgs<CreatePasscodeFragmentArgs>()
+    private val ownerImported by lazy { navArgs.ownerImported }
+
+    @Inject
+    lateinit var settingsHandler: SettingsHandler
+
+    companion object {
```
elgatovital, line 33: please put companion object at the end of the fragment

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/CreatePasscodeFragment.kt
```diff
@@ -0,0 +1,116 @@
...
+    override fun inflateBinding(inflater: LayoutInflater, container: ViewGroup?): FragmentPasscodeBinding =
+        FragmentPasscodeBinding.inflate(inflater, container, false)
+
+    override fun inject(component: ViewComponent) {
+        component.inject(this)
+    }
+
+    override fun onResume() {
+        super.onResume()
+        binding.input.showKeyboardForView()
```
elgatovital, line 46: this can be done in onViewCreated and we won't need onResume

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/CreatePasscodeFragment.kt
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+        with(binding) {
+
+            if (ownerImported) {
+                status.visibility = View.VISIBLE
```
elgatovital, line 55: more legible would be 
```
status.visible(ownerImported)
```

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/CreatePasscodeFragment.kt
```diff
@@ -0,0 +1,116 @@
...
+            } else {
+                status.visibility = View.INVISIBLE
+            }
+
+            createPasscode.setText(R.string.settings_passcode_create_a_6_digit_passcode)
```
elgatovital, line 60: can we set this text in xml?

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/CreatePasscodeFragment.kt
biafra23: Problem is: visible(false) -> View.GONE

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/CreatePasscodeFragment.kt
biafra23: Yeah. One of those (create/repeat) can be set in xml.

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/PasscodeSettingsFragment.kt
```diff
@@ -5,15 +5,17 @@ import android.view.LayoutInflater
...
 import io.gnosis.safe.databinding.FragmentSettingsAppPasscodeBinding
 import io.gnosis.safe.di.components.ViewComponent
-import io.gnosis.safe.ui.base.fragment.BaseViewBindingFragment
+import io.gnosis.safe.ui.base.SafeOverviewBaseFragment
 import io.gnosis.safe.ui.settings.app.SettingsHandler
 import pm.gnosis.svalinn.common.utils.visible
 import javax.inject.Inject
 
-class PasscodeSettingsFragment : BaseViewBindingFragment<FragmentSettingsAppPasscodeBinding>() {
+class PasscodeSettingsFragment : SafeOverviewBaseFragment<FragmentSettingsAppPasscodeBinding>() {
```
elgatovital, line 18: this doesn't have to be SafeOverviewBaseFragment. BaseViewBindingFragment should be enough

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/CreatePasscodeFragment.kt
biafra23: The keyboard is gone when coming back to the app without this.

elgatovital: then we can remove same invocation from onViewCreated()

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/CreatePasscodeFragment.kt
elgatovital: you can pass hidden visibility
```
status.visible(ownerImported, View.INVISIBLE)
```

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/PasscodeSettingsFragment.kt
biafra23: The snackbars are in SafeOverviewBaseFragment. Can we move them to BaseViewBindingFragment?

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/CreatePasscodeFragment.kt
biafra23: Fixed in bfeb375a9e072086cde46d36cdbb63c05fcd6b55

elgatovital:
when passcode setup flow is started after successful import of an owner key and user decides to skip it by clicking back button we shouldn't go back to import owner key
image

elgatovital:
submit button changes from
image
to
image
Is there a way to keep it as is? If you navigate back from repeat passcode screen, you are not getting the initial submit icon

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/CreatePasscodeFragment.kt
biafra23: Apparently they were not even needed. I deleted them.

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/DisablePasscodeFragment.kt
```diff
@@ -0,0 +1,109 @@
...
+                    if (input.text.length < 6) {
+                        digits.forEach {
+                            it.background = ContextCompat.getDrawable(requireContext(), R.color.surface_01)
+                        }
+                        (1..text.length).forEach { i ->
+                            digits[i - 1].background = ContextCompat.getDrawable(requireContext(), R.drawable.ic_circle_passcode_filled_20dp)
+                        }
+                    } else {
+
+                        val success = encryptionManager.unlockWithPassword(text.toString().toByteArray())
```
elgatovital, line 79: this should be moved to a view model and tested

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/PasscodeSettingsFragment.kt
elgatovital: I would prefer to keep BaseViewBindingFragment as simple as possible. We can leave it as SafeOverviewBaseFragment for now and move snackbar handling to a separate helper later

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/DisablePasscodeFragment.kt
biafra23: Moved to viewModel. Not yet tested.

elgatovital:
image
I think we could compare passcodes not only at the end, but when every digit is entered. No reason to wait for a complete input if the very first digit is different. Also `passcodes don't match` label could be hidden after user starts new input and entered digits do match with corresponding digits of the passcode entered on the previous screen. Otherwise error label will persist and it informs about error that is no longer actual and pertains to the previous input

elgatovital (approved):

Issue #1296, opened 2021-02-26
[Android] Known Addresses V2: Creation transaction screen


sche:
Known addresses are displayed for:
- Creator address
- Mastercopy used
- Factory used

## Design
https://zpl.io/bzKvKQE 
![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/9760c34b-3496-4a6d-aea5-ae37162e66fa)

## Backend
[Add `KnownAddress` for all addresses in the `CreationTransaction`](https://github.com/gnosis/safe-client-gateway/issues/245)

## Spec
[Known Addresses V2](https://docs.google.com/document/d/1if8Sowyj1VhiVcRQojwmzMuDdo48dyndRBFrzIEa6zk/edit#)

liliya-soroka:
Questions:
1. What should be shown in the case non-known address?
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/eaf4b334-22a6-4202-8046-59bed55bac1e)

2. Will be the owner name taken from the imported owner key name?

liliya-soroka:
Verified.
The case when the owner key is imported and saved with the name will be re-tested during multiple owner key verification
app version 2.12.0 -2235

Issue #1156, opened 2020-12-21
[Android] Known Addresses V2: Safe Settings screens


sche:
Known addresses are displayed for:
- Safe owner address (not from backend)
Imported owner key (on web it comes from address book, on mobile imported owner key has a default name ‘Imported owner key’, and a custom name once “Multiple keys” feature is implemented)
- Safe version (from backend but not as AddressInfo)
- Fallback handler (from backend)
- Enabled modules (from backend)

## Design 
[Safe Settings Android (Zeplin)](https://zpl.io/bAe0e7n)
[Advanced Safe Settings Android (Zeplin)](https://zpl.io/bzKvKq7) 
![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/6c899727-436c-4fc8-a8a7-7ffeb4536b27) ![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/3b79723b-2bb4-4392-bbf9-a11869cd0c28)

## Backend
[`SafeInfo` endpoint in the gateway including`AddressInfo`](https://github.com/gnosis/safe-client-gateway/issues/242)

biafra23:
@tschubotz Is it ok, if we skip the `Safe owner address` in this version? I think the same as for confirmations applies. We will not have names in the backend for these. We can certainly show local owner names when we have multiple owners.

liliya-soroka:
Reopened:
1. Known address is missed for "Safe version" on the safe settings page 
Steps:
1. staging, mainnet , safe=alice.eth
2. Check Safe creation ( you will see the Mastercopy 1.0.0 name) 
3. Go to the safe settings
Current result: 1.0.0 hardcode is displayed instead of known address

elgatovital:
@liliya-soroka I don't understand what's wrong here. We show version and not name at this place.

liliya-soroka:
Verified
app version 2.13.0 - 2315rc

PR #1321, opened 2021-03-17
Known Addresses V2: Safe Settings & Tx Creation screens

elgatovital:
Handles #1156
Handles #1296

- Use client gw for safe info
- Use exteded address info from client gw to set module & fallback handler names & images
- Set names for owners in safe settings for locally stored owners
- Use & display name & logo returned by backed for creator, factory, and implementation on tx creation screen

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/settings/safe/SafeSettingsFragment.kt
```diff
@@ -106,25 +108,39 @@ class SafeSettingsFragment : BaseViewBindingFragment<FragmentSettingsSafeBinding
...
+    private fun ownerView(owner: Solidity.Address, localOwners: List<Owner>): View {
+
+        val localOwner = localOwners.find { it.address == owner }
+
+        return if (localOwner != null) {
+            NamedAddressItem(requireContext()).apply {
+                background = ContextCompat.getDrawable(requireContext(), R.drawable.background_selectable_white)
+                layoutParams = LinearLayout.LayoutParams(LinearLayout.LayoutParams.MATCH_PARENT, resources.getDimension(R.dimen.item_address).toInt())
+                address = owner
+                name = localOwner.name
```
biafra23, line 134: ATM this seems to be the null for the imported owner. Can we show "Imported owner key" instead? Until we have multiple owner names?

biafra23: Right now it looks like this: 
image

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionView.kt
```diff
@@ -120,8 +120,11 @@ sealed class TransactionView(
         @ColorRes val statusColorRes: Int,
         val dateTimeText: String,
         val creator: String,
+        val creatorInfo: AddressInfoData.Remote,
         val implementation: String?,
+        val implementationInfo: AddressInfoData.Remote,
```
biafra23, line 125: This is mandatory?

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewHolderFactory.kt
```diff
@@ -415,8 +415,14 @@ private fun navigateToCreationDetails(view: View, details: TransactionView.Creat
                 statusTextRes = details.statusText,
                 dateTimeText = details.dateTimeText,
                 implementation = details.implementation,
+                implementationName = details.implementationInfo.name,
```
biafra23, line 418: Since `implementationInfo` is optional (on the backend side). Should we have a fail safe String to show here? Like "Unknown" for example? Same for the other *Infos.

app/src/main/java/io/gnosis/safe/ui/settings/safe/SafeSettingsFragment.kt
elgatovital: yes, this is how it is actually defined in spec. Thank you for pointing it out

biafra23 (approved):
LGTM (pending the "unknown owner" label)

Issue #1339, opened 2021-03-25
Unit tests for Passcode ViewModels


biafra23:

liliya-soroka:
closed as no QA required

PR #1343, opened 2021-03-28
Unit tests for passcode

biafra23:
Handles #1339

- Merges two ViewModels into one to reduce duplication
- Tests for all PasscodeViewModel functions
- should contain all fixes from release branch
   
@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/PasscodeViewModel.kt
```diff
@@ -37,28 +38,28 @@ class PasscodeViewModel
...
     fun onForgotPasscode() {
         safeLaunch {
-
-            encryptionManager.removePassword()
-            encryptionManager.lock()
-            settingsHandler.usePasscode = false
-
             credentialsRepository.owners().forEach {
                 credentialsRepository.removeOwner(it)
+                notificationRepository.unregisterOwner()
```
elgatovital, line 43: we can call unregisterOwner one time outside the loop. It is better to rename it to notificationRepository.unregisterOwners()

elgatovital: yes

elgatovital (approved):

PR #1348, opened 2021-03-29
Handle back button correctly in change password flow.

biafra23:
Handles #

- Override back button to skip flow.

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/ChangeCreatePasscodeFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/ChangeCreatePasscodeFragment.kt
@@ -5,6 +5,7 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.view.inputmethod.EditorInfo
+import androidx.activity.OnBackPressedCallback
 import androidx.core.widget.doOnTextChanged
 import androidx.navigation.fragment.findNavController
 import androidx.navigation.fragment.navArgs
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/ChangeCreatePasscodeFragment.kt
```diff
@@ -69,7 +75,7 @@ class ChangeCreatePasscodeFragment : BaseViewBindingFragment<FragmentPasscodeBin
         }
     }
 
-    private fun FragmentPasscodeBinding.skipPasscodeSetup() {
+    private fun skipPasscodeSetup() {
 
         findNavController().popBackStack(R.id.changePasscodeFragment, true)
         findNavController().currentBackStackEntry?.savedStateHandle?.set(SafeOverviewBaseFragment.OWNER_IMPORT_RESULT, false)
```
elgatovital, line 81: why do we have this flag about owner import in change create fragment?

biafra23: It is only necessary to set this to false when coming from OwnerSelectionFragment. So it can be deleted here.

biafra23: There is more cleanup coming with the tests.

elgatovital (approved):

Issue #1349, opened 2021-03-29
[Android] No keyboard when forgot password is cancelled on enter passcode screen/change passcode screen


liliya-soroka:
**Describe the bug**
No keyboard when forgot password is cancelled on enter passcode screen/change passcode screen

**To Reproduce**
1. Import owner key and set up passcode
2. Go to the change passcode or confirm transaction
3. Click "forgot passcode" link on enter/change passcode
4. Cancel forgot passcode 
**Current result** No keyboard on enter passcode screen. The user has to cancel the process of passcode entering and restart the enter/change passcode flow

**Expected behavior**
The keyboard should be used each time when the passcode entering is required

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/517473de-386a-47e8-9db7-aab7be570518)

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.13.0 -2337rc

**Environment**
staging, rinkeby 
**Additional context**

PR #1352, opened 2021-03-30
Hide keyboard only on confirmation

biafra23:
Handles #1349 

- Hide keyboard only when dialog is confirmed
-
-

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/DisablePasscodeFragment.kt
```diff
@@ -99,6 +98,7 @@ class DisablePasscodeFragment : BaseViewBindingFragment<FragmentPasscodeBinding>
                     confirmColor = R.color.error
                 ) {
                     viewModel.onForgotPasscode()
+                    input.hideSoftKeyboard()
```
elgatovital, line 101: @biafra23 we could hide keyboard in onStop(). I think this will be more robust.

elgatovital (approved):

Issue #1346, opened 2021-03-29
[Android] Confirm button doesn't disappear after the tx is confirmed on tx details page


liliya-soroka:
**Describe the bug**
Confirm button doesn't disappear after the tx is confirmed on tx details page 
Note: it was already implemented on the back-end and ios, so It's expected that no confirm button if the tx was just confirmed

**To Reproduce**
1. Import owner key
2. Create tx that required confirmation from the imported owner
3. Go to the Queue 
4. Open transaction details for the tx that requires confirmation from the imported owner
5. Click Confirm button
Current result:  Tx details is updated with the info about the imported owner confirmation on tx details , confirm message appears but the Confirm button is still available

**Expected behavior**
The  confirm button should disappear from the tx details page as soon as the tx is confirmed by the imported owner

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/9be3f893-3359-4636-872a-a0a6ed5d5bb7)

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.13.0 - 2315rc

**Environment**
staging, rinkeby
**Additional context**

biafra23:
Should it really disappear or be disabled?

liliya-soroka:
In the case listed above : the buttons should be disabled

liliya-soroka:
Reopened:
safe = lilltestsafe.eth , staging, rinkeby ( the same on prod mainnet)
1. Import owner ( PR - ca30cbabffad80de6a0049be5eae6197b7c4fef98d5304436cd738e4e209dc73) and set up passcode
2. Go to the Queue
3. Open original tx in Needs execution status
Current result: "Confirm" button is displayed for the tx in "Needs execution" status
Expected result: Txs in the Needs execution status should not have any Reject or Confirm button

app version 2.13.0 - 2377rc

liliya-soroka:
Verified
app version 2.13.0 -2381rc , but another additional case with Confirm button exists and reported separately

PR #1364, opened 2021-03-31
Fix confirm button still available after tx confirmation

elgatovital:
Handles #1346

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
@@ -197,7 +197,7 @@ class TransactionDetailsViewModel
                 var rejectable = false
                 var safeOwner = false
                 if (executionInfo is DetailedExecutionInfo.MultisigExecutionDetails) {
-                    awaitingConfirm = isAwaitingOwnerConfirmation(executionInfo, txDetails!!.txStatus, owners)
+                    awaitingConfirm = false
                     rejectable = canBeRejectedFromDevice(executionInfo, txDetails!!.txStatus, owners)
                     safeOwner = isOwner(executionInfo, owners)
                 }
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
```diff
@@ -197,7 +197,7 @@ class TransactionDetailsViewModel
                 var rejectable = false
                 var safeOwner = false
                 if (executionInfo is DetailedExecutionInfo.MultisigExecutionDetails) {
-                    awaitingConfirm = isAwaitingOwnerConfirmation(executionInfo, txDetails!!.txStatus, owners)
+                    awaitingConfirm = false
```
biafra23, line 200: Why is it always false and why is isAwaitingOwnerConfirmation() not working correctly here?

elgatovital: because we just submitted a confirmation

elgatovital: we haven't updated our txDetails and that was the reason isAwaitingOwnerConfirmation was not working correctly

biafra23 (approved):

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
```diff
@@ -190,6 +190,7 @@ class TransactionDetailsViewModel
                     credentialsRepository.signWithOwner(owners[0], executionInfo.safeTxHash.hexToByteArray())
                 )
             }.onSuccess {
+                txDetails = it
```
biafra23, line 193: This doesn't seem to be enough. The button is still there until I pull to refresh.

Issue #1330, opened 2021-03-17
[Android] UI issues with App icon in case of safe app long name


liliya-soroka:
**Describe the bug**
If the safe app has long have there are UI issues on txs list and tx details pages with displaying App icon

**To Reproduce**
1. Go to safe=0x8675B754342754A30A2AeF474D114d8460bca19b on prod mainnet
2. Open txs list and search for the "Balancer Pool Management" safe app transaction in the list
**Current result:** 
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/20789eeb-53d0-40da-8cb0-83840486c1c8)
3. Open "Balancer Pool Management" safe app transaction
**Current result:**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/8cfaa9db-bc08-40ec-b5a2-a3b126a0931a)
**Expected behavior**
No UI issues

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.12.0 - 2191rc

**Environment**
prod mainnet
**Additional context**

liliya-soroka:
Looks better but still not seen fully:
Txs list
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/1caf34a0-7b5a-4d36-836c-49adf64f6a32)

The issue is not fixed on tx details 
the same screenshot as in the issue description 

app version 2.13.0 -2375rc

biafra23:
@liliya-soroka What Font size (Settings -> Display -> Advanced -> Font size) have you configured on your device?

liliya-soroka:
@biafra23 , I have never edited them - ![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/f5866a56-3e2b-4404-91f8-fe48a7de0bb2)
The view in the lates version:

![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/dd9454db-280b-46a9-a987-36096c92a603)
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/e93784b6-1136-4675-8dc3-6a47ddb16256)

Does it look better for you ?

app version 2.14.0 - 2482

biafra23:
@liliya-soroka 
What resolution and density does your phone have? 
I am trying to create an emulator, that has the same properties.

liliya-soroka:
@biafra23 
[...]

PR #1369, opened 2021-03-31
App label in list and details

biafra23:
Handles #1330

-
-
-

@gnosis/mobile-devs


biafra23:
Since the bug is minor. These change should go to master and not release.

Issue #1259, opened 2021-02-15
[Android] Change the "Import Owner Key - Choose Key" screen


biafra23:
# Enter Key screen
- [x] If user enters private key that was already imported, then show the error:
  - [x]  **Error ID: 1111** according to specification: https://docs.google.com/document/d/1foQ9DSMgO2iN-6u4JsJIDJWB7mUHX-D8ApZWvx6oKAo/edit#bookmark=id.iuuwrj8172ax 

![image.png](https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/98ff9531-8dca-4baf-b55f-418367be128f)

# Choose Key screen
- [x] Disable keys that are already imported
- [x] Show key names for already imported keys
- [x] If the "default" key already imported, then show it as disabled and do NOT select any key. User has to "show more derived keys" and select the next available key.
- [x] Import button should be disabled if no key is selected.

[Zeplin Link](https://zpl.io/blGKO5v)

![Screen Shot 2021-02-15 at 14.44.32.png](https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/403cd8f7-170c-4d53-8cd6-254ab154597a)

liliya-soroka:
Verified
app version 2.14.0- 2482

PR #1382, opened 2021-04-13
Feature/1259 disable imported keys

biafra23:
Handles #1259

- Error when importin private key which is already imported 
- Disable selection for already imported derived keys
-

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/settings/owner/selection/DerivedOwnerListAdapter.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/settings/owner/selection/DerivedOwnerListAdapter.kt
@@ -8,15 +8,19 @@ import androidx.paging.PagingDataAdapter
 import androidx.recyclerview.widget.DiffUtil
 import androidx.recyclerview.widget.RecyclerView
 import io.gnosis.safe.databinding.ItemDefaultOwnerKeyBinding
+import io.gnosis.safe.databinding.ItemOwnerSelectionDisabledOwnerBinding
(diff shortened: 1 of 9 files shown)


app/src/main/java/io/gnosis/safe/ui/settings/owner/selection/DerivedOwnerListAdapter.kt
```diff
@@ -65,40 +84,82 @@ class DerivedOwnerListAdapter() : PagingDataAdapter<Solidity.Address, DerivedOwn
...
+        DISABLED_OWNER,
+        DISABLED_DEFAULT_OWNER
     }
 
     abstract class BaseOwnerViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
-        abstract fun bind(address: Solidity.Address, position: Int)
+        abstract fun bind(address: OwnerHolder, position: Int)
+    }
+
+    inner class DisabledOwnerViewHolder(private val binding: ItemOwnerSelectionDisabledOwnerBinding) : BaseOwnerViewHolder(binding.root) {
```
elgatovital, line 117: it seems like a redundancy to me. Can OwnerViewHolder handle both disabled and enabled state?

biafra23: As I said I had problems getting the two different (with and without label) layouts into one. That's why I decided to have two layouts.

elgatovital (approved):

Issue #1264, opened 2021-02-15
[Android] Show Owner Key Names 


biafra23:
# Safe Settings
- Owner list of addresses should show the name of the keys that are imported

[Zeplin Link](https://zpl.io/VQ6PnB4)

![image.png](https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/38d61de9-7942-4943-9f9f-d51f19a865b8)

# Transaction Details
- Transaction timeline should show key names of the imported keys

[Zeplin Link](https://zpl.io/2vwp6P7)

![image.png](https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/38780084-ccf6-40bb-9fca-75c190ce7acd)

liliya-soroka:
Verified
app version 2.13.0 - 2452

PR #1385, opened 2021-04-14
Show owner key names in safe settings & tx confirmations

elgatovital:
Handles #1264

@gnosis/mobile-devs


biafra23:
Quite often the creator address in the creation tx is one of the owners. Would it make sense to show the name too?

biafra23 (approved):

elgatovital:
> Quite often the creator address in the creation tx is one of the owners. Would it make sense to show the name too?

yes, this would be in sync with what we do / how we handle known addresses in other places

biafra23 (approved):

Issue #1219, opened 2021-01-21
[Android] Signing V2: Add "read only" label to safes


biafra23:
### Expected behavior
Show `read only` label in app bar next to loaded safe name if the imported owner key is not an owner for this safe or no owner key has been imported

## Design

[Android (Zeplin)](https://zpl.io/anxRkdl)

![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/6cd1d492-9d24-4c7b-9709-cfc2e2464f6f)

DmitryBespalov:
@tschubotz on iOS currently the "Read-only" label is shown until we get backend response and can confirm that signing keys are owners. Then the label hides.

What happens is that for a small moment (1 sec or less) the Read-only is shown and then hidden, so there're blinking. 

Should we have done it differently? I.e. don't show it by default and only show read-only when we receive the backend response.

tschubotz:
> What happens is that for a small moment (1 sec or less) the Read-only is shown and then hidden, so there're blinking.

Does this happen only once or e.g. everyime the app is opened?

DmitryBespalov:
it would happen often

elgatovital:
@tschubotz I suggest to revert this logic. Do not show anything and show read only after backend response is received and verified. Better not to show anything than to show false information

tschubotz:
> Better not to show anything than to show false information

Ya, seems better than the opposite 😬

elgatovital:
@DmitryBespalov do you need a ticket for this adjustment?

DmitryBespalov:
@elgatovital yes please

liliya-soroka:
Verified
app version 2.14.0 - 2501

PR #1392, opened 2021-04-15
Check if safe can confirm/reject tx and show read-only label in toolbar

elgatovital:
Handles #1219

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
@@ -176,7 +176,7 @@ class StartActivity : BaseActivity(), SafeOverviewNavigationHandler {
             safeName.post {
                 val readOnlySpace = readOnly.measuredWidth + readOnly.marginLeft + readOnly.marginRight
                 if (safeNameLength > safeAddress.measuredWidth - readOnlySpace) {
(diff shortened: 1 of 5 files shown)


biafra23:
image
- Should the read only be next to the name of the safe?
- Should we ellipsize instead of breaking it into two lines?

app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
```diff
@@ -164,8 +166,27 @@ class StartActivity : BaseActivity(), SafeOverviewNavigationHandler {
...
+        with(toolbarBinding) {
+            val bounds = Rect()
+            safeName.paint.getTextBounds(safeName.text.toString(), 0, safeName.text.length, bounds)
+            val safeNameLength = bounds.right - bounds.left
+
+            // wait till views are measured
+            safeName.post {
+                val readOnlySpace = readOnly.measuredWidth + readOnly.marginLeft + readOnly.marginRight
+                if (safeNameLength > safeAddress.measuredWidth - readOnlySpace) {
+                    safeName.width = safeNameLength - (safeName.marginLeft + safeName.marginRight) - readOnlySpace
```
biafra23, line 179: There might be something wrong with the calculation:
image

biafra23 (approved):
Looking good! 👍🏻

Issue #1386, opened 2021-04-14
[Android]Confirm and reject button has incorrect status( enabled) if more confirmations are required but no more imported owners


liliya-soroka:
**Describe the bug**
The confirmation button is not disabled after tx confirmation and the reject button status is changed from disabled to enabled after tx confirmation if more confirmations are required but no more imported owner keys
The confirm button status is not updated to the correct if more confirmations are required but there is no more imported owners

**To Reproduce**
1st case :Steps to reproduce the behavior:
Precondition:
0. Sfe with 3 or more required confirmation
1. Create tx and reject it as owner 1 
2. import owner 2 key to the mobile app

Steps
1. Go to Queue
2. Open transaction details for tx from the precondition
Current result: reject button is disabled and Confirm is enabled

3. Click confirm and select owner to confirm tx
Current result: The tx is confirmed message appears but the Confirm button is not disabled

4. Click Confirm again
Current result: Select owner to confirm screen is opened. The list is empty

5. Go back
Current result: Confirm button disappears but Reject button is enabled instead of disabled. The user can reject tx and his signature will be saved to the cancellation tx

2nd case:
Precondition"
1. safe with 3 required confirmation
2. owner 1 creates transaction and rejects the transaction
3. Owner 2 imported to the mobile app

Steps:
1. go to the cancellation tx details and click Confirm button
Current result: The signing from the owner 2 is saved but the Confirm button is still available.

**Expected behavior**
The correct state of the buttons should be used
1. both buttons should be disabled if the tx was confirmed and cancellation tx exists when more confirmations are required but no more imported owners
2. Disabled state of the reject button should not be changed to enabled if cancellation tx exists

**Video**
https://drive.google.com/file/d/1OWmjgipdXKHo1FnIG2J7AKB4Uiknz-Jr/view?usp=sharing

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.14.0 - 2417rc

**Environment**
staging, rinkeby
**Additional context**

tschubotz:
hmm, tough one. `critical` or `major`. I'd lean towards marking as `major`, i.e. not blocking the current release since while being inconvenient, nothing bad happens and it's quite a special case. What do you think? @liliya-soroka @gnosis/android

If not in the current release, I think this should be fixed in the "New York" sprint then.

liliya-soroka:
[...]

Issue #1371, opened 2021-04-01
[Android]Confirm button is displayed for already confirmed tx if the cancellation tx is not confirmed by the imported owner


liliya-soroka:
**Describe the bug**
Confirm button is displayed for already confirmed tx if the cancellation tx is not confirmed by the imported owner.
So it's displayed even when the tx is in "needs execution" status. After Confirmation is clicked again the successfully confirmed message appears but the button doesn't disappear

**To Reproduce**
Precondition:
Safe with 2 required confirmation
1. Import one of the owner
2. Another owner creates tx on web
3. Imported owner confirmed original tx 
Current result: The tx status is changed to Needs execution and No buttons on UI

4. Go to the web as the tx creator and rejects transaction
5. Open original tx on mobile
Current result: Confirm button is displayed in both original and cancellation tx

**Expected behavior**
No confirm button in original or cancellation tx if it's already confirmed by the imported owner

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/b6795339-7ef8-406a-a50f-0e0faed7cc2d)

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.13.0 - 2381rc

**Environment**
prod rinkeby
**Additional context**

biafra23:
@tschubotz @liliya-soroka 
> Expected behaviour:
> No confirm button in original or cancellation tx if it's already confirmed by the imported owner

Is that really the expected behaviour? An owner should not be able to change their mind to rejecting after they confirmed?

biafra23:
I think in this case the rejection tx should still be confirmable. But the original tx should indeed have no button.

tschubotz:
I think because of 

> 3. Imported owner confirmed original tx

The imported owner did already confirm the original tx before. That's why Liliya was expecting no confirm button in the original tx.



> An owner should not be able to change their mind to rejecting after they confirmed?

Agree, an owner should be able to change their mind, i.e. be able to create a cancellation tx after previously confirm the original tx.

liliya-soroka:
Verified
app version 2.14.0 -2479

PR #1393, opened 2021-04-16
ButtonStateHelper rewrite

biafra23:
Handles #1386, #1371

- Complete rewrite of ButtonStateHelper
- Tests adapted to scenarios in https://docs.google.com/document/d/13GirjSwieQaqUPz-8HXzJP_rfgmobSp7E4TpbnLlb0M/edit#bookmark=id.3cvx2oq4xmgz

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/details/ButtonStateHelper.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/details/ButtonStateHelper.kt
@@ -1,73 +1,64 @@
 package io.gnosis.safe.ui.transactions.details
 
 class ButtonStateHelper(
+    val canSign: Boolean,
+    val hasOwnerKey: Boolean,
     val hasBeenRejected: Boolean,
-    val needsYourConfirmation: Boolean,
     val isRejection: Boolean,
-    val needsExecution: Boolean,
-    val canReject: Boolean,
-    val isOwner: Boolean,
-    val completed: Boolean
+    val awaitingConfirmations: Boolean,
+    val awaitingExecution: Boolean
 ) {
-
-    fun rejectButtonIsVisible(): Boolean =
+    fun confirmationButtonIsVisible(): Boolean =
         when {
-            !isOwner || completed -> {
-                false
-            }
-            isRejection || !canReject -> {
-                false
-            }
-            !isRejection && !needsYourConfirmation && hasBeenRejected && needsExecution -> {
-                false
-            }
-            else -> true
+            isRejection && awaitingConfirmations && canSign -> true
+            isRejection && awaitingConfirmations && hasOwnerKey && !canSign -> true
 
-        }
+            !isRejection && awaitingConfirmations && canSign -> true
+            !isRejection && awaitingConfirmations && hasOwnerKey && !canSign -> true
 
-    fun confirmButtonIsVisible(): Boolean =
-        when {
-            !isOwner || completed -> {
-                false
-            }
-            needsYourConfirmation -> {
-                true
-            }
-            !isRejection && !hasBeenRejected && !needsExecution -> {
-                true
-            }
-            needsExecution && !canReject && !needsYourConfirmation-> {
-                false
-            }
-            !isRejection && !needsYourConfirmation && hasBeenRejected && needsExecution -> {
-                false
-            }
+            !isRejection && awaitingExecution && hasOwnerKey && hasBeenRejected -> false
+
             else -> false
         }
 
-    fun rejectButtonIsEnabled(): Boolean =
-        when {
-            !rejectButtonIsVisible() -> {
-                false
-            }
-            !isRejection && needsYourConfirmation && hasBeenRejected && !needsExecution -> {
-                false
-            }
-            else -> true
-        }
+    fun confirmationButtonIsEnabled(): Boolean = when {
+        !confirmationButtonIsVisible() -> false
+        isRejection && awaitingConfirmations && canSign -> true
+        isRejection && awaitingConfirmations && hasOwnerKey && !canSign -> false
 
-    fun confirmButtonIsEnabled(): Boolean = when {
-        !confirmButtonIsVisible() -> {
-            false
-        }
-        !needsYourConfirmation && !needsExecution -> {
-            false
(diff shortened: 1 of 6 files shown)


app/src/main/java/io/gnosis/safe/ui/transactions/details/ButtonStateHelper.kt
```diff
@@ -1,73 +1,64 @@
...
-                false
-            }
-            else -> true
+            isRejection && awaitingConfirmations && canSign -> true
+            isRejection && awaitingConfirmations && hasOwnerKey && !canSign -> true
+
+            !isRejection && awaitingConfirmations && canSign -> true
+            !isRejection && awaitingConfirmations && hasOwnerKey && !canSign -> true
+
+            !isRejection && awaitingExecution && hasOwnerKey && hasBeenRejected -> false
```
jpalvarezl, line 19: Seems a bit contrived to check for the predicates and return a hard-coded boolean value. Not sure if it would be better to return the value of the predicate directly (or its negation)  :thinking: Seems to be the style with which it was previously written, so I suppose changing it now would be out of scope.

jpalvarezl: As a suggestion for future improvements, if you want, I think that enums or sealed classes would be a good use case to map into and then depending on the case, return `true` or `false` for visibility.

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
```diff
@@ -291,7 +291,7 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
                     txStatus.setStatus(
                         txType.titleRes,
                         txType.iconRes,
-                        getStringResForStatus(txDetails.txStatus, needsYourConfirmation),
+                        getStringResForStatus(txDetails.txStatus, canSign && awaitingConfirmations),
```
jpalvarezl, line 294: Maybe for added readability you could assign this to `needsYourConfirmation` ? I am assuming that `canSign` checks that you are safe owner that hasn't signed yet and `awaitingConfirmation` just needs the confirmation? Seems like the old flag, just split up.

jpalvarezl: it seems to me that you pushed down the calculation of this flag to this method only, right?

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
```diff
@@ -191,25 +183,19 @@ class TransactionDetailsViewModel
                 )
             }.onSuccess {
                 txDetails = it
+                val newExecutionInfo = txDetails?.detailedExecutionInfo as? DetailedExecutionInfo.MultisigExecutionDetails
+                    ?: throw MissingCorrectExecutionDetailsException
```
jpalvarezl, line 187: So in case you get `ExecutionDetails` that aren't from the type you are expecting, it throws? Couldn't we leverage types better here? Or, if we are in branch of code in which we know for sure the type, just simply do a hard cast?

app/src/test/java/io/gnosis/safe/ui/transactions/details/ButtonStateHelperTest.kt
```diff
@@ -6,299 +6,213 @@ import org.junit.Test
 
 class ButtonStateHelperTest {
 
-    // Rejections For numbering, see: https://docs.google.com/document/d/1dlhVAv9WiD1EPi1LuZI6opk17ivD2lpMbwbWP4aArIY/edit
+    // For numbering, see: https://docs.google.com/document/d/13GirjSwieQaqUPz-8HXzJP_rfgmobSp7E4TpbnLlb0M/edit#
```
jpalvarezl, line 9: :sweat_smile: hopefully this document Id doesn't change in the future again :)

app/src/test/java/io/gnosis/safe/ui/transactions/details/ButtonStateHelperTest.kt
```diff
@@ -6,299 +6,213 @@ import org.junit.Test
...
-                completed = false
+                awaitingExecution = false
             )
 
-        assertFalse(buttonStateHelper.confirmButtonIsVisible())
-        assertFalse(buttonStateHelper.rejectButtonIsVisible())
-        assertFalse(buttonStateHelper.buttonContainerIsVisible())
+        assertTrue(buttonStateHelper.confirmationButtonIsVisible())
+        assertTrue(buttonStateHelper.confirmationButtonIsEnabled())
+        assertTrue(buttonStateHelper.buttonContainerIsVisible())
```
jpalvarezl, line 25: How come we asserted for the direct opposite before? The button container visibility should not have changed? Although asserting to `true` makes more sense than `false`.

jpalvarezl (approved):

app/src/test/java/io/gnosis/safe/ui/transactions/details/ButtonStateHelperTest.kt
biafra23: There is no relation between the old test class and the new test class. I think this being opposite is just a coincidence.

app/src/main/java/io/gnosis/safe/ui/transactions/details/ButtonStateHelper.kt
biafra23: I tried optimizing this. But it became less readable/changeable/debuggable. I don't think I understand how you would make this better. Can you give and example?

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
biafra23: Yeah. I was lazy. Looking into it once more.

elgatovital (approved):

Issue #1337, opened 2021-03-23
[Android] Mess on the safe settings screen when no internet connection


liliya-soroka:
**Describe the bug**
Partly loaded data + Data can't be loaded are displayed together on the safe settings screen when nop internet connection 

**To Reproduce**
1. Open app on assets
2. Turn internet connection off
2. Go to the settings 
3. Switch from safe settings to app settings 
4. Open Fiat currency page
5. Go back to safe settings
**Current result** 
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/cbc81a61-6e7b-4719-8f8a-0aff289eb032)
**Expected behavior**
Error message + data can not be loaded should be displayed

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.12.0 - 2235

**Environment**
staging, rinkeby
**Additional context**

liliya-soroka:
The app is crashed if the interner connection off and the user tries to open any page
Steps:
1. run app on balance page
2. turn internet connection off
3. go to the transaction list of safe settings
Current result: the app is crahsed

app version 2.13.0 - 2452

liliya-soroka:
Change to Majour for 2.15.0 release because of crash

liliya-soroka:
Verified
app version 2.14.0 - 2490

PR #1399, opened 2021-04-20
Fail silently if read-only check cannot be performed

elgatovital:
Handles #1337

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
@@ -196,6 +196,8 @@ class StartActivity : BaseActivity(), SafeOverviewNavigationHandler {
                     val localOwners = creadentialsRepository.owners().map { it.address }.toSet()
                     toolbarBinding.readOnly.visible(safeOwners.intersect(localOwners).isEmpty(), View.INVISIBLE)
                 }
+            }.onFailure {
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
```diff
@@ -188,11 +188,14 @@ class StartActivity : BaseActivity(), SafeOverviewNavigationHandler {
...
-                val localOwners = creadentialsRepository.owners().map { it.address }.toSet()
-                toolbarBinding.readOnly.visible(safeOwners.intersect(localOwners).isEmpty(), View.INVISIBLE)
+            // fail silently if safe info cannot be loaded
+            kotlin.runCatching {
+                val activeSafe = safeRepository.getActiveSafe()
+                activeSafe?.let {
+                    val safeOwners = safeRepository.getSafeInfo(it.address).owners.map { it.value }.toSet()
+                    val localOwners = creadentialsRepository.owners().map { it.address }.toSet()
+                    toolbarBinding.readOnly.visible(safeOwners.intersect(localOwners).isEmpty(), View.INVISIBLE)
+                }
```
biafra23, line 198: Can we maybe log the exception to Firebase in `onFailure {}`
So we find out, when this happens often.

biafra23 (approved):

Issue #1290, opened 2021-02-23
Confirmation notification: No unregistration for deleted owner key 


liliya-soroka:
**Describe the bug**
The confirmation notification is received when the owner key was deleted and did not receive when a new owner key was added after it

**To Reproduce**
1. Go to the app 
2. Import owner key using seed phrase
3. Check that the "Confirmation required" notification is received
4. remove the owner key
5. reject any existed in Queue list
Current result: "A confirmation required" notification is received after owner key removing 

6. Import new owner key ( Note I used the same seed phrase as in step 2 but selected another owner)
7. reject any existed tx in Queue list as removed owner
Current result: "A confirmation required" notification is not received. I have a feeling that the removed owner is not removed completely 

**Expected behavior**
"A confirmation required" notification should not be received if the owner key was deleted
"A confirmation required" notification should be received if the new owner is added and a new tx is created by the owner whose key was deleted from the mobile app.

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.9.0 -21111

**Environment**
staging, rinkeby
**Additional context**

tschubotz:
Marking this as minor since this is for sure a bug, however the owner of the device had a valid Safe owner imported before, so it's not terribly bad that this happens.

biafra23:
It seems registration fails with a 400 error:
```
2021-04-20 10:03:11.424 5791-5916/io.gnosis.safe.debug I/okhttp.OkHttpClient: --> POST https://safe-transaction.staging.gnosisdev.com/api/v1/notifications/devices/
2021-04-20 10:03:11.424 5791-5916/io.gnosis.safe.debug I/okhttp.OkHttpClient: Content-Type: application/json; charset=UTF-8
2021-04-20 10:03:11.425 5791-5916/io.gnosis.safe.debug I/okhttp.OkHttpClient: Content-Length: 605
[...]

PR #1400, opened 2021-04-20
Notification registration

biafra23:
Handles #1290

- Store uuid
- Reset Firebase Messaging token if registration fails

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/notifications/NotificationRepository.kt
```diff
@@ -224,9 +228,17 @@ class NotificationRepository(
             }
             .onFailure {
                 registrationUpdateFailed = true
+                resetFirebaseToken()
```
biafra23, line 231: This will loop, if the new token doesn't fix the error. For example if we try to register an a key that's not an owner.

app/src/main/java/io/gnosis/safe/ErrorHandler.kt
```diff
@@ -53,6 +54,7 @@ sealed class Error(
 
     object Error104 : Error(104, 104, R.string.error_network_unknown_host_reason, R.string.error_network_unknown_host_fix)
     object Error400 : Error(400, 400, R.string.error_network_request_reason, R.string.error_network_request_fix)
+    object ErrorCloudMessagingTokenIsLinkedToAnotherDevice : Error(400, 400, R.string.error_network_request_reason, R.string.error_network_request_fix)
```
elgatovital, line 57: could you please define error in https://docs.google.com/document/d/1foQ9DSMgO2iN-6u4JsJIDJWB7mUHX-D8ApZWvx6oKAo/edit?ts=5ff5e055# with separate id and use our naming scheme (Error<Id>) . Please check with ios if this should be an android error (17xx) or if ios is also affected. Then it would be common client error (11xx)

app/src/main/java/io/gnosis/safe/ErrorHandler.kt
```diff
@@ -91,12 +93,23 @@ sealed class Error(
...
 
-    object ErrorUdUnsupportedDomain: Error(6357, null, R.string.error_client_UD_invalid_domain_reason, R.string.error_client_UD_invalid_domain_fix)
-    object ErrorUdUnregistered: Error(6358, null, R.string.error_client_UD_name_not_registered_reason, R.string.error_client_UD_name_not_registered_fix)
-    object ErrorUdRecordNotFound: Error(6359, null, R.string.error_client_UD_record_not_found_reason, R.string.error_client_UD_record_not_found_fix)
-    object ErrorUdUnspecifiedResolver: Error(6360, null, R.string.error_client_UD_domain_not_configured_reason, R.string.error_client_UD_domain_not_configured_fix)
-    object ErrorUdBlockhainDown: Error(6361, null, R.string.error_client_UD_blockchain_provider_is_not_accessible_reason, R.string.error_client_UD_blockchain_provider_is_not_accessible_fix)
-    object ErrorUdUnknownCurrency: Error(6357, null, R.string.error_client_UD_currency_not_found_reason, R.string.error_client_UD_currency_not_found_fix)
+    object ErrorUdUnsupportedDomain : Error(6357, null, R.string.error_client_UD_invalid_domain_reason, R.string.error_client_UD_invalid_domain_fix)
```
elgatovital, line 96: Error 6357 and all following errors are not defined in error spec  https://docs.google.com/document/d/1foQ9DSMgO2iN-6u4JsJIDJWB7mUHX-D8ApZWvx6oKAo/edit?ts=5ff5e055#. Could you please add them? Also we should stick to our naming scheme.

app/src/main/java/io/gnosis/safe/ErrorHandler.kt
```diff
@@ -134,14 +147,19 @@ fun Throwable.toError(): Error =
         // Network-related errors
         this is HttpException -> {
             this.let {
+                val errorBodyString = it.response()?.errorBody()?.string()
                 when {
-                    this.code() == HttpCodes.BAD_REQUEST -> Error.Error400
+                    this.code() == HttpCodes.BAD_REQUEST -> return if (errorBodyString == "[\"Cloud messaging token is linked to another device\"]") {
```
elgatovital, line 152: Is it correct that this state is only possible if we don't save deviceUuid after registration? With assumption that correct deviceUuid is always saved, we might not have to deal with this error. Checking body string doesn't look very reliable to me.

app/src/main/java/io/gnosis/safe/notifications/NotificationRepository.kt
```diff
@@ -224,9 +229,24 @@ class NotificationRepository(
...
     }
 
+    private suspend fun handleCloudMessagingTokenIsLinkedToAnotherDeviceError(throwable: Throwable) {
+        val error = throwable.toError()
+        if (error == Error.ErrorCloudMessagingTokenIsLinkedToAnotherDevice) {
+            resetFirebaseToken()
+        }
+    }
+
+    private suspend fun resetFirebaseToken() {
```
elgatovital, line 243: :+1:

app/src/main/java/io/gnosis/safe/notifications/NotificationRepository.kt
```diff
@@ -224,9 +229,24 @@ class NotificationRepository(
...
+        kotlin.runCatching {
+            FirebaseInstanceId.getInstance().deleteInstanceId()
+            FirebaseInstanceId.getInstance().deleteToken(getCloudMessagingToken()!!, INSTANCE_ID_SCOPE)
+        }
+    }
+
     suspend fun unregister() {
         kotlin.runCatching {
             deviceUuid?.let {
```
elgatovital, line 252: we should delete deviceUuid after successful unregister request

app/src/main/java/io/gnosis/safe/notifications/NotificationRepository.kt
```diff
@@ -224,9 +229,24 @@ class NotificationRepository(
...
+    private suspend fun handleCloudMessagingTokenIsLinkedToAnotherDeviceError(throwable: Throwable) {
+        val error = throwable.toError()
+        if (error == Error.ErrorCloudMessagingTokenIsLinkedToAnotherDevice) {
+            resetFirebaseToken()
+        }
+    }
+
+    private suspend fun resetFirebaseToken() {
+        kotlin.runCatching {
+            FirebaseInstanceId.getInstance().deleteInstanceId()
```
elgatovital, line 245: suggestion to make it more legible:
```kotlin
with(FirebaseInstanceId.getInstance()){
  deleteInstanceId()
  deleteToken(getCloudMessagingToken()!!, INSTANCE_ID_SCOPE)
}
```

elgatovital: can we always assume token won't be null? Otherwise it will crash here

app/src/main/java/io/gnosis/safe/notifications/NotificationRepository.kt
```diff
@@ -260,7 +280,8 @@ class NotificationRepository(
     }
 
     private fun generateUUID(): String {
-        return UUID.randomUUID().toString().toLowerCase()
+        deviceUuid = UUID.randomUUID().toString().toLowerCase()
+        return deviceUuid!!
```
elgatovital, line 284: why do we need !! here?

app/src/main/java/io/gnosis/safe/ErrorHandler.kt
biafra23: I hesitated to add this because we're never going to show it to the user. Why have it in the list, if it is only internally used?

app/src/main/java/io/gnosis/safe/ErrorHandler.kt
biafra23: We will have to deal with this, because there are app instances out there that might have this problem. When users upgrade, it should be fixed for good. 

I am happy to check anything here, when we have a specification from the backend. I asked Uxio, and he said there is only the source code.

app/src/main/java/io/gnosis/safe/notifications/NotificationRepository.kt
biafra23: Because the type of `deviceUuid` is nullable.

app/src/main/java/io/gnosis/safe/ErrorHandler.kt
elgatovital: I think having it in spec would improve overview about errors that we handle on the client

app/src/main/java/io/gnosis/safe/notifications/NotificationRepository.kt
elgatovital: but not the generated string from UUID?

app/src/main/java/io/gnosis/safe/notifications/NotificationRepository.kt
biafra23: Solved

app/src/main/java/io/gnosis/safe/notifications/NotificationRepository.kt
biafra23: About `unregister()`. It is never called. Maybe we can delete it. Who should call it and under what circumstances?

biafra23: I think it should be called when the last safe is removed from the app. But then the app is quite useless.

elgatovital (approved):

Issue #1391, opened 2021-04-15
[Android] Adjust colors for confirm / reject / primary buttons


elgatovital:
- [x] Use enabled/disabled colors from designs
- [x] Check pressed state colors

liliya-soroka:
Verified
app version 2.15.0 - 2503rc

PR #1401, opened 2021-04-21
Adjust colors for confirm/reject/primary buttons

elgatovital:
Handles #1391

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/helpers/ConnectivityInfoProvider.kt
+++ b/app/src/main/java/io/gnosis/safe/helpers/ConnectivityInfoProvider.kt
@@ -12,12 +12,12 @@ class ConnectivityInfoProvider(private val connectivityManager: ConnectivityMana
     var offline: Boolean = true
 
     private val networkCallback = object : ConnectivityManager.NetworkCallback() {
-        override fun onAvailable(network: Network?) {
+        override fun onAvailable(network: Network) {
             super.onAvailable(network)
             offline = false
         }
 
-        override fun onLost(network: Network?) {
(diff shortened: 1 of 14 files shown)


app/src/main/java/io/gnosis/safe/helpers/ConnectivityInfoProvider.kt
```diff
@@ -12,12 +12,12 @@ class ConnectivityInfoProvider(private val connectivityManager: ConnectivityMana
     var offline: Boolean = true
 
     private val networkCallback = object : ConnectivityManager.NetworkCallback() {
-        override fun onAvailable(network: Network?) {
+        override fun onAvailable(network: Network) {
```
biafra23, line 15: This fixes another bug, right? Or was this not meant to be in this PR?

build.gradle
```diff
@@ -35,11 +35,11 @@ subprojects {
     afterEvaluate { project ->
         if (project.hasProperty("android")) {
             android {
-                compileSdkVersion 29
+                compileSdkVersion 30
```
biafra23, line 38: Sneaky!

biafra23 (approved):
This PR seems to fix more than only colors. If that was intentional. Then: 👍🏻

app/src/main/java/io/gnosis/safe/helpers/ConnectivityInfoProvider.kt
elgatovital: no, it was fixed in previous pr with try catch around read-only network request

Issue #1398, opened 2021-04-20
[Android]Crash on prod - DeserializationContext.java line 855


liliya-soroka:
**Describe the bug**
https://console.firebase.google.com/u/0/project/safe-firebase-mainnet/crashlytics/app/android:io.gnosis.safe/issues/8a08756bd6d34c37840b17de86c1bf97?time=last-seven-days&sessionEventKey=607DDFAA0121000121C51B64594D43DA_1531405338255356257

**To Reproduce**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/202891c6-9ec5-40c1-8c6f-82669fe2e101)

**Expected behavior**

**Screenshots**

**Environment**
prod, mainnet
**Additional context**

biafra23:
I contacted JohnnyJumper#6176 on Discord and asked him where I should put the stack traces.

biafra23:
```
Fatal Exception: java.lang.ExceptionInInitializerError
       at com.unstoppabledomains.config.network.NetworkConfigLoader.getContractAddress(NetworkConfigLoader.java:26)
       at com.unstoppabledomains.resolution.naming.service.CNS.<init>(CNS.java:32)
       at com.unstoppabledomains.resolution.Resolution$Builder$2.<init>(Resolution.java:283)
       at com.unstoppabledomains.resolution.Resolution$Builder.build(Resolution.java:282)
       at io.gnosis.safe.di.modules.RepositoryModule.providesDomainResolutionLibrary(RepositoryModule.java:59)
       at io.gnosis.safe.di.modules.RepositoryModule_ProvidesDomainResolutionLibraryFactory.proxyProvidesDomainResolutionLibrary(RepositoryModule_ProvidesDomainResolutionLibraryFactory.java:32)
       at io.gnosis.safe.di.modules.RepositoryModule_ProvidesDomainResolutionLibraryFactory.provideInstance(RepositoryModule_ProvidesDomainResolutionLibraryFactory.java:22)
       at io.gnosis.safe.di.modules.RepositoryModule_ProvidesDomainResolutionLibraryFactory.get(RepositoryModule_ProvidesDomainResolutionLibraryFactory.java:18)
       at io.gnosis.safe.di.modules.RepositoryModule_ProvidesDomainResolutionLibraryFactory.get(RepositoryModule_ProvidesDomainResolutionLibraryFactory.java:8)
       at dagger.internal.DoubleCheck.get(DoubleCheck.java:47)
[...]

PR #1408, opened 2021-04-28
Don't crash in initializer

biafra23:
Handles #1398

- Capture exception in initializing unstoppable dependency
- This does not fix the bug itself but it prevents crashing
- Until the lib gets an update, we should probably remove the menu entry.

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/di/modules/RepositoryModule.kt
+++ b/app/src/main/java/io/gnosis/safe/di/modules/RepositoryModule.kt
@@ -3,6 +3,9 @@ package io.gnosis.safe.di.modules
 import com.unstoppabledomains.config.network.model.Network
 import com.unstoppabledomains.resolution.DomainResolution
 import com.unstoppabledomains.resolution.Resolution
+import com.unstoppabledomains.resolution.TickerVersion
+import com.unstoppabledomains.resolution.dns.DnsRecord
+import com.unstoppabledomains.resolution.dns.DnsRecordsType
 import com.unstoppabledomains.resolution.naming.service.NamingServiceType
 import dagger.Module
 import dagger.Provides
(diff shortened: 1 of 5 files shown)


app/src/main/java/io/gnosis/safe/di/modules/RepositoryModule.kt
```diff
@@ -51,20 +53,86 @@ class RepositoryModule {
...
-        val network = Network.valueOf(firstWord.toUpperCase());
-        return Resolution.builder()
+        val firstWord = BuildConfig.BLOCKCHAIN_NET_URL.removePrefix("https://").split(".").first()
+        val network = Network.valueOf(firstWord.toUpperCase())
+        return try {
+            Resolution.builder()
                 .chainId(NamingServiceType.CNS, network)
                 .infura(NamingServiceType.CNS, BuildConfig.INFURA_API_KEY)
                 .build()
+        } catch (throwable: Throwable) {
```
biafra23, line 63: Is there a better way than providing an "empty" interface? If not, then I'd move this to separate file called DummyDomainResolution.

app/src/main/java/io/gnosis/safe/di/modules/RepositoryModule.kt
```diff
@@ -51,20 +53,86 @@ class RepositoryModule {
...
+            Timber.e(throwable, "Error initializing UnstoppableDomains")
+            object : DomainResolution {
```
biafra23, line 65: Is there a better way than providing an "empty" interface? If not, then I'd move this to separate file called DummyDomainResolution.

app/src/main/java/io/gnosis/safe/di/modules/RepositoryModule.kt
```diff
@@ -51,20 +50,25 @@ class RepositoryModule {
...
+        val firstWord = BuildConfig.BLOCKCHAIN_NET_URL.removePrefix("https://").split(".").first()
+        val network = Network.valueOf(firstWord.toUpperCase())
+        return try {
+            Resolution.builder()
                 .chainId(NamingServiceType.CNS, network)
                 .infura(NamingServiceType.CNS, BuildConfig.INFURA_API_KEY)
                 .build()
+        } catch (throwable: Throwable) {
+            Timber.e(throwable, "Error initializing UnstoppableDomains")
+            DummyDomainResolution()
```
elgatovital, line 62: wouldn't it lead to a crash when using this dummy?

elgatovital:
tests are failing

app/src/main/java/io/gnosis/safe/di/modules/RepositoryModule.kt
biafra23: No. It leads to a generic error :-)

biafra23: image

biafra23:
I don't get what the failing tests have to do with this PR. This is the message I found in the logs:

com.android.ddmlib.InstallException: Unknown failure: Error: Could not access the Package Manager.  Is the system running?

elgatovital (approved):

PR #1411, opened 2021-04-29
Disable UD

biafra23:
Handles #

-
-
-

@gnosis/mobile-devs


elgatovital (approved):
An alternative would be to add group element to the layout and set visibility of this group to gone (easier to revert it later)

Issue #1407, opened 2021-04-28
[Android]No keyboard for password entering if the app is called from background


liliya-soroka:
**Describe the bug**
No keyboard for passcode entering during change passcode flow if the app is called from the background 

**To Reproduce**
Precondition:
set up passcode
1. Go to the app settings ->passcode
2. Click Change passcode
3. send the app to background on Enter current passcode or on the set up a new one
4. call the app from the background
**Current result** No keyboard and the user has to reopen page with the "Enter passcode" field to be able to continue

**Expected behavior**
The keyboard should be opened by default on "enter passcode ( current/new/confirm new) when the app was called from background

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/58a53b3b-8a57-44d2-aeb8-44a576ee1cb3)

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.15.0 - 2503rc

**Environment**
prod mainnet
**Additional context**

biafra23:
In onResume() showKeyboardForView() is called. But the Keyboard is not shown :-(

We need to investigate, why. And use a different method to show the keyboard if necessary.

biafra23:
I tried the following to fix this:
- Use different flags for showSoftInput() (I tried SHOW_FORCED, SHOW_IMPLICIT and 0)
- I tried View.hideSoftKeyboard() and then showSoftKeyboard()

showKeyboardForView does work, when called from onResume() the first tome the screen is opened.

biafra23:
Wrapping showKeyBoardForView() in postDelayed() works:
```
    postDelayed({
        showKeyboardForView()
    }, 200)
```

liliya-soroka:
Verified
app version 2.15.0 - 2539

PR #1416, opened 2021-05-03
Delay showKeyboard by 200ms

biafra23:
Handles #1407

- delay show keyboard 
-
-

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/EnterPasscodeFragment.kt
```diff
@@ -113,3 +113,9 @@ class EnterPasscodeFragment : BaseViewBindingFragment<FragmentPasscodeBinding>()
         }
     }
 }
+
+fun View.delayedShowKeyboardForView() {
```
elgatovital, line 117: you could pass delay value as an argument with default value to this function

biafra23: I also moved it into a separate file: PasscodeUtils

elgatovital (approved):

Issue #1355, opened 2021-03-30
[Android] Passcode setup flow


DmitryBespalov:

elgatovital:
biometrics are handled in #1429

liliya-soroka:
Verified without biometric( it was moved to another ticket - https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-android/1354 )
app version 2.16.0 -2612rc

PR #1420, opened 2021-05-06
Ask for passcode setup on first app start

biafra23:
Handles #1355

- Ask user once to setup a passcode after fresh install

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
```diff
@@ -58,6 +58,12 @@ class StartActivity : BaseActivity(), SafeOverviewNavigationHandler {
         setupNav()
 
         handleNotifications(intent)
+
+        if (settingsHandler.appStartCount <= 1) {
+            Navigation.findNavController(this@StartActivity, R.id.nav_host).navigate(R.id.createPasscodeFragment, Bundle().apply {
```
elgatovital, line 63: on first app start user would have to accept terms and conditions. Could we extend flow on a Terms screen, so that after clicking accept user will be asked to setup a passcode?

elgatovital: on the other thought, we have deal with already installed apps and in case of upgrades this would works better. Could you move it to onResume and extend the when statement there?

elgatovital: Navigation.findNavController(this@StartActivity, R.id.nav_host).navigate... should be moved to a private method (something like setupPasscode())

biafra23: Moving this into onResume has no advantage. We only want to do this once after the first app start.  And the disadvantage that the check not to do it will be more complicated during the first use of the app.

biafra23: I extracted the navigation into a function.

biafra23: The order is:
1st app start:
- accept terms
- update software
- setup passcode

app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
```diff
@@ -58,6 +58,16 @@ class StartActivity : BaseActivity(), SafeOverviewNavigationHandler {
         setupNav()
 
         handleNotifications(intent)
+
+        if (settingsHandler.appStartCount <= 1) {
+            setupPasscode()
```
elgatovital, line 63: what if user already has passcode in place? 
Also `appStartCount` is used as a counter to decide if rate the app dialog should be shown. So it will be set to 0 periodically

biafra23: Can I reuse `showPasscodeBanner()` instead of the appStartCount?

biafra23: Is there a safe way to check whether a passcode is set already? WIll there be an error from the encryptionManager, when I try to `setupPassword()` when it is locked? 

To me it looks like `setupPassword()` will fail, if there is already a password and it was not given to `setupPassword()`

app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
biafra23: I think it makes sense to check this order, when the last PR is being merged.

app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
```diff
@@ -58,6 +58,17 @@ class StartActivity : BaseActivity(), SafeOverviewNavigationHandler {
         setupNav()
 
         handleNotifications(intent)
+
+        if (settingsHandler.showPasscodeBanner) {
+            setupPasscode()
+            settingsHandler.showPasscodeBanner = false
```
elgatovital, line 64: this user won't see showPasscodeBanner anymore and this functionality can be removed?

biafra23: I use a separate boolean.

elgatovital (approved):

Issue #1429, opened 2021-05-11
Integrate biometric prompt


elgatovital:
- [x] Integrate biometric api
- [x] Show biometric prompt if biometric authentication is available on device
- [x] Display biometric icon on enter passcode screen if biometric option is available
- [x] Display biometric setting item in passcode settings if biometric option is available
- [x] React to system changes regarding biometric data

    Touch ID/Face ID button will request the system biometry login instead of entering passcode.
    Touch ID / Face ID button must be hidden if the "Login with biometrics" switch is OFF.
    The button is only for requesting TouchID/FaceID, it doesn't change or navigate to the settings.

Ask to enable biometry after the passcode is set

    Fingerprint icon
    Cancel button
    Text “Would you like to enable biometry for passcode?”

Remember user’s choice - that will set the position of the “Login with biometry” setting.

...
[spec link](
https://docs.google.com/document/d/1NegSN4t1dsysUewY_K7v3FLi_sORpS-QpfiTAUZkrn8/edit#bookmark=id.y00zl3xybmf1)

Note: we don't have test device for Face ID testing

liliya-soroka:
For testing : https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-android/1353 - retest each point from the ticket

biafra23:
Here is a screenshot of the biometric prompt. Every string that ends in (customizable) can be changed. 
The fields are called:
- Title
- Subtitle
- Description
- NegativeButton

![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/35d38a80-88c1-4749-a5f7-32168542de4b)

biafra23:
Example for app access:

![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/9170be71-7086-4320-ae15-b70dc37afec6)

biafra23:
Example for signature authorization:
![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/40e14512-5eda-4946-869b-cc81c7d7fcc7)

liliya-soroka:
Returned back:
Description: Can not activate biometric optiuon after updating the app to the rc version

1. Install the previous app version
2. add safe-> import owners-> set up passcode
3. update to the lates version ( rc version)
4. Go to the passcode settings
5. turn on "Biometric" option 
[...]

PR #1435, opened 2021-05-12
Use Biometrics for app unlock and confirmations.

biafra23:
Handles #1429

- Allow verification with fingerprint or face identification for opening the app and signing
- Enroll fingerprints when phone is not yet setup 
- Use Cipher returned by successful biometric authentication to decrypt passcode when signing txs
- Invalidate biometric keys it fingerprints were changed those 
- show login screen when opening from notification

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/ChangeRepeatPasscodeFragment.kt
```diff
@@ -81,6 +82,16 @@ class ChangeRepeatPasscodeFragment : BaseViewBindingFragment<FragmentPasscodeBin
             input.doOnTextChanged(onSixDigitsHandler(digits, requireContext()) { digitsAsString ->
                 if (digitsAsString == passcodeArg) {
                     viewModel.disableAndSetNewPasscode(newPasscode = passcodeArg, oldPasscode = oldPasscode)
+                    val cm = CryptographyManager()
```
elgatovital, line 85: cryptography manager operations should not be in fragment. I suggest to add this code to HeimdallEncryptionManager and use it in a view model of the fragment

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/ConfigurePasscodeFragment.kt
```diff
@@ -91,6 +94,10 @@ class ConfigurePasscodeFragment : BaseViewBindingFragment<FragmentPasscodeBindin
             }
 
             input.doOnTextChanged(onSixDigitsHandler(digits, requireContext()) { digitsAsString ->
+                val cipher = cm.getInitializedCipherForEncryption(KEY_NAME)
```
elgatovital, line 97: same  here as in the comment above

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/CryptographyManager.kt
```diff
@@ -0,0 +1,185 @@
+package io.gnosis.safe.ui.settings.app.passcode
```
elgatovital, line 1: it creates confusion having 2 separate classes for handling authentication. Now its HeimdallEncryptionManager and CryptographyManager. I think they should be combined

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/EnterPasscodeFragment.kt
```diff
@@ -50,6 +62,79 @@ class EnterPasscodeFragment : BaseViewBindingFragment<FragmentPasscodeBinding>()
         super.onResume()
         binding.input.setRawInputType(InputType.TYPE_CLASS_NUMBER)
         binding.input.delayShowKeyboardForView()
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
elgatovital, line 65: this could be extracted to a separate method to keep onResume more legible

app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
```diff
@@ -48,6 +49,8 @@ class StartActivity : BaseActivity(), SafeOverviewNavigationHandler, AppStateLis
     private val toolbar by lazy { findViewById<View>(R.id.toolbar) }
     private val navBar by lazy { findViewById<BottomNavigationView>(R.id.nav_bar) }
 
+    private val handler = Handler()
```
elgatovital, line 52: this is deprecated. Change to
```kotlin
val handler = Handler(Looper.getMainLooper())
```

app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
```diff
@@ -303,19 +307,21 @@ class StartActivity : BaseActivity(), SafeOverviewNavigationHandler, AppStateLis
...
 
     private fun askForPasscode() {
-        Navigation.findNavController(this@StartActivity, R.id.nav_host).navigate(R.id.enterPasscodeFragment, Bundle().apply {
-            putBoolean("requirePasscodeToOpen", true)
-        })
+        handler.postDelayed({
+            Navigation.findNavController(this@StartActivity, R.id.nav_host).navigate(R.id.enterPasscodeFragment, Bundle().apply {
+                putBoolean("requirePasscodeToOpen", true)
+            })
+        }, 600)
```
elgatovital, line 325: this delay here is only needed if askForPasscode is called in onCreate before handleIntent. So 
```kotlin
if (settingsHandler.requirePasscodeToOpen && settingsHandler.usePasscode) {
            askForPasscode()
}
could be moved to handleIntent after all other navigation activities. I can imagine 600ms is unreliable value and we can't garantee all transitions will happen withing this interval. When comming from background this delay is not needed at all.
```

biafra23: The delay in mainly needed when opening from a push notification. But also when coming back from background. When I put this into handleIntent.  Then askForPasscode() is called twice when opening from a push notification.

app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
```diff
@@ -151,6 +159,10 @@ class StartActivity : BaseActivity(), SafeOverviewNavigationHandler, AppStateLis
         val navController = Navigation.findNavController(this, R.id.nav_host)
         navBar.setupWithNavController(navController)
         navController.addOnDestinationChangedListener { _, destination, _ ->
+
+            if (settingsHandler.appStartCount >= 3 && (destination.id == R.id.assetsFragment || destination.id == R.id.settingsFragment || destination.id == R.id.transactionsFragment)) {
```
elgatovital, line 163: we should add here !handlingPushNotification to the condition

elgatovital (changes requested):
see last comment

elgatovital (approved):

PR #1436, opened 2021-05-14
Ignore sem ver extensions when comparing remote config values

elgatovital:
@gnosis/mobile-devs

### DIFF
--- a/app/src/test/java/io/gnosis/safe/utils/SemVerTest.kt
+++ b/app/src/test/java/io/gnosis/safe/utils/SemVerTest.kt
@@ -144,23 +144,23 @@ class SemVerTest {
         assertFalse(version2.isInside(rangeList1, true))
 
         val rangeList2 = "2.15.0...2.17.0"
-        assert(version1.isInside(rangeList2, true))
-        assert(version2.isInside(rangeList2, true))
+        assertTrue(version1.isInside(rangeList2, true))
+        assertTrue(version2.isInside(rangeList2, true))
 
         val rangeList3 = "2.11.0,2.12...2.13.0,2.14...2.16.0"
-        assert(version1.isInside(rangeList3, true))
+        assertTrue(version1.isInside(rangeList3, true))
         assertFalse(version2.isInside(rangeList3, true))
 
         val rangeList4 = ""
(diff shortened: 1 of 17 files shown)


app/src/test/java/io/gnosis/safe/utils/SemVerTest.kt
```diff
@@ -52,6 +65,43 @@ class SemVerTest {
...
+        val rangeString1 = "2.15.0...2.17.0"
+        val range1 = SemVer.parseRange(rangeString1, true)
+        assert(range1.first != null)
+        assert(range1.second != null)
+        assertEquals(version1, range1.first)
+        assertEquals(version2, range1.second)
+
+        val rangeString2 = "2.15.0"
+        val range2 = SemVer.parseRange(rangeString2, true)
+        assert(range2.first != null)
```
rmeissner, line 84: https://stackoverflow.com/questions/2966347/assert-vs-junit-assertions 😛

elgatovital: thank you for pointing this out

app/src/test/java/io/gnosis/safe/utils/SemVerTest.kt
```diff
@@ -18,6 +17,20 @@ class SemVerTest {
         assertEquals(SemVer(2, 15, 0, "255rc-internal"), version3)
     }
 
+    @Test
+    fun `parse (versionString, ignoreExtensions) should return SemVer`() {
+        val version1 = SemVer.parse("2.15.0", true)
+        assertEquals(SemVer(2, 15, 0), version1)
+
+        val version2 = SemVer.parse("2.15.0-255-internal", true)
+        assertNotEquals(SemVer(2, 15, 0, "255-internal"), version2)
```
biafra23, line 26: Why does this have to be not equal?

biafra23 (approved):

app/src/test/java/io/gnosis/safe/utils/SemVerTest.kt
elgatovital: yes. Here we test that extension part is ignored during parsing

Issue #1440, opened 2021-05-17
[Android]Set up in 2.15.0 passcode is reseted in the 2.16.0


liliya-soroka:
**Describe the bug**
Set up in 2.15.0 passcode is reseted in the 2.16.0. The user is asked to set up a passcode on the app start after updating to 2.16.0 even when the passcode was set up in 2.15.0.
After the app restart the passcode is marked as set up but all options "on app start", "transaction confirmation" are off by default 
**To Reproduce**
1. Set up 2.15.0 app version
2. Set up passcode with safe and owner key
3. Update the app to the 2.16.0
4. Start the app
Current result: the user is asked to set up passcode

5. restart the app and go to the passcode settings
Current result: The passcode option is on, by "on app start " and "on confirmation" are off by default.

**Expected behavior**
The passcode should not be reseted by the app update.
request the passcode on the transaction confirmation should be on by default if the app was updated with the enabled passcode

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.16.0 - 2621rc

**Environment**
prod mainnet 
**Additional context**

liliya-soroka:
Verified
app version 2.16.0 - 2631rc

PR #1441, opened 2021-05-17
Fix passcode reset on update

elgatovital:
Handles #1440

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
@@ -131,7 +131,15 @@ class StartActivity : BaseActivity(), SafeOverviewNavigationHandler, AppStateLis
                 // do not start rate flow and update screen together
                 when {
                     settingsHandler.showUpdateInfo -> askToUpdate()
-                    settingsHandler.askForPasscodeSetupOnFirstLaunch -> setupPasscode()
+                    settingsHandler.askForPasscodeSetupOnFirstLaunch -> {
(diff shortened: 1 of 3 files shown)


app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
```diff
@@ -131,7 +131,15 @@ class StartActivity : BaseActivity(), SafeOverviewNavigationHandler, AppStateLis
                 // do not start rate flow and update screen together
                 when {
                     settingsHandler.showUpdateInfo -> askToUpdate()
-                    settingsHandler.askForPasscodeSetupOnFirstLaunch -> setupPasscode()
+                    settingsHandler.askForPasscodeSetupOnFirstLaunch -> {
+                        if (!settingsHandler.usePasscode) {
+                            setupPasscode()
+                        } else {
+                            settingsHandler.askForPasscodeSetupOnFirstLaunch = false
+                            settingsHandler.requirePasscodeToOpen = true
```
biafra23, line 139: Is that the default for upgrading users?

biafra23: That is what I meant. So if the do nothing, they need the passcode to open the app after upgrading.

elgatovital: yes, as discussed in https://gnosisinc.slack.com/archives/CQB729LLF/p1621249552233600

biafra23 (approved):

Issue #1452, opened 2021-05-20
Changes to Edit Key Name screen


DmitryBespalov:
**Issue  by:** <a href='https://github.com/DmitryBespalov'>DmitryBespalov</a>
**Original date:** 2021-05-20 11:40:49 UTC
**Originally opened as:** <a href='https://github.com/gnosis/safe-ios/issues/1182'>gnosis/safe-ios/issues/1182</a>
**Original assignees:** 
**Status on 2021-05-20:** open
<hr>

**Title:** Edit Key Name

Delete the “Remove owner key” button from this screen

**Zeplin**

*   Links

**Tracking**

*   Existing screen id: screen_owner_edit_name

liliya-soroka:
Can not be tested in 2.16.0 - 2683 , as nothing happens on the owner key click on "owner keys" list
When will it be available for testing ?
@elgatovital

liliya-soroka:
Verified
app version 2.16.0 -2695

Issue #1449, opened 2021-05-20
Changes to Owner Keys screen


DmitryBespalov:
**Issue  by:** <a href='https://github.com/DmitryBespalov'>DmitryBespalov</a>
**Original date:** 2021-05-20 11:40:17 UTC
**Originally opened as:** <a href='https://github.com/gnosis/safe-ios/issues/1179'>gnosis/safe-ios/issues/1179</a>
**Original assignees:** 
**Status on 2021-05-20:** open
<hr>

**Elements**

*   **Plus (+)** button
    *   Opens the **How To Add Key **screen
*   Tap on the key row
    *   Opens **Owner Key Details** screen
    *   Disable the current functionality of copy key address on tapping on the key

**Zeplin**

https://zpl.io/aR4BrBe

*   Links

**Tracking**

*   Existing screen id: screen_owner_list

liliya-soroka:
Blocked in 2.16.0 -2683
Nothing is happening on the owner key click

liliya-soroka:
Verified
app version 2.16.0  -2695

Issue #1447, opened 2021-05-20
Create Owner Key Intro screen


DmitryBespalov:
**Issue  by:** <a href='https://github.com/DmitryBespalov'>DmitryBespalov</a>
**Original date:** 2021-05-20 11:39:53 UTC
**Originally opened as:** <a href='https://github.com/gnosis/safe-ios/issues/1177'>gnosis/safe-ios/issues/1177</a>
**Original assignees:** 
**Status on 2021-05-20:** open
<hr>

**Elements:**

*   **Title: **Create Owner Key
*   **How does it work?**
    *   To use this app as an owner in the Gnosis Safe, you can create a key and add it as an owner of your Safe. When you tap Next, the app will create a new private key from a new unique seed phrase.
*   **How secure is that?**
    *   _(iOS)_ The owner key and the seed phrase are stored in the device's secure store - iOS Keychain.
    *   _(Android)_ The owner key and the seed phrase are securely stored.
    *   How is the private key stored on mobile?
        *   Link to [https://help.gnosis-safe.io/en/articles/4866738-how-are-private-keys-stored-on-gnosis-safe-mobile](https://help.gnosis-safe.io/en/articles/4866738-how-are-private-keys-stored-on-gnosis-safe-mobile) 
*   **How to export?**
    *   To export your owner key or seed phrase, navigate to the key details.
*   **Next **button
    *   When Next is pressed, the app creates a unique seed phrase and derives the key at index 0, and uses that to show the key address in the next screen.
    *   When the user goes back to this screen and presses “Next” again then the user must be able to see the same key again.

**Zeplin**

*   Links

**Tracking**

*   Screen id: screen_owner_generate_info

liliya-soroka:
Reopened:
Can not test "
Next button
When Next is pressed, the app creates a unique seed phrase and derives the key at index 0, and uses that to show the key address in the next screen.
When the user goes back to this screen and presses “Next” again then the user must be able to see the same key again. "
as nothing happens on the Next link click on "Create new key intro" page

app version 2.16.0 -2683
Huawei P30 /10

liliya-soroka:
Verified
app version 2.16.0 -2695

Issue #1446, opened 2021-05-20
Create How to Add Key screen


DmitryBespalov:
**Issue  by:** <a href='https://github.com/DmitryBespalov'>DmitryBespalov</a>
**Original date:** 2021-05-20 11:39:32 UTC
**Originally opened as:** <a href='https://github.com/gnosis/safe-ios/issues/1176'>gnosis/safe-ios/issues/1176</a>
**Original assignees:** 
**Status on 2021-05-20:** open
<hr>

**Texts**

*   **Title**
    *   Add Owner Key 
*   Create New Owner Key
    *   Create a new key that you can use as owner of your Gnosis Safe
*   Import Existing Owner Key
    *   Import an existing owner key or a seed phrase

**Zeplin**

https://zpl.io/ag3M5KA 

**Tracking**

*   Just screen tracking: screen_owner_options

liliya-soroka:
Verified
app version 2.16.0 -2683

PR #1457, opened 2021-05-26
Start screens for generate key flow; adjust owner key import ui

elgatovital:
Handles #1446 
Handles #1447
Handles #1449
Handles #1452

- Adjust setting item
- Add "How to Add Owner" screen
- Delete the “Remove owner key” button from Edit Key Name Screen
- Add intro screen for "Create Owner Key" flow
- Adjust navigation

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/di/components/ViewComponent.kt
```diff
@@ -73,8 +75,12 @@ interface ViewComponent {
 
     fun inject(fragment: SettingsFragment)
 
+    fun inject(fragment: OwnerHowToAddFragment)
```
biafra23, line 78: Why `OwnerHowToAddFragment` and not `OwnerAddFragment` ?

app/src/main/java/io/gnosis/safe/ui/settings/owner/list/OwnerListAdapter.kt
```diff
@@ -128,8 +122,7 @@ class LocalOwnerForSigningViewHolder(private val viewBinding: ItemOwnerLocalBind
                     R.string.settings_app_imported_owner_key_default_name,
                     owner.address.shortChecksumString()
                 ) else owner.name
-            remove.visible(false)
-            edit.visible(false)
+            arrow.visible(false)
```
biafra23, line 125: Why should this chevron/arrow be invisible in any case?

app/src/main/java/io/gnosis/safe/di/components/ViewComponent.kt
elgatovital: because it's not adding an owner. This fragment informs about different possibilities how to add owner and proposes different ways to proceed.

app/src/main/res/layout/view_settings_item.xml
```diff
@@ -4,76 +4,97 @@
...
         app:layout_constraintStart_toEndOf="@+id/image"
         app:layout_constraintTop_toTopOf="parent"
+        app:layout_constraintVertical_chainStyle="packed"
+        app:layout_goneMarginBottom="16dp"
         app:layout_goneMarginLeft="16dp"
         tools:text="Fiat currency" />
 
+    <TextView
+        android:id="@+id/description"
+        style="@style/TextLight"
```
biafra23, line 59: Somehow this looks smaller on my phone than the Zeplin design. But the sizes seem to be the right value.

app/src/main/java/io/gnosis/safe/di/components/ViewComponent.kt
biafra23: What about `OwnerAddOptionsFragment` ?

elgatovital:
> In `fragment_owner_list.xml` the IMPORT button should be replaced by a "+" button.

It would be inconsistent with all other screens where we use text labels as actionable toolbar item. Also import - export is more consistent than + - export

Issue #1462, opened 2021-05-26
[Android] Enter touch id for the app screen blocks the device 


liliya-soroka:
**Describe the bug**
Enter touch id for the app screen blocks the device. The user can not answer on called, cencel the screen, send the app to background and etc. The device is locked by the our biometric screen till the either valid fingerprint is used or the wrong is entered a few times

**To Reproduce**
1. Set up biometric option on
2. send the app to background
3. call the app from background
4.try to swipe the app to go to the another app without unlocking it 
5. call on messenger installed on the device
Current result: the user can not answer because the device is locked by the Enter biometric data gnosis multisig app screen 

**Expected behavior**
It should be possible to send app to background without unlocking it when the biometric option is used.It should not lock the user's device

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.17.0 -2674rc

**Environment**
rinkeby, staging 
**Additional context**

biafra23:
This behavior seems to be device dependent. I have no problem on any of my devices to put the app into background when the fingerprint prompt is shown. What device are you using?

What about tapping on "Use passcode"? That should be the option to not use the fingerprint.

liliya-soroka:
Verified in another apps, the same screen that locks the device is used

PR #1463, opened 2021-05-26
Don't start login prompt twice.

biafra23:
Handles #1462, #1465

- Show passcode prompt only once when going to background multiple times
- When updating from previous app version do not enable request passcode on app start

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/EnterPasscodeFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/EnterPasscodeFragment.kt
@@ -31,7 +31,6 @@ import pm.gnosis.svalinn.common.utils.snackbar
 import pm.gnosis.svalinn.common.utils.visible
 import javax.inject.Inject
 
-
 class EnterPasscodeFragment : BaseViewBindingFragment<FragmentPasscodeBinding>() {
 
     override fun screenId() = ScreenId.PASSCODE_ENTER
@@ -62,7 +61,6 @@ class EnterPasscodeFragment : BaseViewBindingFragment<FragmentPasscodeBinding>()
             alreadyStarted = false
         } else {
             alreadyStarted = true
-            binding.input.setRawInputType(InputType.TYPE_CLASS_NUMBER)
             binding.input.delayShowKeyboardForView()
             authenticateWithBiometrics()
         }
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/EnterPasscodeFragment.kt
```diff
@@ -54,9 +56,16 @@ class EnterPasscodeFragment : BaseViewBindingFragment<FragmentPasscodeBinding>()
...
-        binding.input.setRawInputType(InputType.TYPE_CLASS_NUMBER)
-        binding.input.delayShowKeyboardForView()
-        authenticateWithBiometrics()
+        if (alreadyStarted) {
+            findNavController().popBackStack(R.id.enterPasscodeFragment, true)
+            binding.input.hideSoftKeyboard()
+            alreadyStarted = false
+        } else {
+            alreadyStarted = true
+            binding.input.setRawInputType(InputType.TYPE_CLASS_NUMBER)
```
elgatovital, line 65: why do we need to set input type in onResume? Can it be done in onViewCreated or even better in xml?

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/EnterPasscodeFragment.kt
```diff
@@ -54,9 +56,16 @@ class EnterPasscodeFragment : BaseViewBindingFragment<FragmentPasscodeBinding>()
 
     override fun onResume() {
         super.onResume()
-        binding.input.setRawInputType(InputType.TYPE_CLASS_NUMBER)
-        binding.input.delayShowKeyboardForView()
-        authenticateWithBiometrics()
+        if (alreadyStarted) {
+            findNavController().popBackStack(R.id.enterPasscodeFragment, true)
```
elgatovital, line 60: could you elaborate what happens here exactly and why are we removing fragment from stack? Is it possible for EnterPasscodeFragment to be added multiple times?

elgatovital: you could use nav options  to prevent such problem if this is the case here
```
 navigate(
    @NonNull NavDirections directions,
    @Nullable NavOptions navOptions
)
```

```
NavOptions.Builder().setLaunchSingleTop(true).build() 
```

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/EnterPasscodeFragment.kt
biafra23: I think we can remove that. It should be set in xml. But Liliya did not see the right keyboard. So I put it here too. But in the end it was her phone.

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/EnterPasscodeFragment.kt
biafra23: It happenned because each time the app is pulled into foreground, appInForeground() is called an it starts a new enterPasscodeFragment.

biafra23: And since we have no feedback when it is closed, I tried to prevent multiple openings here. But the flag launchSingleTop could also work.

Issue #1450, opened 2021-05-20
Owner Key Details screen


DmitryBespalov:
**Issue  by:** <a href='https://github.com/DmitryBespalov'>DmitryBespalov</a>
**Original date:** 2021-05-20 11:40:28 UTC
**Originally opened as:** <a href='https://github.com/gnosis/safe-ios/issues/1180'>gnosis/safe-ios/issues/1180</a>
**Original assignees:** 
**Status on 2021-05-20:** open
<hr>

**Elements**

*   Title
    *   Owner Key
*   Identicon - of the key address
*   Key Name
*   **Edit Key Name** button
    *   Opens **Edit Key Name** screen
*   QR Code of the address
    *   Contents of the QR code is checksummed address string with 0x prefix
*   **Export **button
    *   If passcode option “Require for exporting keys” is set, request passcode first
    *   Opens **Export **screen
*   **Remove Owner Key **button
    *   Starts **Remove Key** flow (existing)

**Zeplin**

https://zpl.io/VQNg4ek

*   Links

**Tracking**

*   Screen id: screen_owner_details
*   When removing: trigger [user_key_deleted](https://github.com/gnosis/safe/wiki/Tracking#user_key_deleted)
    *   Update user prop [num_keys_imported](https://github.com/gnosis/safe/wiki/Tracking#num_keys_imported) or [num_keys_generated](https://github.com/gnosis/safe/wiki/Tracking#num_keys_generated), respectively

liliya-soroka:
Verified ( except Export option - It will be tested in Export task)

app version 2.16.0 - 2695

PR #1471, opened 2021-05-31
Add owner key details screen

elgatovital:
Handles #1450

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/settings/owner/details/OwnerDetailsViewModel.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/settings/owner/details/OwnerDetailsViewModel.kt
@@ -1,0 +1,70 @@
+package io.gnosis.safe.ui.settings.owner.details
+
+import android.graphics.Bitmap
+import android.graphics.Color
+import io.gnosis.data.repositories.CredentialsRepository
+import io.gnosis.safe.Tracker
+import io.gnosis.safe.notifications.NotificationRepository
+import io.gnosis.safe.ui.base.AppDispatchers
+import io.gnosis.safe.ui.base.BaseStateViewModel
+import pm.gnosis.crypto.utils.asEthereumAddressChecksumString
+import pm.gnosis.model.Solidity
+import pm.gnosis.svalinn.common.utils.QrCodeGenerator
+import timber.log.Timber
+import javax.inject.Inject
+
+class OwnerDetailsViewModel
+@Inject constructor(
+    private val credentialsRepository: CredentialsRepository,
+    private val notificationRepository: NotificationRepository,
+    private val tracker: Tracker,
+    private val qrCodeGenerator: QrCodeGenerator,
+    appDispatchers: AppDispatchers
+) : BaseStateViewModel<OwnerDetailsState>(appDispatchers) {
+
+    override fun initialState() = OwnerDetailsState(ViewAction.None)
+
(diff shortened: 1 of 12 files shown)


app/src/main/java/io/gnosis/safe/ui/settings/owner/list/OwnerListFragment.kt
```diff
@@ -90,6 +92,17 @@ class OwnerListFragment : BaseViewBindingFragment<FragmentOwnerListBinding>(), O
     override fun onResume() {
         super.onResume()
         viewModel.loadOwners()
+
+        //FIXME: find better way to pass results in nav graph
```
biafra23, line 96: This StackOverflow question might have an answer for us. I am not sure which one really works and is better than what we have today: https://stackoverflow.com/questions/50334550/navigation-architecture-component-passing-argument-data-to-the-startdestination

app/src/main/java/io/gnosis/safe/ui/settings/owner/details/OwnerDetailsViewModel.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    fun loadOwnerDetails(address: Solidity.Address) {
+        safeLaunch {
+
+            val ownerName = credentialsRepository.owner(address)!!.name
+
+            val qrCode = runCatching {
+                qrCodeGenerator.generateQrCode(
+                    address.asEthereumAddressChecksumString(),
+                    512,
```
biafra23, line 35: Should this be an extracted constant? Does it have a relationship with the `owner_qr_code` views width and height?

app/src/main/res/layout/fragment_owner_details.xml
```diff
@@ -0,0 +1,176 @@
...
+            android:ellipsize="end"
+            android:maxLines="1"
+            android:text="@string/signing_owner_details_key_address"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toBottomOf="@+id/owner_address" />
+
+        <com.google.android.material.imageview.ShapeableImageView
+            android:id="@+id/owner_qr_code"
+            android:layout_width="124dp"
```
biafra23, line 137: This looks rather small on my phone. But it is probably the same size as in the share dialog?

elgatovital: yes

elgatovital: yes

app/src/main/java/io/gnosis/safe/ui/settings/owner/list/OwnerListFragment.kt
elgatovital: I would create separate issue for this refactoring (and maybe investigating other available alternatives) as it touches places not related to this feature.

app/src/main/java/io/gnosis/safe/ui/settings/owner/details/OwnerDetailsViewModel.kt
elgatovital: it's size of a bitmap. I would address this in scope of export owner screen, as there we have yet another qr code :)

biafra23 (approved):

Issue #1470, opened 2021-05-31
Biometric dialog can be closed without keyboard opening


rmeissner:
**Describe the bug**
When closing the biometric input dialog without pressing "use passcode" the keyboard is not opened and it is not possible to input the passcode

**To Reproduce**
1. Setup biometric unlock and set that it is used on app start
2. Open the app and close biometric input 
  * Press back
  * Press outside dialog

**Expected behavior**
The keyboard should be open when the biometric input is not visible anymore. And there should be an obvious way to open the keyboard if not (e.g. pressing the 6 passcode dots).

**Smartphone (please complete the following information):**
 - Device: Pixel 3 XL
 - Android Version: 11
 - App Version: 2.16.0-2683

**Environment**
- Mainnet Production

**Additional context**
- The keyboard can be opened on the lock screen by pressing back which is unexpected, as I would expect that the app closes.

biafra23:
@rmeissner Actually in this case the back button is the button to close the keyboard. I am not sure, we can prevent that. If it is on screen the triangle is directed downwards.  I will try to find a way to not close the keyboard on back.

biafra23:
This SO question seems to have three different answers to intercept the close-soft-keyboard button: https://stackoverflow.com/questions/3940127/intercept-back-button-from-soft-keyboard

liliya-soroka:
Verified
the link is added to open keyboard
app version 2.17.0 -2741rc

Issue #1469, opened 2021-05-28
[Android]The app is unlocked without passcode on second call from background if the biometric option is disabled by the system for some time


liliya-soroka:
**Describe the bug**
[Android]The app is unlocked without a passcode on the second call from the background if the biometric option is disabled by the system for some time
The system blocks the fingering option if the wrong fingerprint was used about 5 times

**To Reproduce**
Steps to reproduce the behaviour:
1. Set up passcode with fingerprint in the app
2. Start the app
3. Use wrong fingerprint many times till this option is blocked by the system
4. Don't enter the passcode and send the app to the background
5. call the app from the background and don't enter the passcode
6. send the app to the background
7. call app from background 
Current result: The app is unlocked without a passcode or fingerprint. The app should be send to background while the fingerprint is locked by the system

**Expected behavior**
It should not be possible to use app without entering correct passcode or fingerprint if passcode is on

https://drive.google.com/file/d/1QqgZ3X4Q7o0pAvZwUKEnrM0NMwHoQQEM/view?usp=sharing

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.17.0-2680rc

**Environment**

**Additional context**

biafra23:
Cannot reproduce this with my Pixel 3a (Android 12 beta) will try more phones.

liliya-soroka:
Verified on Huawei P30
app version 2.16.0 -2734

PR #1472, opened 2021-05-31
Use launch single top instead of tracking opening and closing.

biafra23:
Handles #1469, #1470

- Use singleTop to show passcode fragment only once after sending app to background multiple times
- Try to show keyboard with a bigger delay and show button to reopen keyboard if that fails

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/EnterPasscodeFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/EnterPasscodeFragment.kt
@@ -52,7 +52,7 @@ class EnterPasscodeFragment : BaseViewBindingFragment<FragmentPasscodeBinding>()
 
     override fun onResume() {
         super.onResume()
-        binding.input.delayShowKeyboardForView(600)
+        binding.input.delayShowKeyboardForView()
         authenticateWithBiometrics()
     }
 
@@ -92,7 +92,7 @@ class EnterPasscodeFragment : BaseViewBindingFragment<FragmentPasscodeBinding>()
                 backButton.visible(false)
                 requireActivity().onBackPressedDispatcher.addCallback(viewLifecycleOwner, object : OnBackPressedCallback(true) {
                     override fun handleOnBackPressed() {
-                        input.delayShowKeyboardForView(600)
+                        input.delayShowKeyboardForView()
                     }
                 })
             } else {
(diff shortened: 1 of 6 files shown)


app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/EnterPasscodeFragment.kt
```diff
@@ -194,18 +188,17 @@ class EnterPasscodeFragment : BaseViewBindingFragment<FragmentPasscodeBinding>()
     }
 
     private fun onUsePasscode() {
-        binding.input.delayShowKeyboardForView()
+        binding.input.delayShowKeyboardForView(600)
     }
 
     private fun onBiometricsAuthFailed() {
-        binding.input.delayShowKeyboardForView()
+        binding.input.delayShowKeyboardForView(600)
```
elgatovital, line 195: would you adjust the default delay value instead?

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/EnterPasscodeFragment.kt
```diff
@@ -141,6 +130,12 @@ class EnterPasscodeFragment : BaseViewBindingFragment<FragmentPasscodeBinding>()
                     input.hideSoftKeyboard()
                 }
             }
+            contentView.setOnClickListener {
```
elgatovital, line 133: having content view as click target is not enough?

biafra23: Probably not. Let me check if the "button" still reacts, if only the background has the listener.

app/src/main/res/layout/fragment_passcode.xml
```diff
@@ -215,6 +216,22 @@
                 app:layout_constraintEnd_toEndOf="parent"
                 app:layout_constraintStart_toStartOf="parent" />
 
+
+            <TextView
+                android:id="@+id/show_keyboard"
+                style="@style/TextLink"
+                android:layout_width="256dp"
```
elgatovital, line 223: why does it have fixed margins and dimensions? Layout should be responsive.
Also please make sure this label won't be pushed up when keyboard appears -> keyboard should be displayed on top of this label.

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/PasscodeSettingsFragment.kt
```diff
@@ -78,7 +78,7 @@ class PasscodeSettingsFragment : SafeOverviewBaseFragment<FragmentSettingsAppPas
             useBiometrics.settingSwitch.setOnClickListener {
                 if (useBiometrics.settingSwitch.isChecked) {
                     if (canAuthenticate() == BIOMETRIC_ERROR_NONE_ENROLLED) {
-                        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
+                        if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.M) {
                             startActivity(Intent(Settings.ACTION_SECURITY_SETTINGS));
                         } else {
                             startActivityForResult(Intent(Settings.ACTION_FINGERPRINT_ENROLL), 1 /* REQUESTCODE_FINGERPRINT_ENROLLMENT */)
```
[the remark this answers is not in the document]
biafra23: We are ignoring it anyways.
But I can define a constant for it.

elgatovital (approved):

PR #1478, opened 2021-06-01
Using NullableConverters

biafra23:
- Use NullableConverters for ExcrypteVyteArray and EncryptedString


data/src/main/java/io/gnosis/data/models/Owner.kt
```diff
@@ -26,7 +27,7 @@ data class Owner(
     val privateKey: EncryptedByteArray? = null,
```
elgatovital, line 27: privateKey should also have nullable type converter

elgatovital:
@biafra23 please check tests

elgatovital (approved):

Issue #1451, opened 2021-05-20
Export Key screen


DmitryBespalov:
**Issue  by:** <a href='https://github.com/DmitryBespalov'>DmitryBespalov</a>
**Original date:** 2021-05-20 11:40:40 UTC
**Originally opened as:** <a href='https://github.com/gnosis/safe-ios/issues/1181'>gnosis/safe-ios/issues/1181</a>
**Original assignees:** 
**Status on 2021-05-20:** open
<hr>

*   Title
    *   Export 
*   Segmented buttons
    *   **Seed Phrase**
        *   Shows the seed phrase with numbered words
        *   NO QR-code
    *   **Private Key**
        *   Shows the private key as a hexadecimal string
        *   QR code: content is the private key string
*   **Share **button
    *   Opens system “Share” sheet to share either seed phrase or a private key (whatever is currently selected
*   **Back** button
    *   Closes the flow and goes back to previous screen (key details)

**Zeplin**

https://zpl.io/VDgYQMg
https://zpl.io/2j3BdmA

*   Links

**Tracking**

*   Screen id when seed is active: screen_owner_export_seed
*   Screen id when private key is active: screen_owner_export_key

liliya-soroka:
Verified
app version 2.17.0 (450)

tracking will be checked separately for all tasks related to the "generate key" feature

liliya-soroka:
Reopened:
Nothing happens on Export option click 

app version 2.16.0 -2695
Huawei P30/10

elgatovital:
it wasn't even in qa yet :)

liliya-soroka:
Verified
app version 2.16.0 -2734

PR #1481, opened 2021-06-02
Owner export UI

elgatovital:
Handles #1451

- Add owner export screens
- Adjust navigation

@gnosis/mobile-devs


biafra23:
Te passcode will be required in a separate PR, I guess?

biafra23:
When exporting an imported key (no passphrase) the sub title is malformed:
image

app/src/main/java/io/gnosis/safe/ui/settings/owner/details/OwnerDetailsFragment.kt
```diff
@@ -90,6 +89,7 @@ class OwnerDetailsFragment : BaseViewBindingFragment<FragmentOwnerDetailsBinding
 
                         ownerQrCode.setImageBitmap(viewAction.ownerDetails.qrCode)
                         content.animate().alpha(1f).setDuration(1000)
+                        exportButton.isEnabled = viewAction.ownerDetails.exportable
```
biafra23, line 92: exportable would be false for hardware wallet backed keys?

biafra23:
The seed export screen looks much more polished than the key export screen. But that's how our design is ATM.

elgatovital:
> Te passcode will be required in a separate PR, I guess?

yes

app/src/main/java/io/gnosis/safe/ui/settings/owner/details/OwnerDetailsFragment.kt
elgatovital: yes, there would be no private key or seed phrase stored on device

biafra23 (approved):

Issue #1453, opened 2021-05-20
Changes to Passcode Settings screen


DmitryBespalov:
**Issue  by:** <a href='https://github.com/DmitryBespalov'>DmitryBespalov</a>
**Original date:** 2021-05-20 11:40:59 UTC
**Originally opened as:** <a href='https://github.com/gnosis/safe-ios/issues/1183'>gnosis/safe-ios/issues/1183</a>
**Original assignees:** 
**Status on 2021-05-20:** open
<hr>

*   Add new toggle “Require for exporting keys”
    *   When enabled, the passcode is asked to open **Export** screen from **Key Details**
    *   When disabled, the passcode is NOT asked to open **Export** screen from **Key Details**

liliya-soroka:
Verified
app version 2.17.0 - 2741rc

PR #1483, opened 2021-06-03
Protect owner key export with passcode

elgatovital:
Handles #1453

- New setting item in passcode settings
- New flags in settings handler
- No export without passcode if option is enabled
- Adjust navigation

@gnosis/mobile-devs


biafra23:
- Asking for passcode for exporting key should be ON when enabling passcode
- Asking for passcode for exporting key should be ON when upgrading the app to a version that allows export (if passcode was on)

biafra23 (approved):

Issue #1431, opened 2021-05-11
[Android]Crash: ECDomainParameters.java line 25


liliya-soroka:
**Describe the bug**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/275431a5-747e-4441-a5d2-fcd0f18c0930)
Crash - https://console.firebase.google.com/u/0/project/safe-firebase-mainnet/crashlytics/app/android:io.gnosis.safe/issues/5513ff0beec2d8974ab58d2b033abe9e?time=last-seven-days&sessionEventKey=609A40F3012500012427708092846F9D_1539393190441211088
and the same :
https://console.firebase.google.com/u/0/project/safe-firebase-mainnet/crashlytics/app/android:io.gnosis.safe/issues/c3cc206d1e1aacb42b9f3116aebbe769?time=last-thirty-days&sessionEventKey=6097D26B005F000164BA9141F8345A77_1538709035143595528
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/3424db34-3a0f-4ae6-978a-c7d5e69af7fc)
**To Reproduce**
1. ????
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/59a30dd7-8998-49ca-b0ce-1bcf9b444943)
**Expected behavior**
No crash 

**Screenshots**

**Smartphone (please complete the following information):**
 - Device: different
 - Android Version: 11
 - App Version 2.15.0 (2531)

**Environment**
prod mainnet
**Additional context**

biafra23:
Timeboxed as it didn't happen in the last version. Only in 2.15.0

biafra23:
I am running this endless loop for a while:
```
            while (true) {
                val mnemonic = bip39.generateMnemonic(languageId = R.id.english)
                derivator.initialize(mnemonic)
                val key = derivator.keyForIndex(0)
                Timber.i("---> mnemonic: $mnemonic")
                val address = derivator.addressesForPage(0, 1)[0]
                credentialsRepository.saveOwner(address, key, "no name")
                val owner  = credentialsRepository.owner(address)!!
                val decryptedKey = credentialsRepository.decryptKey(owner.privateKey!!)
                try {
                    CheckedKeyPair.fromPrivate(key)
                    CheckedKeyPair.fromPrivate(decryptedKey)
                } catch (e: java.lang.IllegalArgumentException) {
                    e.printStackTrace()
                    throw e
                }
[...]

PR #1498, opened 2021-06-09
More key validation

biafra23:
Handles #1431 

- I could not reproduce the problem even after generating more than 10000 keys
- We now check that a key entered by the suer can be used to sign

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/settings/owner/OwnerSeedPhraseViewModel.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/settings/owner/OwnerSeedPhraseViewModel.kt
@@ -10,6 +10,8 @@ import pm.gnosis.mnemonic.Bip39
 import pm.gnosis.model.Solidity
 import pm.gnosis.utils.asBigInteger
 import pm.gnosis.utils.hexAsBigInteger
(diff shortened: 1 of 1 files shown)


biafra23:
Maybe it would be ok to merge this into master and not release?

app/src/main/java/io/gnosis/safe/ui/settings/owner/OwnerSeedPhraseViewModel.kt
```diff
@@ -47,8 +49,11 @@ class OwnerSeedPhraseViewModel
             safeLaunch {
                 updateState { ImportOwnerKeyState.Error(InvalidPrivateKey) }
             }
+        } else if (!keyCanBeUsedForSigning(input.hexAsBigInteger())) {
+            safeLaunch {
+                updateState { ImportOwnerKeyState.Error(InvalidPrivateKey) }
+            }
         } else {
-
             safeLaunch {
```
elgatovital, line 57: it would be more legible if this method has just one top safeLaunch block instead of launching them for every if branch

biafra23:
Is it possible to change the base branch on an open PR?

elgatovital (approved):

Issue #1503, opened 2021-06-10
[Android] App crashed on tx confirmation and Export screen opening when biometric is used


liliya-soroka:
**Describe the bug**
App crashed on tx confirmation and Export screen opening when biometric is used. It happens when the passcode was activated the first time in the app together with biometric at the same time 
The bug will touch:
1. any new user who will install app and turn on passcode with biometric at the same time
2. any users with the old version who did not have passcode set up in old version and will activate it in 2.17.0 the first time together with biometric enabling 

Crash logs : https://console.firebase.google.com/u/0/project/safe-firebase-rinkeby/crashlytics/app/android:io.gnosis.safe.rinkeby/issues/5ed470def35f1bb440a26ea3a07cfe9a?time=last-seven-days&sessionEventKey=60C2470F02BB00016C6C4ECD9E4155DC_1550658975035548011

**To Reproduce**
1. Clean app install 
2. Set up passcode together with enabling biometric ( passcode +biometric activated together the first time in the app)
3. add safe and import owner key
4a. Try to open Export using fingeprint for the imported key
4b. Try to confirm transaction using fingerprint with imported owner

Current result: the app is crashed 

**Expected behavior**
It should be possible to confirm transaction and open Export screen with valid fingeprint

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.17.0 -2758rc

**Environment**
prod, rinkeby
**Additional context**

tschubotz:
marking as critical to fix before release since this is a quite common onboarding flow I suppose

biafra23:
Looks like the passcode is not is not stored when biometrics is enabled automatically when enabling passcode.

liliya-soroka:
Verified
app version 2.17.0 - 2761rc

liliya-soroka:
Reopened:
the same crash - https://console.firebase.google.com/u/0/project/safe-firebase-rinkeby/crashlytics/app/android:io.gnosis.safe.rinkeby/issues/5ed470def35f1bb440a26ea3a07cfe9a?time=last-seven-days&sessionEventKey=60C71F80016A00016AEB6E24BFF75900_1552022872914912328 but with a little bit different steps

1. install 2.16.0 
2. set up passcode
3. add safe without key importing
4. update app to the 2.17.0 -2761rc
5. go to the passcode settings and enable biometric
6. import owner key
7. try to open Export
8. use correct fingerprint
[...]

PR #1504, opened 2021-06-11
Update passcode decrypted by biometric key

biafra23:
Handles #1503

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/PasscodeViewModel.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/PasscodeViewModel.kt
@@ -61,13 +61,6 @@ class PasscodeViewModel
                         updateState { PasscodeState(PasscodeCommandExecuted) }
                     }
                     BIOMETRICS_ENABLE -> {
-                        val cipher = biometricPasscodeManager.getInitializedRSACipherForEncryption(BiometricPasscodeManager.KEY_NAME)
-                        val encrypted = biometricPasscodeManager.encryptData(passcode, cipher)
-                        biometricPasscodeManager.persistEncryptedPasscodeToSharedPrefs(encrypted,
-                            BiometricPasscodeManager.FILE_NAME,
-                            Context.MODE_PRIVATE,
-                            BiometricPasscodeManager.KEY_NAME
-                        )
                         settingsHandler.useBiometrics = true
                         updateState { PasscodeState(PasscodeCommandExecuted) }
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/PasscodeViewModel.kt
```diff
@@ -61,13 +61,6 @@ class PasscodeViewModel
                         updateState { PasscodeState(PasscodeCommandExecuted) }
                     }
                     BIOMETRICS_ENABLE -> {
-                        val cipher = biometricPasscodeManager.getInitializedRSACipherForEncryption(BiometricPasscodeManager.KEY_NAME)
```
biafra23, line 64: not necessary here.

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/RepeatPasscodeFragment.kt
```diff
@@ -115,6 +115,7 @@ class RepeatPasscodeFragment : BaseViewBindingFragment<FragmentPasscodeBinding>(
             input.doOnTextChanged(onSixDigitsHandler(digits, requireContext()) { digitsAsString ->
                 if (passcodeArg == digitsAsString) {
                     viewModel.setupPasscode(digitsAsString)
+                    viewModel.encryptPasscodeWithBiometricKey(digitsAsString)
```
biafra23, line 118: This is the fix. Update after setting the passcode

elgatovital: what happens if device doesn't support biometrics?

biafra23: This does not depend on the biometrics capability. But OTOH it is not necessary on devices that do not have biometrics functionality at all.

elgatovital: so check should be added to see if this action is necessary? Also this seems to be part of password setup. Why not combine it in a view model?

elgatovital: please add todo's, fixme's and let's revisit it later

elgatovital (approved):

Issue #1489, opened 2021-06-03
[Android]  First added safe does not receive notifications when the second safe with the same imported owner is added


liliya-soroka:
Bug description
 First added safe is not received notification if the second safe with the same imported owner is added

From Uxio:
@safe-android @safe-ios Are you sending always all the Safes you want to get notifications from when you send the creation request?

Steps To Reproduce

Add safe = 0x1230B3d59858296A31053C1b8562Ecf89A2f888b
Import owner ( contact Liliya for PK)
create tx and receive notification for the safe in step 1 ( make sure that the push notifications are received
Add safe = 0xf2565317F3Ae8Ae9EA98E9Fe1e7FADC77F823cbD
Send the tx from the first safe to the second safe /or send ERC20 to the first safe
Current result: Only notifications for the second safe are received
The user has not received notifications for the first safe

Expected Result
The user should receive notification for both safe even if the same owner is used in both safes

Device & App version

Device/OS: iPhone 11/14.3
App version: 2.16.0
Environment
prod rinkeby
Additional info
(Optional)

liliya-soroka:
the same issue was on ios ( client side ) - https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-ios/1203

tschubotz:
marking this as major also since this use cases seems quite likely and pushes are one of the nice features currently on the mobile app compared to web.

liliya-soroka:
Verified
app version 2.18.0 -2816rc

PR #1510, opened 2021-06-14
We need to send all safe names with all registrations.

biafra23:
Handles #1489 

- Register all safes for all owners every time.

@gnosis/mobile-devs


biafra23:
We could probably delete register() and user registerOwners() instead if it wasn't for 

https://github.com/gnosis/safe-android/pull/1510/files#diff-4e03d291fa24b5aeaaab3f85132dc282170830173513eefb356af3b8b436f253R155
`notificationManager.updateNotificationChannelGroupForSafe(safe)`
and `safeRepository.saveSafeMeta(SafeMetaData(safe.address, true))`

app/src/main/java/io/gnosis/safe/notifications/NotificationRepository.kt
```diff
@@ -131,10 +131,13 @@ class NotificationRepository(
         kotlin.runCatching {
```
elgatovital, line 131: registerSafe(safe: Safe) should be renamed to registerSafe

biafra23: Right now the parameter is needed to to the onSuccess() stuff. And if we do not need the parameter, we can remove registerSafe and call registerOwners instead. And add the two methods from onSuccess for all safes.

biafra23: Is it safe to call `notificationManager.updateNotificationChannelGroupForSafe(safe)` and `safeRepository.saveSafeMeta(SafeMetaData(safe.address, true))` in `onSuccess()` for all safes everytime?

biafra23: For updateNotificationChannelGroupForSafe I think it is safe. But I am not sure about saveSafeMeta.

biafra23:
I don't think it is a good idea to pass the list of safe. For example some places that call registerSafes() do not have a safeRepository yet.

elgatovital (approved):

Issue #1530, opened 2021-06-22
[Android] Fetch networks info and store it


sche:
The list of supported networks is fetched on the app start.

We do not implement pagination here. We assume that we fetch all networks in one call.

- [ ] Update network information in a database **only** for safes that we have.
- [ ] If we have a safe but safe's network was not returned in the list of supported networks, we rely on cached data in this iteration
- [ ] Use https://safe-client.staging.gnosisdev.com/v1/chains for staging

**QA items**
- not verifiable

liliya-soroka:
https://app.zenhub.com/workspaces/mobile-development-kanban-60cc78740f488d000e996335/issues/gnosis/safe-android/1532  - the netwrok collor will be available for testing when "Fetch networks info and store it" task is ready

liliya-soroka:
The colour and mane are verified
closed 
The task will be tested fully during the functional testing for Multinetwork support

PR #1540, opened 2021-06-23
Fetch network info and parse it

biafra23:
Handles #1530 

- Introduce `ChainInfoRepository`
- Fetch from new endpoint (`v1/chains`) and parse JSON
- Modify all other calls to previous endpoints to have a default chaind of 4 (Rinkeby)

@gnosis/mobile-devs

### DIFF
--- a/data/src/main/java/io/gnosis/data/backend/GatewayApi.kt
+++ b/data/src/main/java/io/gnosis/data/backend/GatewayApi.kt
@@ -1,6 +1,7 @@
 package io.gnosis.data.backend
 
 import io.gnosis.data.BuildConfig
+import io.gnosis.data.models.ChainInfo
 import io.gnosis.data.models.Page
 import io.gnosis.data.models.SafeInfo
 import io.gnosis.data.models.assets.CoinBalances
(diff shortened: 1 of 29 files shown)


data/src/main/java/io/gnosis/data/backend/GatewayApi.kt
```diff
@@ -14,43 +15,71 @@ import java.util.*
 
 interface GatewayApi {
 
-    @GET("/v1/balances/supported-fiat-codes")
-    suspend fun loadSupportedCurrencies(): List<String>
+    @GET("/{chainId}/v1/balances/supported-fiat-codes")
+    suspend fun loadSupportedCurrencies(
+        @Path("chainId") chainId: Int = 4
```
elgatovital, line 20: default value could come from build constants (BuildConfig.CHAIN_ID). I think we even have it already

data/src/main/java/io/gnosis/data/db/daos/ChainDao.kt
```diff
@@ -0,0 +1,27 @@
+package io.gnosis.data.db.daos
+
+import androidx.room.*
+import io.gnosis.data.models.Chain
+
+@Dao
+interface ChainDao {
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    suspend fun insert(chain: Chain)
```
elgatovital, line 10: insert -> save as we will probably use this to also update the chain info

data/src/main/java/io/gnosis/data/models/Chain.kt
```diff
@@ -0,0 +1,24 @@
+package io.gnosis.data.models
+
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+import androidx.room.PrimaryKey
+import io.gnosis.data.models.Chain.Companion.TABLE_NAME
+
+@Entity(tableName = TABLE_NAME)
+data class Chain(
```
elgatovital, line 9: we could combine Chain & ChainInfo into one class as we currently do not have separate dto's and use models from cgw directly

data/src/main/java/io/gnosis/data/models/ChainInfo.kt
```diff
@@ -0,0 +1,30 @@
...
+
+
+@JsonClass(generateAdapter = true)
+data class ChainInfo(
+    @Json(name = "chainId") val chainId: Int,
+    @Json(name = "chainName") val chainName: String,
+    @Json(name = "rpcUrl") val rpcUrl: String,
+    @Json(name = "blockExplorerUrl") val blockExplorerUrl: String,
+    @Json(name = "nativeCurrency") val nativeCurrency: NativeCurrency,
+    @Json(name = "transactionService") val transactionService: String,
```
elgatovital, line 14: we need transactionService url only to possibly show it on network details page?

data/src/main/java/io/gnosis/data/repositories/ChainInfoRepository.kt
```diff
@@ -0,0 +1,28 @@
+package io.gnosis.data.repositories
+
+import io.gnosis.data.backend.GatewayApi
+import io.gnosis.data.db.daos.ChainDao
+import io.gnosis.data.models.Chain
+import io.gnosis.data.models.ChainInfo
+
+class ChainInfoRepository(
```
elgatovital, line 8: how about putting chain related stuff into SafeRepository?

data/src/main/java/io/gnosis/data/models/ChainInfo.kt
biafra23: For the DTO model I wanted to include everything we get. No matter if we need it. I would reduce the data we store to what we need now or until the feature is delivered.

data/src/main/java/io/gnosis/data/repositories/ChainInfoRepository.kt
biafra23: I put it there at first. But there might be parts of the app that want to get network info without using a safe.

data/src/main/java/io/gnosis/data/models/Chain.kt
biafra23: How would we handle embedded types like NativeCurrency and ChainTheme? Flatten them into the chain table or have relations?

elgatovital: flatten them

elgatovital (approved):

Issue #1537, opened 2021-06-22
[Android] Select network when loading safe


sche:
The networks list logic is similar to transactions list logic:
- [x] proper errors handling
- [x] loading indicator
- [x] pull to refresh
- [x] empty list placeholder
- [ ] pagination

Header should be scrollable with the list.
We do not use cached networks data here (or from our database)

- [ ] We update cached data in the database only for the network we select and at the moment when we actually add the safe. There should not be networks without safes.

**QA**
Question: is it possible to check pagination? - probably some custom networks list will be needed (mitmproxy)

liliya-soroka:
to test it fully , I need rinkeby and mainnet in one app at the same time

liliya-soroka:
Verified
the pagination will be checked separately

app version 2.18.0 -2966

PR #1550, opened 2021-06-30
Paginated list of chains

elgatovital:
Handles #1537

- Load pages of chain info
- Paginated list of chain items
- Pass selected chain through add safe flow
- Adjust nav graph

@gnosis/mobile-devs


data/src/main/java/io/gnosis/data/models/Chain.kt
```diff
@@ -19,7 +20,7 @@ data class Chain(
 
     @ColumnInfo(name = COL_BACKGROUND_COLOR)
     val backgroundColor: String
-) {
+) : Serializable {
```
biafra23, line 23: Why is this needed?

biafra23 (approved):

data/src/main/java/io/gnosis/data/models/Chain.kt
elgatovital: to pass it as an argument in a bundle

PR #1561, opened 2021-07-06
Parse chainId in push notifications

biafra23:
Handles #

- The backend already sends the chainId

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/notifications/models/PushNotification.kt
+++ b/app/src/main/java/io/gnosis/safe/notifications/models/PushNotification.kt
@@ -7,79 +7,79 @@ import java.math.BigInteger
 sealed class PushNotification(
     val type: String
 ) {
-
+    abstract val chainId: Int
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/gnosis/safe/notifications/models/PushNotification.kt
```diff
@@ -12,30 +12,34 @@ sealed class PushNotification(
 
     data class ConfirmationRequest(
         override val safe: Solidity.Address,
-        val safeTxHash: String
+        val safeTxHash: String,
+        val chainId: Int
```
elgatovital, line 16: chainId should be in base class

```kotlin
abstract val chainId: Int
abstract val safe: Solidity.Address
```
In subclasses I would put chainId as a first parameter (first we identify chain, then safe on this chain, then transaction for this safe. This sequence of parameters seems sensible to me)

elgatovital (approved):

Issue #1533, opened 2021-06-22
[Android] Add ribbon to transaction details and all nested details screens


sche:
Transaction details screen should have a ribbon
- [x] opened from tx list
- [x] opened from a push notification

*Nested tx details screens*
- [x] multisend actions list 
- [x] action details
- [x] advanced tx details
- [x] multisend decoded transactions

The ribbon should not be scrollable. It should be stick to the header.

**QA**
See above. Please assure that we did not forget any other tx details internal page.

biafra23:
Ribbon is missing on  `advanced tx details`

liliya-soroka:
Verified
app version 2.18.0 -2932

PR #1569, opened 2021-07-07
Add chain ribbon to tx details

elgatovital:
Handles #1533

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewModel.kt
```diff
@@ -73,6 +74,8 @@ class TransactionListViewModel
         type: TransactionPagingSource.Type
     ): Flow<PagingData<TransactionView>> {
 
+        val chain = safe.chain!!
```
biafra23, line 77: This is only ok, if the safe comes from a database. And it does in this case. But is there a better way. For example: Could we make chain not nullable? And then add the default chain data everytime we create a safe on the fly?

biafra23:
Do we really need the chain in TransactionView.*?

biafra23 (approved):

Issue #1566, opened 2021-07-06
Adjust client gw paths


elgatovital:
see https://github.com/gnosis/safe-client-gateway/issues/457

liliya-soroka:
Verified
app version 2.18.0 - 2966

PR #1571, opened 2021-07-07
Move chainId

biafra23:
Handles #1566

- Changes paths to be more RESTful as requested by backend
- **Do not merge before deployed on staging!**

@gnosis/mobile-devs

### DIFF
--- a/data/src/main/java/io/gnosis/data/backend/GatewayApi.kt
+++ b/data/src/main/java/io/gnosis/data/backend/GatewayApi.kt
@@ -33,41 +33,40 @@ interface GatewayApi {
         @Path("fiat") fiat: String = "usd"
     ): CoinBalances
 
-    @GET("v1/chains/{chainId}/transactions/{transactionId}")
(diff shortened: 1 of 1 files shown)


data/src/main/java/io/gnosis/data/backend/GatewayApi.kt
```diff
@@ -15,59 +15,58 @@ import java.util.*
 
 interface GatewayApi {
 
-    @GET("/{chainId}/v1/balances/supported-fiat-codes")
+    @GET("/v1/chains/{chainId}/balances/supported-fiat-codes")
```
jpalvarezl, line 18: This endpoint is chain independent. As we have a single common source for fiat conversion rates

Should be `/v1/balances/supported-fiat-codes`

jpalvarezl (approved):
I will approve to not block you but the only thing requiring a change is the `supported-fiat` endpoint as it is chain independent.

Issue #1557, opened 2021-07-05
Add dynamic ENS support


biafra23:
During Add safe flow
- Use ENS resolver address when selected chain resolver address is known

MouazAlzahabi:
[Backend ticket](https://github.com/gnosis/safe-config-service/issues/119#issuecomment-872103898)

biafra23:
Backend pull request: https://github.com/gnosis/safe-config-service/pull/143

liliya-soroka:
Reopened:
Add safe using ENS works on mainnet( alice.eth) but doesn't work on Rinkeby ( liltestsafe.eth) 

app: staging mainnet app
app version : 2.18.0-2966

liliya-soroka:
Verified
app version 2.18.0 -2983

PR #1573, opened 2021-07-07
Enable Ens based on ensRegistryAddress in ChainInfo

biafra23:
Handles #1557 

- Use ensRegistryAddress from `/v1/chains` response
- **If we merge this then ENS will be disabled until backend sends an ensRegistryAddress**

@gnosis/mobile-devs

### DIFF
--- a/data/src/main/java/io/gnosis/data/models/ChainInfo.kt
+++ b/data/src/main/java/io/gnosis/data/models/ChainInfo.kt
@@ -2,13 +2,16 @@ package io.gnosis.data.models
 
 import com.squareup.moshi.Json
 import com.squareup.moshi.JsonClass
+import pm.gnosis.common.adapters.moshi.DecimalNumber
+import java.math.BigInteger
+import io.gnosis.contracts.BuildConfig.ENS_REGISTRY
 
 
 @JsonClass(generateAdapter = true)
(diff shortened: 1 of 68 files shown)


data/src/main/java/io/gnosis/data/models/ChainInfo.kt
```diff
@@ -4,12 +4,14 @@ import com.squareup.moshi.Json
...
 import pm.gnosis.common.adapters.moshi.DecimalNumber
 import java.math.BigInteger
+import io.gnosis.contracts.BuildConfig.ENS_REGISTRY
 
 
 @JsonClass(generateAdapter = true)
 data class ChainInfo(
     @Json(name = "chainId") @field:DecimalNumber val chainId: BigInteger,
     @Json(name = "chainName") val chainName: String,
+    @Json(name = "ensRegistryAddress") val ensRegistryAddress: String? = ENS_REGISTRY,
```
elgatovital, line 14: this is probably wrong. We can't assume default ens registry would work on all chains

biafra23: This should not be there anymore. It was only for testing.

elgatovital (approved):

Issue #1563, opened 2021-07-06
Migrate push notification registration to client gateway


biafra23:
- use client gateway to subscribe to push notifications
- remove transaction gateway endpoint 

Backend ticket: https://github.com/gnosis/safe-client-gateway/issues/454

PR #1585, opened 2021-07-16
Push notification registration with multiple network support

elgatovital:
Handles #1563

- Adjust registration endpoint
- Adjust registration payload
- Add worker for registration work

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/notifications/NotificationRepository.kt
+++ b/app/src/main/java/io/gnosis/safe/notifications/NotificationRepository.kt
@@ -6,13 +6,11 @@ import io.gnosis.data.models.Owner
 import io.gnosis.data.models.Safe
 import io.gnosis.data.repositories.CredentialsRepository
 import io.gnosis.data.repositories.SafeRepository
-import io.gnosis.data.utils.toSignatureString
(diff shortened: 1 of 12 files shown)


app/src/main/java/io/gnosis/safe/notifications/NotificationRepository.kt
```diff
@@ -72,125 +65,102 @@ class NotificationRepository(
...
-                    Timber.d("notification service registration failure")
-                    deviceUuid = null
-                    handleCloudMessagingTokenIsLinkedToAnotherDeviceError(it)
-                    registrationUpdateFailed = true
+                kotlin.runCatching {
+                    notificationService.register(registration)
                 }
+                    .onSuccess {
+                        Timber.d("notification service registration success")
+                        deviceUuid = registration.uuid
```
DmitryBespalov, line 93: didn't we say that notification service won't return the uuid anymore and we would just generate it in the app and submit it?

elgatovital: it's not returned. Here locally generated uuid is saved if request is successful

DmitryBespalov (approved):

Issue #1544, opened 2021-06-24
[Android] Handle native currencies in the app 


biafra23:
For the assets tab we need to show amounts in the chains native currency. Also we need the decimals in several places. 
This information is provided by the client gateway at the endpoint: `/v1/chains`

**ATM the endpoint is missing a logo.** 

**UPDATE**:
**Logo uri is available at https://safe-client.staging.gnosisdev.com/v1/chains**

Backend ticket: https://github.com/gnosis/safe-client-gateway/issues/461

PR #1588, opened 2021-07-21
Handle native currencies of the chain

elgatovital:
Handles #1544

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/utils/TxUtils.kt
+++ b/app/src/main/java/io/gnosis/safe/utils/TxUtils.kt
@@ -1,7 +1,7 @@
 package io.gnosis.safe.utils
 
 import android.content.res.Resources
-import io.gnosis.data.BuildConfig
+import io.gnosis.data.models.Chain
 import io.gnosis.data.models.transaction.TransactionDirection
 import io.gnosis.data.models.transaction.TransactionStatus
 import io.gnosis.data.models.transaction.TransferInfo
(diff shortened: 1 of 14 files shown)


app/src/main/java/io/gnosis/safe/utils/TxUtils.kt
```diff
@@ -27,18 +27,18 @@ fun TransactionStatus.isCompleted(): Boolean =
...
+            balanceFormatter.formatAmount(txInfo.value, false, chain.currency.decimals, chain.currency.symbol)
         }
         is TransactionInfoViewData.Transfer -> {
             val incoming = txInfo.direction == TransactionDirection.INCOMING
             val decimals: Int = when (val transferInfo = txInfo.transferInfo) {
                 is TransferInfo.Erc20Transfer -> {
                     transferInfo.decimals ?: 0
                 }
-                is TransferInfo.EtherTransfer -> 18
+                is TransferInfo.NativeTransfer -> 18
```
DmitryBespalov, line 41: I may be misunderstanding the syntax, but shouldn't the native transfer decimals taken from the current network's native currency details?

DmitryBespalov (approved):

app/src/main/java/io/gnosis/safe/utils/TxUtils.kt
elgatovital: definitely, missed that one

PR #1590, opened 2021-07-21
Add guard section to advanced safe settings

elgatovital:
Handles #1491

- Add guard section to advanced safe settings
- Show guard for contracts >= 1.3.0

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/settings/safe/AdvancedSafeSettingsFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/settings/safe/AdvancedSafeSettingsFragment.kt
@@ -108,7 +108,7 @@ class AdvancedSafeSettingsFragment : BaseViewBindingFragment<FragmentSettingsSaf
             } ?: run { moduleLabel.visible(false) }
 
             val contractVersion = SemVer.parse(safeInfo.version)
-            if (contractVersion >= SemVer(1, 1, 3)) {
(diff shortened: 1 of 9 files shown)


app/src/main/java/io/gnosis/safe/ui/settings/safe/AdvancedSafeSettingsFragment.kt
```diff
@@ -103,19 +106,41 @@ class AdvancedSafeSettingsFragment : BaseViewBindingFragment<FragmentSettingsSaf
                     modulesContainer.addView(labeledAddress(module.value, module.name ?: getString(R.string.unknown_module), module.logoUri))
                 }
             } ?: run { moduleLabel.visible(false) }
+
+            val contractVersion = SemVer.parse(safeInfo.version)
+            if (contractVersion >= SemVer(1, 1, 3)) {
```
DmitryBespalov, line 111: the contract version seems to be strange, should it be 1.3.0?

DmitryBespalov (approved):

app/src/main/java/io/gnosis/safe/ui/settings/safe/AdvancedSafeSettingsFragment.kt
elgatovital: yes, it should

Issue #1490, opened 2021-06-07
[Android] The app is crashing if cancel Biometric-enabling is selected in prompt after calling the app from background


liliya-soroka:
**Describe the bug**
The app is crashed if cancel Biometric enabling is selected in prompt after calling the app from background

**To Reproduce**
Precondition:
the fingerprint is set up on device
the passcode is disabled in app

1. Go to app settings
2. enable passcode
3. set up passcode
4. send app to the background when the Biometric activation prompt is displayed
5. call the app from background and click "Cancel"
Current result: the app is crashed - https://console.firebase.google.com/u/0/project/safe-firebase-staging/crashlytics/app/android:io.gnosis.safe.internal/issues/afab68100d7f7399b3ff369aa6b4d10a?time=last-seven-days&sessionEventKey=60BE315B00640001524DD44C47C3419F_1549514436943634791

**Expected behavior**
The app should not be crashed

**Smartphone (please complete the following information):**
 - Device: Huawei P30
 - Android Version: 10
 - App Version 2.17.0 - 2741rc

**Environment**
staging, rinkeby 
**Additional context**

tschubotz:
marking as minor since this sounds like an edge case, however since it's a bug, certainly worth to fix.

biafra23:
@tschubotz Would it be ok to dismiss the dialog (an not activate biometrics) when the app goes to background?

liliya-soroka:
Verified
app version 2.19.0 ( 3047)

PR #1601, opened 2021-07-28
Only hide keyboard when still added to activity.

biafra23:
Handles #1490

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/RepeatPasscodeFragment.kt
```diff
@@ -144,7 +144,10 @@ class RepeatPasscodeFragment : BaseViewBindingFragment<FragmentPasscodeBinding>(
             findNavController().popBackStack(R.id.createPasscodeFragment, true)
         }
 
-        binding.input.hideSoftKeyboard()
+        if(isAdded) {
```
elgatovital, line 147: I would catch this issue earlier and prevent dismissCreatePasscodeFragment() from being triggered if fragment is not added. This check could be used in state observer
```kotlin
lifecycle.currentState.isAtLeast(Lifecycle.State.STARTED)
```
how come PasscodeViewModel.PasscodeSetup is observed in disposed fragment?

biafra23: The CreatePasscodeFragment is still visible. And needs to be dismissed. We also need to show the confirmation that the passcode was setup. 
But the keyboard is not visible when coming back from the locked screen. So it doesn't need to be hidden.

elgatovital (approved):

Issue #1511, opened 2021-06-15
[Android] Crash: HeimdallEncryptionManager.java line 283


liliya-soroka:
**Describe the bug**
5 users faced the issue via one day 
https://console.firebase.google.com/u/0/project/safe-firebase-mainnet/crashlytics/app/android:io.gnosis.safe/issues/29f6c56c8b8313597ee6719133a37544?time=last-seven-days&sessionEventKey=60C843FD0390000133F90AF3525F96F4_1552344471616535734

**To Reproduce**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/c7250339-72f8-492c-aabc-392d9bfbc726)

**Expected behavior**
No crash

**Smartphone (please complete the following information):**
 - Device: [e.g. Pixel 2]
 - Android Version: 11
 - App Version 2.17.0 

**Environment**
prod, mainnet
**Additional context**

liliya-soroka:
Verified on emulator :
disable /enable passcode
change passcode

liliya-soroka:
reopened:
the crash appears again on the prod in 2.17.1- 2799 version
https://console.firebase.google.com/u/0/project/safe-firebase-mainnet/crashlytics/app/android:io.gnosis.safe/issues/29f6c56c8b8313597ee6719133a37544?time=last-seven-days&sessionEventKey=60D333A303160001558569DEDC054F9D_1555422208066858614

liliya-soroka:
Unfortunatly the issue still exists in 2.18.0 ![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/d7c15dc3-bd41-4547-8026-89f1481e949e)

biafra23:
This is the underlying cause:
```
Caused by java.lang.IllegalStateException: At least one biometric must be enrolled to create keys requiring user authentication for every use
       at android.security.keystore.KeymasterUtils.addSids(KeymasterUtils.java:288)
       at android.security.keystore.KeymasterUtils.addUserAuthArgs(KeymasterUtils.java:352)
       at android.security.keystore.AndroidKeyStoreKeyPairGeneratorSpi.initialize(AndroidKeyStoreKeyPairGeneratorSpi.java:386)
       at java.security.KeyPairGenerator$Delegate.initialize(KeyPairGenerator.java:697)
       at java.security.KeyPairGenerator.initialize(KeyPairGenerator.java:438)
[...]

PR #1602, opened 2021-07-28
Prevent crash when no fingerprints enrolled although BIOMETRIC_SUCCESS was returned

biafra23:
Handles #1511

- catch IllegalStateException
- disable biometrics

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/RepeatPasscodeFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/RepeatPasscodeFragment.kt
@@ -25,6 +25,7 @@ import io.gnosis.safe.ui.settings.app.SettingsHandler
 import io.gnosis.safe.utils.CustomAlertDialogBuilder
 import pm.gnosis.svalinn.common.utils.showKeyboardForView
 import pm.gnosis.svalinn.common.utils.visible
+import timber.log.Timber
 import javax.inject.Inject
 
 class RepeatPasscodeFragment : BaseViewBindingFragment<FragmentPasscodeBinding>() {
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/RepeatPasscodeFragment.kt
```diff
@@ -84,7 +85,14 @@ class RepeatPasscodeFragment : BaseViewBindingFragment<FragmentPasscodeBinding>(
                             },
                             confirmCallback = { dialog ->
                                 viewModel.enableBiometry()
-                                viewModel.encryptPasscodeWithBiometricKey(viewAction.passcode)
+                                try {
+                                    //Some device still throw an IllegalStateException
+                                    // (due to no fingerprints enrolled) even though they return BIOMETRIC_SUCCESS
+                                    viewModel.encryptPasscodeWithBiometricKey(viewAction.passcode)
+                                } catch (e: Exception) {
+                                    settingsHandler.useBiometrics = false
```
elgatovital, line 93: let's add disableBiometry to viewModel and set useBiometrics to false there, like it is set to true in enableBiometry()

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/RepeatPasscodeFragment.kt
```diff
@@ -84,7 +85,14 @@ class RepeatPasscodeFragment : BaseViewBindingFragment<FragmentPasscodeBinding>(
                             },
                             confirmCallback = { dialog ->
                                 viewModel.enableBiometry()
```
elgatovital, line 87: I would place this call after viewModel.encryptPasscodeWithBiometricKey. This sequence seems more correct to me. Biometric is enabled only after all necessary work is done.

biafra23: This way `settingsHandler.useBiometrics = false` should not be necessary

app/src/main/java/io/gnosis/safe/ui/settings/app/passcode/RepeatPasscodeFragment.kt
```diff
@@ -83,8 +84,14 @@ class RepeatPasscodeFragment : BaseViewBindingFragment<FragmentPasscodeBinding>(
                                 dismissCreatePasscodeFragment()
                             },
                             confirmCallback = { dialog ->
-                                viewModel.enableBiometry()
-                                viewModel.encryptPasscodeWithBiometricKey(viewAction.passcode)
+                                try {
+                                    //Some device still throw an IllegalStateException
+                                    // (due to no fingerprints enrolled) even though they return BIOMETRIC_SUCCESS
+                                    viewModel.encryptPasscodeWithBiometricKey(viewAction.passcode)
```
elgatovital, line 90: encryptPasscodeWithBiometry and enableBiometry are always used together => the whole try catch block can and should be moved to a viewModel.  enableBiometry method could not only set the setting but also do the biometric key preparation work.

Issue #1622, opened 2021-08-06
Add chain_id info to the Feedback email body 


liliya-soroka:
Add chain_id info to the Feedback email body  
https://gnosisinc.slack.com/archives/CQB729LLF/p1626453093358300

liliya-soroka:
Verified
app version 3.0.0 - 3090rc

PR #1627, opened 2021-08-09
Add chain info to the feedback email body

elgatovital:
Handles #1622

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/settings/app/GetInTouchFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/settings/app/GetInTouchFragment.kt
@@ -116,7 +116,7 @@ class GetInTouchViewModel
         val text = StringBuilder()
         text.appendln("${context.getString(R.string.app_name)} v${BuildConfig.VERSION_NAME}")
         text.appendln(context.getString(R.string.feedback_safe, activeSafe?.address?.asEthereumAddressChecksumString() ?: ""))
-        text.appendln(context.getString(R.string.feedback_chain, activeSafe?.chain?.name ?: ""))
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/gnosis/safe/ui/settings/app/GetInTouchFragment.kt
```diff
@@ -116,6 +116,7 @@ class GetInTouchViewModel
         val text = StringBuilder()
         text.appendln("${context.getString(R.string.app_name)} v${BuildConfig.VERSION_NAME}")
         text.appendln(context.getString(R.string.feedback_safe, activeSafe?.address?.asEthereumAddressChecksumString() ?: ""))
+        text.appendln(context.getString(R.string.feedback_chain, activeSafe?.chain?.name ?: ""))
```
biafra23, line 119: Is it possible to have more than one chain with the same name? If so, then we should include the chain id.

biafra23 (approved):
Maybe the chain id would be useful too?

app/src/main/java/io/gnosis/safe/ui/settings/app/GetInTouchFragment.kt
elgatovital: Chain: <chain name> (<id>) ?

app/src/main/res/values/strings.xml
```diff
@@ -235,6 +235,7 @@
     <string name="feedback_subject">Feedback</string>
 
     <string name="feedback_safe">Safe address: %1$s</string>
+    <string name="feedback_chain">Chain: %1$s (%2$s)</string>
```
sche, line 238: Will it be "Chain:  (nil)" when no safe is selected @elgatovital ?

biafra23: No it is `Chain: (null)` :-)

Issue #1623, opened 2021-08-06
[Android] Eth icon is used instead of native currency logo on the tx details


liliya-soroka:
**Describe the bug**
Eth icon is used instead of native currency logo on the tx details

**To Reproduce**
1. add safe = 0x82FC6069FB9290F98c6ae02b6cfb7B6aB0ab97bf on the xDai network on prod
2. go to the history and open incoming +5xdai tx
Current result: xDai has Eth icon on the tx details

**Expected behavior**
The native currency logo should be used - nhttps://safe-client.gnosis.io/v1/chains/100/

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/17274b00-cab7-49e5-8e18-97160f6f4804)
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/26497637-aed5-41e8-938e-45633077a907)

**Smartphone (please complete the following information):**
 - Device: [e.g. Pixel 2]
 - Android Version: [e.g. Android 9]
 - App Version [e.g. 1.0.1]

**Environment**

**Additional context**

elgatovital:
request: 
https://safe-client.gnosis.io/v1/chains/100/transactions/ethereum_0x82FC6069FB9290F98c6ae02b6cfb7B6aB0ab97bf_0xf2b6b319786670e3084e866e8b26f90a3df74a10975374d30408a4c9a471eca0_0x50cde0408433f069
```json
{
  "executedAt": 1602852525000,
  "txStatus": "SUCCESS",
  "txInfo": {
    "type": "Transfer",
    "sender": {
      "value": "0x8ed44000983B5789798B43AF9d8d9C1e1e2093D0"
    },
    "recipient": {
      "value": "0x82FC6069FB9290F98c6ae02b6cfb7B6aB0ab97bf"
    },
    "direction": "INCOMING",
    "transferInfo": {
      "type": "NATIVE_COIN",
      "value": "5000000000000000000"
    }
  },
  "txData": null,
  "detailedExecutionInfo": null,
  "txHash": "0xf2b6b319786670e3084e866e8b26f90a3df74a10975374d30408a4c9a471eca0"
}
```

`logoUri` is missing in transferInfo

liliya-soroka:
Verified
app version 2.18.0  - 3083

PR #1628, opened 2021-08-09
Load native currency icon for native transfers using chain data

elgatovital:
Handles #1623

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsActionFragment.kt
```diff
@@ -93,7 +93,7 @@ class TransactionDetailsActionFragment : BaseViewBindingFragment<FragmentTransac
 
         address?.let {
             with(binding) {
-                content.addView(getTransferItem(it, amount ?: "", addressInfoIndex?.get(it.asEthereumAddressChecksumString())))
```
biafra23, line 96: Why do we pass the chain here as a parameter, when it is already a property in TransactionDetailsActionFragment?

biafra23 (approved):

Issue #1646, opened 2021-08-31
[Android] Ledger: Adjust key list


elgatovital:
Add badge with key type icon to the owner address blockie.

![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/56946992-ea51-4377-94c9-c4cde11fe2a6)
Zeplin: https://zpl.io/2pgOrmo

liliya-soroka:
waiting for [Android] Ledger: Connect Ledger device UI to check Ledger Nano X icon

Generated and imported key icons are checked - done

liliya-soroka:
Verified
app version 3.1.0 -3285rc

PR #1652, opened 2021-09-01
Show key type in owner list

biafra23:
Handles #1646 

- Add overlay icon in owner list

Day | Night
----| ------
Screenshot_20210901-173854 |Screenshot_20210901-173949

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/settings/owner/list/OwnerListAdapter.kt
```diff
@@ -107,12 +109,21 @@ class LocalOwnerViewHolder(private val viewBinding: ItemOwnerLocalBinding) : Bas
     }
 }
 
+private fun OwnerViewData.LocalOwner.getImageResForKeyType(): Int =
```
elgatovital, line 112: move it to OwnerViewData class? Also there is no need to have a single LocalOwner subclass of the OwnerViewData so the can be combined (leave only OwnerViewData)

elgatovital (approved):

app/src/main/java/io/gnosis/safe/ui/settings/owner/list/OwnerListAdapter.kt
biafra23: Done in 23fa5d6c135a5d7b5fddf596a4a5cf01894fcf4e

Issue #1640, opened 2021-08-31
[Android] Ledger: Enter name screen


elgatovital:
Adjust enter name screen for owner keys: 
- [ ] Set address blockie size to 56dp
- [ ] Add key type badge

![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/9e6bac1c-72d5-4261-848a-ce1c1f3b9970)
Zeplin: https://zpl.io/2vDYgL7

liliya-soroka:
waiting for [Android] Ledger: Connect Ledger device UI to test the task

liliya-soroka:
Verified
app version 3.1.0 -3285rc

PR #1655, opened 2021-09-02
Update "Enter owner name screen"

biafra23:
Handles #1640

- Add key type icon
- increase blockies size

Before | After
-------|------
Screenshot_20210902-184600 | Screenshot_20210903-090337

@gnosis/mobile-devs


elgatovital:
hex address should have 24dp marginTop (or margin bottom for blockie container)

app/src/main/res/layout/fragment_owner_name_enter.xml
```diff
@@ -56,16 +56,57 @@
...
+            android:id="@+id/key_type"
+            android:layout_width="24dp"
+            android:layout_height="24dp"
+            android:src="@drawable/ic_key_type_imported_16dp"
+            app:layout_constraintBottom_toBottomOf="@id/key_type_background"
+            app:layout_constraintEnd_toEndOf="@id/key_type_background"
+            app:layout_constraintStart_toStartOf="@id/key_type_background"
+            app:layout_constraintTop_toTopOf="@id/key_type_background" />
+
+        <androidx.constraintlayout.widget.Guideline
```
elgatovital, line 95: I don't think these guidelines are needed

app/src/main/java/io/gnosis/safe/ui/settings/owner/OwnerEnterNameFragment.kt
```diff
@@ -52,6 +55,7 @@ class OwnerEnterNameFragment : BaseViewBindingFragment<FragmentOwnerNameEnterBin
         with(binding) {
             newAddressBlockies.setAddress(ownerAddress)
             newAddressHex.text = ownerAddress.formatEthAddress(requireContext(), addMiddleLinebreak = false)
+            keyType.setImageResource(ownerType.getImageResForOwnerType())
```
elgatovital, line 58: ownerType.getImageResForOwnerType sounds redundant. Probably could be shortened to ownerType.getImageRes. Or even better with a property: ownerType.imageRes

biafra23: It could be a property on OwnerViewData but shouldn't be a property on Owner.Type. So I kept it as an extension function for both. Also because we will need it in more places for the next PR.

elgatovital (approved):

Issue #1649, opened 2021-08-31
[Android] Owner details screen UI


elgatovital:
Refactor owner details layout
![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/d204abf6-7a31-4b1a-90dd-452d45b8e576

WC section visible only for WC keys
As there is currently no support for WC on android there is no need to implement this section

liliya-soroka:
waiting for [Android] Ledger: Connect Ledger device UI to check Connected Ledger nano x key details page

liliya-soroka:
Verified
app version 3.1.0 -3285rc

PR #1656, opened 2021-09-06
Owner detail screen

biafra23:
Handles #1649

- Tracking added
- toString() for Owner Type
- Updated owner detail screen

Day | Night
----|-----
Screenshot_20210907-143859 |Screenshot_20210907-143917

@gnosis/mobile-devs


data/src/main/java/io/gnosis/data/models/Owner.kt
```diff
@@ -48,6 +48,14 @@ data class Owner(
                 else -> IMPORTED
             }
         }
+
+        override fun toString(): String =
+            when(value) {
+                0 -> "Imported"
```
elgatovital, line 54: we should use localized strings from strings.xml instead of hardcoded strings when they are displayed on screen

elgatovital (approved):

Issue #1648, opened 2021-08-31
[Android] Ledger: Connect Ledger device UI


elgatovital:
- [ ] Add screen for device search
- [ ] Show not found placeholder if results are empty
- [ ] Add pull to refresh for triggering another search

Following implementation details are handled in a separate issue (add placeholders, todo's where necessary, leave methods empty):
- Search for bluetooth devices
- Filter by uuid and present founded ledger nano x devices in the list to the user

![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/9907939e-463c-4bbc-adaa-2e7102e59059)
![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/e6f05a01-8b7d-4ecd-84c6-1b2dc70c18c8)
![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/cb769bcb-5d49-4d08-8c10-247877171de8)
Zeplin
Device list (search): https://zpl.io/29dZkgy
Device list: https://zpl.io/ady7mD7
Device list (not found): https://zpl.io/boAxWnZ

liliya-soroka:
Verified
app version 3.1.0 -3285rc
separate issue is created for the quick loading process before the pairing

Issue #1636, opened 2021-08-17
[Android] Ledger: Connecting & pairing


elgatovital:

liliya-soroka:
Verified
app version 3.1.0 -3285rc

Reported issues:
Critical issue - search for the device doesn't work on Huawei
majour - search for addresses happens without pairing

PR #1657, opened 2021-09-08
Ledger: Connect device UI & BLE

elgatovital:
Handles #1648
Handles #1636

- BLE: device scanning, writing, receiving data
- Ledger
  - APDU wrapping, unwrapping 
  - address command
  - receiving addresses
- Device list screen
- Adjusted navigation & connect ledger key flow
- Updated visuals & text

@gnosis/mobile-devs


app/src/main/AndroidManifest.xml
```diff
@@ -11,10 +11,19 @@
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.NFC" />
 
+    <uses-permission android:name="android.permission.BLUETOOTH"/>
+    <uses-permission android:name="android.permission.BLUETOOTH_ADMIN"/>
+    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
```
biafra23, line 16: Should we tell our users what we need this permission for?

app/src/main/AndroidManifest.xml
```diff
@@ -11,10 +11,19 @@
...
+
+
     <uses-feature
         android:name="android.hardware.camera.autofocus"
         android:required="false" />
 
+    <uses-feature
+        android:name="android.hardware.bluetooth_le"
```
biafra23, line 24: What if users don't have BLE and a Ledger Nano? Shouldn't they still be able to use our app without BLE?

elgatovital: definitely. This line is to be removed

app/src/main/AndroidManifest.xml
elgatovital: I think ledger intro screen does that

app/src/main/java/io/gnosis/safe/ui/settings/owner/ledger/LedgerController.kt
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    suspend fun getAddress(device: BluetoothDevice, path: String): Solidity.Address = suspendCoroutine {
+        ConnectionManager.writeCharacteristic(device, writeCharacteristic!!, wrapADPU(getAddressCommand(path)))
+        addressContinuation = it
+    }
+
+    suspend fun addressesForPage(derivationPath: String, start: Long, pageSize: Int): List<Solidity.Address> {
+        val addressPage = mutableListOf<Solidity.Address>()
+        for (i in start until pageSize) {
```
biafra23, line 202: What if I want to load the second page starting at pagesize+1? 
Shouldn't this be like:
```
for (i in start until start + pageSize) {
[...]
}
```

biafra23:
Maybe we can set the `PAGE_SIZE` to 1 so people don't have to wait longer than necessary? We still need a progress indicator. Not necessarily in this PR.

app/src/main/java/io/gnosis/safe/ui/settings/owner/ledger/LedgerController.kt
elgatovital: until is not inclusive

biafra23:
1. When waiting in the owner list until the ledger locks itself, I found no other way than restarting the application to reconnect to the ledger. It would not reconnect automatically on the list, not when I went back to to the info screen. (happened once, cannot reproduce it ATM)
2. When I go back from the owner list list to the connect screen, it would not list the ledger even though it is unlocked

app/src/main/java/io/gnosis/safe/ui/settings/owner/ledger/LedgerController.kt
```diff
@@ -0,0 +1,275 @@
...
+import android.content.Context
+import android.content.Intent
+import android.content.pm.PackageManager
+import android.os.Build
+import android.os.ParcelUuid
+import androidx.core.content.ContextCompat
+import androidx.fragment.app.Fragment
+import io.gnosis.safe.ui.settings.owner.ledger.LedgerWrapper.parseGetAddress
+import io.gnosis.safe.ui.settings.owner.ledger.LedgerWrapper.splitPath
+import io.gnosis.safe.ui.settings.owner.ledger.LedgerWrapper.unwrapADPU
```
biafra23, line 22: I think this is a type. Shouldn't it be called `APDU`? There is several places where it is called ADPU instead of APDU.

app/src/main/java/io/gnosis/safe/ui/settings/owner/ledger/LedgerController.kt
```diff
@@ -0,0 +1,275 @@
...
+            onCharacteristicRead = { _, characteristic -> }
+
+            onCharacteristicWrite = { _, characteristic -> }
+
+            onMtuChanged = { _, mtu -> }
+
+            onCharacteristicChanged = { _, characteristic ->
+                val unwrappedResponse = unwrapADPU(characteristic.value)
+                val address = parseGetAddress(unwrappedResponse)
+                Timber.d(address)
```
biafra23, line 91: I don't mind having debug log statements. But when seeing them in logcat, it would helpful to have some context to this address. For example: 
```
Timber.d("onCharacteristicChanged() | Parsed address: $address")
```

app/src/main/java/io/gnosis/safe/ui/settings/owner/ledger/LedgerController.kt
```diff
@@ -0,0 +1,275 @@
...
+                notifyingCharacteristics.add(characteristic.uuid)
+            }
+
+            onNotificationsDisabled = { _, characteristic ->
+                notifyingCharacteristics.remove(characteristic.uuid)
+            }
+        }
+    }
+
+    init {
```
biafra23, line 106: Is this the right place for `init` in this file?

app/src/main/java/io/gnosis/safe/ui/settings/owner/ledger/LedgerController.kt
```diff
@@ -0,0 +1,275 @@
...
+    var connectedDevice: BluetoothDevice? = null
+        private set
+
+    private var deviceConnectedCallback: DeviceConnectedCallback? = null
+    private var addressContinuation: Continuation<Solidity.Address>? = null
+
+    var writeCharacteristic: BluetoothGattCharacteristic? = null
+    var notifyCharacteristic: BluetoothGattCharacteristic? = null
+
+    fun loadDeviceCharacteristics() {
```
biafra23, line 48: Is this the right place in this file for a public function?

app/src/main/java/io/gnosis/safe/ui/settings/owner/ledger/LedgerController.kt
```diff
@@ -0,0 +1,272 @@
...
+        }
+
+        commandData.add(0xe0.toByte())
+        commandData.add(0x02.toByte())
+        commandData.add((if (displayVerificationDialog) 0x01.toByte() else 0x00.toByte()))
+        commandData.add((if (chainCode) 0x01.toByte() else 0x00.toByte()))
+        commandData.addAll(pathsData.toList())
+
+        val command = commandData.toByteArray()
+        Timber.d(command.toHexString())
```
biafra23, line 190: I think it would helpful to have some context to this hex string. For example: 
```
Timber.d("Get address command: ${command.toHexString()}")
```

app/src/main/java/io/gnosis/safe/ui/settings/owner/ledger/LedgerDeviceListAdapter.kt
```diff
@@ -0,0 +1,58 @@
...
+    interface DeviceListener {
+        fun onDeviceClick(position: Int)
+    }
+}
+
+abstract class BaseOwnerViewHolder(
+    viewBinding: ViewBinding
+) : RecyclerView.ViewHolder(viewBinding.root)
+
+class LedgerDeviceViewHolder(private val viewBinding: ItemLedgerDeviceBinding) : BaseOwnerViewHolder(viewBinding) {
```
biafra23, line 47: Don't we need a DisabledLedgerDeviceViewHolder?

app/src/main/java/io/gnosis/safe/ui/settings/owner/ledger/LedgerController.kt
elgatovital: why not?

app/src/main/java/io/gnosis/safe/ui/settings/owner/ledger/LedgerController.kt
elgatovital: where would be the right place?

app/src/main/java/io/gnosis/safe/ui/settings/owner/ledger/LedgerDeviceListAdapter.kt
elgatovital: What would this view holder show? What is a disabled ledger device?

app/src/main/java/io/gnosis/safe/ui/settings/owner/ledger/LedgerController.kt
elgatovital: APDU -> LedgerWrapper; APDU everywhere

app/src/main/java/io/gnosis/safe/ui/settings/owner/ledger/LedgerController.kt
elgatovital: this fun should not be public in the first place

app/src/main/java/io/gnosis/safe/ui/settings/owner/ledger/LedgerDeviceListAdapter.kt
biafra23: One that has already imported and named. It is possible to import more than one ledger owner address, isn't it?

app/src/main/java/io/gnosis/safe/ui/settings/owner/ledger/LedgerController.kt
biafra23: In the middle of private variable declarations? IIRC there was team decision for a certain order of properties and functions etc.

app/src/main/java/io/gnosis/safe/ui/settings/owner/ledger/LedgerDeviceListAdapter.kt
elgatovital: But this one is for the device list, and not the address list. You are most likely talking about next screen. Device can be either discovered - then it is shown, or it is not available - then it is missing in the list

app/src/main/java/io/gnosis/safe/ui/settings/owner/ledger/LedgerController.kt
elgatovital: I want to register the listener when creating an instance and the listener definition should come before init

app/src/main/java/io/gnosis/safe/ui/settings/owner/ledger/LedgerDeviceListAdapter.kt
biafra23: Sorry! I confused the two lists :-)

app/src/main/java/io/gnosis/safe/ui/settings/owner/ledger/transport/LedgerWrapper.kt
```diff
@@ -0,0 +1,65 @@
...
+        apdu.write(0x00)
+        apdu.write(0x00)
+        SerializeHelper.writeUint16BE(apdu, data.count().toLong())
+        apdu.write(data)
+        return apdu.toByteArray()
+    }
+
+    fun unwrapAPDU(data: ByteArray): ByteArray {
+        if (data.size <= 6 || data[0] != 0x05.toByte()) {
+            throw LedgerException(LedgerException.ExceptionReason.IO_ERROR, "invalid data size")
```
biafra23, line 24: Since this is not only about the size. Wouldn't `invalid data format `be more precise?

biafra23 (approved):

Issue #1667, opened 2021-09-14
Ledger: Save selected address & derivation path


elgatovital:
![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/6853d9fd-dcca-41ec-9e47-109e2d6ee109)

When Import is clicked:

- Pass selected address and derivation path to the owner name screen
- On owner name screen

  - When save owner is clicked create new owner object that contains selected address, derivation path, and entered name
  - Save owner to db
  - track owner added event
  - update firebase user properties

liliya-soroka:
tracking will be checked separately 
verified
app version 3.1.0 -3285rc

PR #1671, opened 2021-09-15
Safe ledger address with derivation path

biafra23:
Handles #1667

- 
-
-

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/settings/owner/OwnerEnterNameViewModel.kt
```diff
@@ -63,6 +63,27 @@ class OwnerEnterNameViewModel
...
     }
+
+    fun importLedgerOwner(address: Solidity.Address, name: String, derivationPathWithIndex: String) {
+        safeLaunch {
+            credentialsRepository.saveLedgerOwner(derivationPathWithIndex, address, name)
+            settingsHandler.showOwnerBanner = false
+            settingsHandler.showOwnerScreen = false
+            tracker.logLedgerKeyImported()
+            tracker.setNumKeysLedger(credentialsRepository.ownerCount(Owner.Type.LEDGER_NANO_X))
+            notificationRepository.registerSafes()
```
elgatovital, line 74: this is not necesary at the moment. Maybe just leave a TODO, so that we can address ledger key registration later.

app/src/main/java/io/gnosis/safe/ui/settings/owner/OwnerEnterNameFragment.kt
```diff
@@ -26,6 +27,7 @@ import pm.gnosis.utils.hexAsBigInteger
 import java.math.BigInteger
 import javax.inject.Inject
 
+@ExcludeClassFromJacocoGeneratedReport
```
elgatovital, line 30: are we not exluding files based on name? Can we exclude all *Fragment files from report?

elgatovital (approved):

Issue #1695, opened 2021-09-29
[Android]Error for not found addresses appears before the pairing is done


liliya-soroka:
**Describe the bug**
Error for not found addresses appears before the pairing is done
The search for the addresses happens before the pairing is done and the "failed" error appears to quickly. The user doesn't have time to check pairing message and react 

**To Reproduce**
1. Go to add owner
2. Connect Ledger nano X ( all required for the connection actions are done: bluethoot on, eth app is running )
3. Click on found Nano X
Current result: At the same time the search for the addresses is started and the message about connection appears partly at the top of the screen. In a few sec, no found error appears 

**Expected behavior**
The search for the addresses should not be started if the devices are not paired 
The user should be able to finish the pairing without the harry and than the addresses from the paired Ledger Nano should be displayed

**Screenshots**
https://drive.google.com/file/d/1QhegToA3NIHXpkhZt85OeqlT6BCVkRrZ/view?usp=sharing

**Smartphone (please complete the following information):**
 - Device: Xiaomi Redmi 10
 - Android Version: 11
 - App Version 3.1.0 -3285rc

**Environment**
staging 
**Additional context**

liliya-soroka:
Could we show the pairing dialogue without clicking on notification?
On Huawei P20 the notification is not displayed without opening the notification bar.
Pairing process on Huawei P20:
1. Connection error in app and no notification in app at all about pairing:
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/893392d9-a520-410c-beb8-41ff809a88d7)

2. If the user swaps down the notification bar there is a notification to start pairing( not obvious step when the error is displayed in app) :
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/ca6393a7-f9f5-4889-9d3f-be12180c0b48)

3. Click notification:
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/70503584-c638-4bd1-8ab1-9ddee41c9d87)

Could we show the pairing dialog directly in the app without the additional click on notification ?

PR #1712, opened 2021-10-11
Ask user to check pairing

biafra23:
Handles #1695

- Ask user to check device was paired
- Add bullets to list for device empty screen and address list empty screen

no devices | no addresses
----------- | -----------
image |  image

@gnosis/mobile-devs


elgatovital:
@biafra23 Please split message into 2 TextViews: 
first: No address could be loaded from Ledger Nano X
gravity should be center
For the second TextView same bullet points should be used as in Terms & Conditions:
image

Issue #1754, opened 2022-01-26
[Intercom] Push Notifications


DmitryBespalov:
Implement push notifications for incoming Intercom messages when the App is closed or in the background.

- [x] Display incoming messages
- [x] Opening the push message should open App and show Intercom chat screen in a popover
- [x] Create separate notification group for Intercom

biafra23:
Here is a list of situations we faced problems with Intercom on iOS:
- Opening from background
  - Via Push notification
  - From Home screen
  - App lock activated
  - App lock deactivated
- Opening after killing the app
  - Via Push notification
  - From Home screen
  - App lock activated
  - App lock deactivated

elgatovital:
I would keep the default Intercom notification icon for now

DmitryBespalov:
ok, then I'll move it to another issue as a new requirement for later.

liliya-soroka:
Verified
app version 3.4.0-3428rc

Issue #1752, opened 2022-01-26
[Intercom] Handle incoming messages


DmitryBespalov:
- [x] Subscribe to Intercom notifications
- [x] Display a ‘badge’ for unread messages
* Settings tab
* Settings → Chat with us
- [x] Hide `badge` when Intercom is opened
- [x] Disable default chat overlay

- [x] Only open Intercom window after passcode is unlocked
- [x] Make sure that the Intercom can be opened after minimizing and opening the app again

### Links
[Message indicator](https://www.figma.com/file/JrFUulz4n8Gb8Ayy3UEGhU/Intercom-implementation?node-id=10%3A456)

liliya-soroka:
@elgatovital , 
on staging "chat with us" doesn't work
on prod push is not enabled
blocked for testing for now
in 3.4.0 -3423rc

liliya-soroka:
Verified
app version 3.4.0 -3428rc

PR #1758, opened 2022-02-02
Receive and handle Itercom push notifications

elgatovital:
Handles #1752
Handles #1754 

- Receive and handle push notifications from Intercom
- Opening the push message should open App and show Intercom chat screen in a popover 
- Only open Intercom window after passcode is unlocked

Intercom android sdk creates it's own notification channels:
Screenshot_1643801231
Because Intercom android sdk has closed api there is no direct way to adjust naming/structure of those channels.
Same goes for the format of push notification itself
Screenshot_1643801460
It is possible to adjust the icon in the top left part though. I would see it as low priority now and propose to adjust it if needed in a future pr.
Here is example of a notification for a single message
Screenshot_1643801605

@gnosis/mobile-devs


DmitryBespalov (approved):
Looks good! 🚀

Issue #1770, opened 2022-08-11
Mini rebrand (new splash screen logo and new launcher icons) and updates for Play Store


biafra23:
We will need to submit an update to the Android app for release on Sep. 6
Make sure the app can be released and has the new logo (due on Friday, 12.Sep.)

- Replace app logo with new version
- Remove Gnosis in all visible places.
- Update targetSdkVerison to 32
- Update compileSdkVersion to 32
- Update Gradle to newest relase
- Update Gradle plugin to recommended version
- Update launcher icons and logo in splash screen

- Update Buildkite to also have all the changes to the build environment, that are not within the git repo

https://user-images.githubusercontent.com/246473/185446619-d64ee1e6-6f15-4904-b493-200a637f28c6.mp4

TanyaEfremova:
Here is the logo package! [https://app.zenhub.com/files/96440096/34305b44-ce17-49f7-93dc-63ac736e6822/download](https://app.zenhub.com/files/96440096/34305b44-ce17-49f7-93dc-63ac736e6822/download)

biafra23:
@TanyaEfremova This seems to be the Apple/iOS version of the Logo. Are you able to create the same for Android?

biafra23:
@TanyaEfremova Also: We need an vector image of the logo for newer devices.

biafra23:
At the moment the logo in the splash screen is a vectro log with a transparent background. If I replace this with the logo from the zip file it most like will look horrible.

TanyaEfremova:
Please, ignore the package above. Try this one: [https://app.zenhub.com/files/96440096/4c42255c-4105-4e00-9e72-0ecd45690850/download](https://app.zenhub.com/files/96440096/4c42255c-4105-4e00-9e72-0ecd45690850/download)

Here is the SVG for the splash screen. ![Black.svg](https://images.zenhubusercontent.com/6183a79a848bb3312854c951/607ece1c-3c2b-422d-bde0-1c7306ab5776)

TanyaEfremova:
Due to the discussion[ in this thread](https://5afe.slack.com/archives/C03CT23SU1Z/p1661345009270079), we will need to change the launcher icons temporarily.

Please, see [here](https://www.figma.com/file/rqIpWH6iJabWWTaMatlGqA/Android-UI-Kit-%E2%80%94-Rebrand?node-id=439%3A1983) - we will use the B&W version for the production version.

No new icons are necessary, just a replacement for now.

[...]

PR #1771, opened 2022-08-16
#1770 New Logo and updated Gradle plugin & targetSDK

biafra23:
Handles #1770 

- Update Grade version
- Update Gradle plugin version
- Disable kotlin.extensions (but not for the qr scanner as it nreaks with viewBindings)
- Remove Gnosis from user facing strings
- Update target SDK
- Add new launcher icon for dev/staging/prod
- Change splash screen logo

Preview:
mdpi API 24:
image

Pixel 3a api 30
image

mdpi looks like this now:
image

Splash-Screen
image

@gnosis/mobile-devs


qrscanner/src/main/java/io/gnosis/safe/qrscanner/QRCodeScanActivity.kt
```diff
@@ -8,9 +8,9 @@ import android.os.Bundle
 import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.app.AppCompatActivity
 import androidx.fragment.app.Fragment
-import kotlinx.android.synthetic.main.screen_scan.*
 import pm.gnosis.svalinn.utils.ethereum.ERC67Parser
 import pm.gnosis.utils.asEthereumAddress
+import kotlinx.android.synthetic.main.screen_scan.*
```
elgatovital, line 13: did you want to remove synthetic?

elgatovital (approved):

qrscanner/src/main/java/io/gnosis/safe/qrscanner/QRCodeScanActivity.kt
biafra23: I could not remove it from QRScanner because it would break. We need a separate ticket. Also for migrating it out of its own module. I will make such a ticket.

Issue #1798, opened 2022-09-02
[android] No "what's new"screen if the old version has passcode or biometry check set up


liliya-soroka:
**Describe the bug**
No "what's new"screen is the old version has passcode set up

**To Reproduce**
1. Set up previous release
2. set up passcode ( with and without biometric protection)
3. update the app to the latest release
Current result: No "What's new" screen

**Expected behavior**
What's new screen should be displayed if it was set up for new app version 

**Smartphone (please complete the following information):**
 - Device: pixel 5
 - Android Version: android 12
 - App Version 3.4.0-3503rc

**Environment**

**Additional context**

liliya-soroka:
Verified
app version 3.4.0 - 3506rc

PR #1799, opened 2022-09-02
Adjust what's new showing conditions

elgatovital:
Handles #1798

- Ensure what's new is shown after pin input

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
@@ -42,7 +42,8 @@ import pm.gnosis.utils.asEthereumAddress
 import java.math.BigInteger
 import javax.inject.Inject
 
-class StartActivity : BaseActivity(), SafeOverviewNavigationHandler, AppStateListener, UnreadConversationCountListener {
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
```diff
@@ -362,11 +407,8 @@ class StartActivity : BaseActivity(), SafeOverviewNavigationHandler, AppStateLis
         }
     }
 
-    private fun showWhatsNewIfNeeded() {
```
biafra23, line 365: Why not keep this check in one place? Maybe with an optional flag to check the lifecycle? Or put the lifecycle check around  `showWhatsNewIfNeeded()`

biafra23 (approved):
Works as expected

app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
elgatovital: it won't work if passcode is not required. In this case activity is not yet resumed, but we still need to show the dialog

Issue #1780, opened 2022-08-29
[Send funds] Total balance panel


elgatovital:
[ ] Add panel above cons and collectibles tab for displaying total balance
[ ] Remove total balance item from the coins list

![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/4cdb9d8e-2dc5-4f62-a984-9021518bb8b3)

[ ] show safe info screen when clicking `Receive`
![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/19b44fa1-9df4-4657-a0d9-3f78e2eabaac)

design: https://www.figma.com/file/FH6kqYzmW99z20B6h6Sx5W/Initiate-assets-transer?node-id=701%3A2657

liliya-soroka:
Verified
app version 3.4.0 -3576

PR #1803, opened 2022-09-27
[Send Funds] Total balance header

elgatovital:
Handles #1780

- Add header for total balance
- Remove total balance from coins list
- Show share safe dialog when receive button is clicked

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/res/layout/fragment_assets.xml
+++ b/app/src/main/res/layout/fragment_assets.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
-<androidx.coordinatorlayout.widget.CoordinatorLayout
-    xmlns:android="http://schemas.android.com/apk/res/android"
+<androidx.coordinatorlayout.widget.CoordinatorLayout xmlns:android="http://schemas.android.com/apk/res/android"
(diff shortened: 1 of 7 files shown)


app/src/main/res/layout/fragment_assets.xml
```diff
@@ -10,7 +10,152 @@
...
+                android:textSize="14dp"
+                app:layout_constraintBottom_toBottomOf="parent"
+                app:layout_constraintEnd_toStartOf="@+id/receive_button"
+                app:layout_constraintHorizontal_chainStyle="packed"
+                app:layout_constraintStart_toStartOf="parent"
+                app:layout_constraintTop_toBottomOf="@+id/total_balance_value">
+
+                <aapt:attr name="android:background">
+
+                    <selector>
```
biafra23, line 68: Why is it embedded? Wouldn't it make sense to have it in a separate file in `drawable`?

elgatovital: it is used only here. We don't have any other buttons with this style. For now having everything in one place makes adjuting and editing easier.

biafra23: image

Issue #1782, opened 2022-08-29
[Send funds] Send success screen


elgatovital:
Add screen for showing transaction submitted status.

[ ] Dismiss when clicking on `Done`
[ ] Show transaction details when clicking `View details`

![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/c787af60-bc9a-4ef7-8155-68f73bbe54a2)

design: https://www.figma.com/file/FH6kqYzmW99z20B6h6Sx5W/Initiate-assets-transer?node-id=701%3A2528

TanyaEfremova:
Use these .json files for the animation (light & dark modes): 
[https://app.zenhub.com/files/96440096/8ef06608-7a49-4ba0-b5c5-aae9931f4500/download](https://app.zenhub.com/files/96440096/8ef06608-7a49-4ba0-b5c5-aae9931f4500/download)

PR #1817, opened 2022-10-10
[Send funds] Success screen

elgatovital:
Handles #1782

- Success screen with animated visuals

@gnosis/mobile-devs


app/src/main/res/values/strings.xml
```diff
@@ -349,6 +349,11 @@
     <string name="tx_status_failed">Failed</string>
     <string name="tx_status_success">Success</string>
 
+    <string name="tx_submitted">Your transaction is submitted!</string>
+    <string name="tx_submitted_desc">It normally takes some time for a transaction to be executed.</string>
```
biafra23, line 353: This is probably the text from the specification. But it sounds like no further action is necessary. Shouldn't we tell the user that it needs to executed on web or ios?

elgatovital: @DmitryBespalov @TanyaEfremova wdyt?

DmitryBespalov: This is what used by iOS:

```swift
        let title = "Your transaction is queued!"
        let body = "Your request to send \(formattedAmount) \(token) is submitted and needs to be confirmed by other owners."
```

PR #1825, opened 2022-10-13
Release 3.5.0

elgatovital:
Rebranding without send funds

@gnosis/mobile-devs


biafra23 (approved):
Do we even need the changes for the what's new screen? For the next release they would be different anyway, nor?

elgatovital:
> Do we even need the changes for the what's new screen? For the next release they would be different anyway, nor?

Probably not, but maybe we will reuse some items later to remind user about what was done. I would keep strings for now and we can clean it up later.

Issue #1839, opened 2022-10-28
[Android] [send funds]Open Add key on "send" if no added owner key


liliya-soroka:
**Bug description**
Open Add key on "send" if no added owner key
**Steps To Reproduce**
1.There is no added owner key
2. click send funds
Current result: send funds flow is started

**Expected Result** Add owner key flow should be started
Check on ios 
**Device & App version**
 - Device/OS: android
 - App version: 3.4.0 -3652
 
**Environment**
prod
**Additional info**
(Optional)

elgatovital:
fyi: https://app.zenhub.com/workspaces/mobile-engineering-60cc78740f488d000e996335/issues/safe-global/safe-android/1845

liliya-soroka:
Verified taking into account the comment above
app version  3.5.0 -3676

PR #1844, opened 2022-11-01
[Send funds] Add owner key first

elgatovital:
Handles #1839

- Add owner key first screen

image

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/safe/send_funds/AddOwnerFirstFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/safe/send_funds/AddOwnerFirstFragment.kt
@@ -1,0 +1,44 @@
+package io.gnosis.safe.ui.safe.send_funds
+
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.navigation.fragment.findNavController
+import io.gnosis.safe.R
+import io.gnosis.safe.ScreenId
+import io.gnosis.safe.databinding.FragmentAddOwnerFirstBinding
+import io.gnosis.safe.di.components.ViewComponent
+import io.gnosis.safe.ui.base.fragment.BaseViewBindingFragment
+import java.math.BigInteger
+
+class AddOwnerFirstFragment : BaseViewBindingFragment<FragmentAddOwnerFirstBinding>() {
+
+    override fun screenId() = ScreenId.ASSETS_COINS_TRANSFER_ADD_OWNER
+
+    override suspend fun chainId(): BigInteger? = null
+
+    override fun inject(component: ViewComponent) {
+        component.inject(this)
+    }
+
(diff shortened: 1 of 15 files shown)


biafra23:
When I go through adding an owner, I am back at square one and have to restart the whole send funds flow. Also the imported key is not recognized and I have t go through the add key flow again. I would expect to be directed to the "select asset screen" after adding one key. 

The owner is recognized after going to the settings screen.

app/src/main/java/io/gnosis/safe/ui/safe/send_funds/AddOwnerFirstFragment.kt
```diff
@@ -0,0 +1,44 @@
...
+import android.view.ViewGroup
+import androidx.navigation.fragment.findNavController
+import io.gnosis.safe.R
+import io.gnosis.safe.ScreenId
+import io.gnosis.safe.databinding.FragmentAddOwnerFirstBinding
+import io.gnosis.safe.di.components.ViewComponent
+import io.gnosis.safe.ui.base.fragment.BaseViewBindingFragment
+import java.math.BigInteger
+
+class AddOwnerFirstFragment : BaseViewBindingFragment<FragmentAddOwnerFirstBinding>() {
```
biafra23, line 15: Do we need to track when AddOwnerFirstFragment was opened? Maybe not a question for this PR. But where would I put it?

data/src/main/java/io/gnosis/data/repositories/SafeRepository.kt
```diff
@@ -64,23 +64,56 @@ class SafeRepository(
...
+    }
+
+    suspend fun setActiveSafeSigningOwners(owners: List<Solidity.Address>) {
+        preferencesManager.prefs.edit {
+            putString(
+                ACTIVE_SAFE_SIGNING_OWNERS,
+                if(owners.isEmpty()) {
+                    null
+                } else {
+                    owners.joinToString(";") {
```
biafra23, line 104: Should ";" be extarcted as for example `ADDRESS_SEPARATOR` constant?

biafra23 (approved):
LGTM 🚀

biafra23:
Is it ok not to continue to the select asset screen after adding the owner keys has been added?

app/src/main/java/io/gnosis/safe/ui/safe/send_funds/AddOwnerFirstFragment.kt
elgatovital: it is being tracked already. This fragment has screen id `screen_add_owner_to_send_assets`

data/src/main/java/io/gnosis/data/repositories/SafeRepository.kt
elgatovital: yes, as it is used in multiple places

Issue #1849, opened 2022-11-07
[android][send funds] Missed requirement on the send funds screen 


liliya-soroka:
The next requirement from the  https://app.zenhub.com/workspaces/mobile-engineering-60cc78740f488d000e996335/issues/safe-global/safe-android/1784 is missing " Also REVIEW action is available in the toolbar (in the right cornter; same enabled/disabled logic)" 

**Smartphone (please complete the following information):**
 - Device: pixel 5
 - App Version 3.5.0-3676

**Environment**
prod
**Additional context**

liliya-soroka:
Verified
app version 3.5.0 - 3687

Issue #1840, opened 2022-10-28
[android][send funds] Scan qr code for the recipient address doesn't work 


liliya-soroka:
**Describe the bug**
Scan qr code for the recipient address doesn't work 

**To Reproduce**
1. Go to send funds
2. Click recipient address field
3. Scan QR code
4. scan EOA or safe QR code
Current result: the scanning screen is closed but the address is not detected in the recipient field

**Expected behavior**
make sure that it's possible to scan valid qr code

**Smartphone (please complete the following information):**
 - Device: pixel 5
 - Android Version: [android 
 - App Version 3.4.0 -3652

**Environment**
prod
**Additional context**

liliya-soroka:
Reopened:
can not scan the QR code from the MM ( any networks)

app version 3.5.0 -3676

liliya-soroka:
Verified
app version 3.5.0 - 3687

PR #1850, opened 2022-11-07
Scan MM qr codes

elgatovital:
Handles #1840
Handles #1849

- Add review toolbar button to send asset screen
- Fix scanning MM qr codes

@gnosis/mobile-devs


biafra23 (approved):
I could'nt reproduce the scanner issue (#1840). But this doesn't break anything so: LGTM

Issue #1857, opened 2022-11-25
Collectibles v2 endpoint integration


elgatovital:
- Use collectibles v2 data endpoint
- Implement list pagination

liliya-soroka:
Verified
app version 3.6.0-3699 (3699)

PR #1858, opened 2022-11-25
Paginated collectibles list

elgatovital:
Handles #1857

- Use new endpoint for getting collectibles data
- Paginated collectibles list

@gnosis/mobile-devs


data/src/test/java/io/gnosis/data/repositories/TokenRepositoryTest.kt
```diff
@@ -125,51 +123,6 @@ class TokenRepositoryTest {
         assertEquals("4.021".toBigDecimal(), balances.items[2].fiatBalance)
     }
 
-    @Test
-    fun `loadCollectiblesOf (address) should return grouped list of collectibles`() = runBlocking {
```
biafra23, line 129: Should we have a new test instead of this?

app/src/main/java/io/gnosis/safe/ui/assets/collectibles/paging/CollectiblePagingProvider.kt
```diff
@@ -0,0 +1,29 @@
...
+        ).flow
+    }
+
+    override fun toString(): String {
+        return "CollectiblePagingProvider(tokenRepository=$tokenRepository)"
+    }
+
+    companion object {
+        private const val NETWORK_PAGE_SIZE = 50
+    }
```
biafra23, line 28: NETWORK_PAGE_SIZE is ignored, isn't it?

biafra23 (approved):
LGTM 🚀

data/src/test/java/io/gnosis/data/repositories/TokenRepositoryTest.kt
elgatovital: no, because we are not grouping collectibles anymore and rely on backend to provide them in correct order

Issue #1795, opened 2022-09-01
[android] Can not connect Ledger nano X on Google Pixel 3a/5/6


liliya-soroka:
**Describe the bug**
Can not connect Ledger nano X on android

**To Reproduce**
1. Install the app
2. add safe and try to add Ledger Nano X as a key
3. approve Location permissions
Current result: The app is closed

4. restart the app 
5. start the connect ledger nano x flow
Current result: Issue with connection screen is displayed all the time

**Expected behavior**
It should be possible to connect Ledger Nano X on Google Pixel

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/2153f49f-15fc-48e9-91b6-50f512eea595)

**Smartphone (please complete the following information):**
 - Device: google Pixel  5
 - Android Version: android 12
 - App Version 3.4.0

**Environment**
prod
**Additional context**

biafra23:
This seems to be device specific. When this is to be tackled, can we have access to a Pixel 5, please?

bogdoslavik:
Same for Galaxy S21 5G. After app rebranding it does not see my Ledger Nano X. Tried everything: reboot, safe removal, ledger unpair and pair again. Ledger Live works well.
Can provide logs / data you needs.
Please fix.

javalasers:
Had the same issue with a pixel 6. Just upgraded to pixel 7 and have the same issue on the 7 as well. The pixel 6 worked for a time, but stopped working within the last few months. (Sorry I don't have a precise time).

biafra23:
Works on Android 8 / Nexus 5x
Devices not visible on Pixel 3a (Android 12)
Device visible but cannot be connected on Fairphone 5 with Anddroid 11
Same behaviour on current version in Play store  (3.4.0 and 3.3.1) .

biafra23:
On the Pixel 3a the Nanos shows up (but retrieving keys fails) on build from sha-1 433ee7d9e62f1bb59f96740d57d9f6e7aedbbe96

markusbkoch:
Same issue, Pixel 7 Pro, Android 13

gpersoon:
I just checked the log via `adb logcat` and see the following:
```
[...]

PR #1860, opened 2022-12-03
add BLUETOOTH_SCAN permission

markusbkoch:
Handles #1795 

- add BLUETOOTH_SCAN permission as per https://github.com/safe-global/safe-android/issues/1795#issuecomment-1326625501
-
-

@gnosis/mobile-devs


biafra23 (dismissed):
This will not fix our problems for all devices. But is definitely necessary.

biafra23:
Unfortunately it is not enough to add this permission to the AndroidManifest. It also needs to be requested at runtime.

biafra23:
Replaced by: https://github.com/safe-global/safe-android/pull/1892

PR #1864, opened 2022-12-21
Release 3.6.1

elgatovital:
@gnosis/mobile-devs


leobond2021 (approved):

leobond2021 (approved):

leobond2021 (approved):

app/proguard-rules.pro
```diff
@@ -73,6 +73,9 @@
 -keep @interface pm.gnosis.heimdall.data.adapters.DecimalNumber
 -keep @interface pm.gnosis.heimdall.data.adapters.BigDecimalNumber
 
+-keep interface pm.gnosis.ethereum.rpc.models.** { *; }
+-keep class pm.gnosis.ethereum.rpc.models.** { *; }
```
biafra23, line 77: Just out of curiosity: Why do we need to keep this now and not before? What changed?

app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
```diff
@@ -416,8 +415,14 @@ class StartActivity : BaseActivity(), SafeOverviewNavigationHandler, AppStateLis
...
 
+    private fun onAppUpdated() {
+        showWhatsNew()
+        settingsHandler.showWhatsNew = false
+    }
+
     private fun showWhatsNew() {
-        navController.navigate(R.id.whatsNewDialog)
+        //TODO: uncomment if whats new screen should be shown
+        //navController.navigate(R.id.whatsNewDialog)
```
biafra23, line 425: Why do we need to do this manually?

build.gradle
```diff
@@ -35,11 +35,11 @@ subprojects {
     afterEvaluate { project ->
         if (project.hasProperty("android")) {
             android {
-                compileSdkVersion 32
+                compileSdkVersion 33
 
                 defaultConfig {
                     minSdkVersion 21
-                    targetSdkVersion 32
+                    targetSdkVersion 33
```
biafra23, line 42: Do we have enough Android 13 devices to test on? My Pixel 3a doesn't run Android 13. Latest is Android 12. Maybe I need a new device to develop on?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -4,4 +4,3 @@ distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
 distributionUrl=https\://services.gradle.org/distributions/gradle-7.3.3-all.zip
-x
```
biafra23, line 7: Woot?

biafra23 (approved):
I would not go to targetSdkVersion 33 if we do not have to. 
If we do, then I need an Android 13 capable device for development.

app/proguard-rules.pro
elgatovital: We have updated moshi and kotlin versions. My assumption is that either class generation syntax changed, which lead to this obfuscation error or that ens resolution was not working before in the first place.

app/src/main/java/io/gnosis/safe/ui/StartActivity.kt
elgatovital: we do not have version based logic yet. What's new is always shown for the new version.

build.gradle
elgatovital: you could use Android 13 emulator for now. Should be enough.

gradle/wrapper/gradle-wrapper.properties
elgatovital: that's a harmless typo. Cleanup is coming.

Issue #1879, opened 2023-01-05
Can not add safe or recipient on Send funds using ENS name ( prod app)


liliya-soroka:
**Describe the bug**
Can not add safe using ENS name ( prod app)

**To Reproduce**
1. Clean install prod app
2. Try to add safe using ENS name ( safetest2.eth for Goerli)
Current result: Error message is displayed and the safe is not defined. I don't have such problem in the internal app

**Expected behavior**
Make sure that it's possible to add safe using ENS name

**Smartphone (please complete the following information):**
 - Device: Pixel 5
 - Android Version:android 13
 - App Version 3.6.1-3787rc

**Environment**
prod
**Additional context**

elgatovital:
@liliya-soroka could you check if there was a leading white space when you were entering safetest2.eth? `safetest2.eth` works for me, but ` safetest2.eth` currently is handled as an invalid address. 
<img width="266" alt="image" src="https://user-images.githubusercontent.com/17750984/210802742-0a269811-de23-418f-bfb4-1b9097d55c5c.png">
Is this the same error you were getting?

liliya-soroka:
Verified
app version 3.6.1-3801rc

Issue #1878, opened 2023-01-05
No enable push notification on app start


liliya-soroka:
**Describe the bug**
No enable push notification on app start

**To Reproduce**
1. Clean app install
2. start app
Current result: There is no enable notifications request on app start. By default notifications are disabled 
**Expected behavior**
The user should be asked to enable push notifications .

**Smartphone (please complete the following information):**
 - Device: pixel 5
 - Android Version: android 13
 - App Version 3.6.1-3787rc

**Environment**
prod
**Additional context**

PR #1881, opened 2023-01-05
RC Fixes

elgatovital:
Handles #1878
Handles #1879
Handles #1882

- Use Goerli for default chain values
- Remove obsolete rinkeby flavor directory
- Trim ENS input (e.g. when pasting from clipboard)
- Request push notification permsission on Android 13

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/dialogs/EnsInputDialog.kt
```diff
@@ -124,8 +138,12 @@ class EnsInputDialog : BaseViewBindingDialogFragment<DialogEnsInputBinding>() {
             binding.successViews.visible(false)
             binding.dialogEnsInputUrlLayout.isErrorEnabled = false
             if (text.toString().isNotEmpty()) {
-                binding.dialogEnsInputProgress.visible(true)
-                job = onUrlChanged(text.toString())
+                if (text.toString().startsWith(" ")) {
+                    binding.dialogEnsInputUrl.setText(text.toString().trim())
+                } else {
```
biafra23, line 143: Whay do we not assign the job  if the Url starts with a " "? 
Why not call trim on any input? 
What if the last character is a " "?

elgatovital: - ens resolution fails if address is surrounded by spaces.
- we cannot trip on any input as it will lead to infinite cycle.
- with newest change trailing spaces will also be trimmed.

elgatovital: actually trimming trailing spaces is not necessary as spaces could be part of the ens name. See https://docs.ens.domains/frequently-asked-questions#what-about-foreign-characters-what-about-upper-case-letters-is-any-unicode-character-valid

biafra23: Why would trimming everything lead to an infinite loop? 
Doesn't the section in the docs mean we should apply nameprep to the input before sending it to the resolver?

biafra23: The weird thing now is that leading spaces are deleted automatically removed and trailing spaces show an error.

biafra23 (approved):
👍🏻

PR #1892, opened 2023-02-13
Fix Ledger device scanning for Android 12 devices

biafra23:
Handles #1795 

- Added BLUETOOTH_SCAN runtime permissions
- Ask for BLUETOOTH_SCAN on devices Android S and above and FINE location for below S
- Added debug logging

Does not fix the issue of listing the addresses on some devices. For example the Google Pixel 3a with Android 12. On the Pixel 3a the request to change the MTU to 517 or 156 or any other value above 20 fails and the communication is impossible. Even the Ledger Live Android  app does not work on these phones. 

Importing and signing works:
- Fairphone 4 with Android 12
- Samsung Galaxy S23 with Android 12
- Google Pixel 6 Pro with Android 13

Importing does NOT work
- Google Pixel 3a with Android 12

@gnosis/mobile-devs


app/src/main/java/io/gnosis/safe/ui/settings/owner/ledger/LedgerController.kt
```diff
@@ -173,13 +173,19 @@ class LedgerController(val context: Context) {
...
             promptEnableBluetooth(fragment)
         } else {
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && !context.hasPermission(Manifest.permission.ACCESS_FINE_LOCATION)) {
-                missingLocationPermissionHandler.invoke()
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M
+                && Build.VERSION.SDK_INT < Build.VERSION_CODES.S
+                && (!context.hasPermission(Manifest.permission.ACCESS_FINE_LOCATION))
+            ) {
+                missingPermissionHandler.invoke()
+            } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S
```
elgatovital, line 185: if and else if branches could be combined into one check because in both cases same statement is executed:
missingPermissionHandler.invoke()

elgatovital: I would extract checking logic into a subroutine: checkPermissionsMissing(): Bool
if (checkPermissionsMissing) {
  missingPermissionHandler.invoke()
}

elgatovital: even better to add this to the missingPermissionHandler as a method

elgatovital: It could be added as a callback to the function signature also:
missingPermissionChecker: () -> Bool

app/src/main/java/io/gnosis/safe/ui/settings/owner/ledger/LedgerDeviceListFragment.kt
```diff
@@ -184,8 +183,8 @@ class LedgerDeviceListFragment : BaseViewBindingFragment<FragmentLedgerDeviceLis
         super.onRequestPermissionsResult(requestCode, permissions, grantResults)
         viewModel.handlePermissionResult(
             this,
-            ::handleMissingLocationPermission,
-            ::requestMissingLocationPermission,
+            ::handleMissingPermission,
```
elgatovital, line 186: what is the difference btw :handleMissingPermission and ::requestMissingBLEPermission? Do we need this separation?

app/src/main/java/io/gnosis/safe/HeimdallApplication.kt
```diff
@@ -38,7 +38,7 @@ class HeimdallApplication : MultiDexApplication(), ComponentProvider {
         super.onCreate()
         registerActivityLifecycleCallbacks(activityLifecycleCallbacks())
         if (BuildConfig.DEBUG) {
-            Timber.plant(DebugTree())
+            Timber.plant(LineNumberDebugTree())
```
elgatovital, line 41: 👍

elgatovital (approved):

Issue #1933, opened 2023-05-30
[Android] [ExecTx] Review Transaction Execution Screen UI


elgatovital:
Scrollable screen with the header sections and execution options sections, with “Submit” button.

[Review Screen Figma iOS](https://www.figma.com/file/BHlmCT7Qp3jDpBaK39Vh7D/Execute-transactions?node-id=1255%3A11562)

<img width="200" src="https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/fb73ecef-41b9-4a55-86f0-4673455ae356" />

liliya-soroka:
Please, change the colours according to the https://5afe.slack.com/archives/C03CT23SU1Z/p1692804652054589?thread_ts=1692802811.511989&cid=C03CT23SU1Z

liliya-soroka:
Verified
app version 3.9.0 -4056rc

PR #1945, opened 2023-06-22
Review transaction execution UI

elgatovital:
Handles #1933

- Load chain l2 flag and features
- DB migration
- Add review transaction execution UI
- Adjust tx details & enable execution button
- Extend nav graph
- Cusom view for select key item
- Loading native token balances via rpc
- abi definitions for different safe contract versions

Not part of this PR: estimation & submission

@gnosis/mobile-devs


biafra23:
Not sure this is part of this PR. But the submit button on the  execute transaction screen is not enabled. Even though an owner key with sufficient funds is imported. 
image

elgatovital:
> Not sure this is part of this PR. But the submit button on the execute transaction screen is not enabled. Even though an owner key with sufficient funds is imported. 
execution itself is not part of this pr. Submit button will remain disabled.

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
```diff
@@ -94,6 +101,14 @@ class TransactionDetailsViewModel
                 }
     }
 
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    suspend fun canBeExecutedFromDevice(
+        executionInfo: DetailedExecutionInfo.MultisigExecutionDetails,
+        localOwners: List<Owner>
+    ): Boolean {
+        return localOwners.isNotEmpty() && executionInfo.confirmations.size == executionInfo.confirmationsRequired
```
biafra23, line 109: We're not supposed to check the balances here as well?

biafra23 (approved):
LGTM

Issue #1960, opened 2023-08-18
[Android] [ExecTx] Keystone execution flow


elgatovital:
Adjust execution flow for keystone hardware wallet

liliya-soroka:
Verified
app version 3.8.0-4008

PR #1961, opened 2023-08-18
Adjust execution flow for keystone keys

elgatovital:
Handles #1960

- Ask for password when required before execution
- Adjust execution flow for keystone keys
- Fix crash when trying to use hd keys for push registration

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/execution/TxReviewFragment.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/execution/TxReviewFragment.kt
@@ -25,13 +25,16 @@ import io.gnosis.safe.ui.base.BaseStateViewModel.ViewAction.NavigateTo
 import io.gnosis.safe.ui.base.BaseStateViewModel.ViewAction.ShowError
 import io.gnosis.safe.ui.base.SafeOverviewBaseFragment
 import io.gnosis.safe.ui.base.fragment.BaseViewBindingFragment
+import io.gnosis.safe.ui.settings.owner.details.OwnerDetailsFragment
 import io.gnosis.safe.ui.transactions.details.SigningMode
 import io.gnosis.safe.ui.transactions.details.viewdata.TransactionInfoViewData
 import io.gnosis.safe.utils.BalanceFormatter
 import io.gnosis.safe.utils.ParamSerializer
 import io.gnosis.safe.utils.formattedAmount
+import io.gnosis.safe.utils.getFromCurrent
 import io.gnosis.safe.utils.logoUri
 import io.gnosis.safe.utils.setLink
+import io.gnosis.safe.utils.setToCurrent
 import io.gnosis.safe.utils.toColor
 import io.gnosis.safe.utils.txActionInfoItems
 import pm.gnosis.model.Solidity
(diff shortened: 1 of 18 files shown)


app/src/main/java/io/gnosis/safe/ui/transactions/execution/TxReviewFragment.kt
```diff
@@ -203,34 +213,50 @@ class TxReviewFragment : BaseViewBindingFragment<FragmentTxReviewBinding>() {
 
                             if (decodedData.method.lowercase() == "multisend") {
```
DmitryBespalov, line 214: this whole method looks complicated and difficult to understand w/o any comments & many nested blocks. Maybe extracting each case into a separate method would make the complexity lower?

app/src/main/java/io/gnosis/safe/ui/transactions/execution/TxReviewViewModel.kt
```diff
@@ -162,14 +179,14 @@ class TxReviewViewModel
                     if (!userEditedFeeData) {
                         nonce = minNonce
                         gasLimit = estimationParams.estimate
-                        maxPriorityFeePerGas = Wei(BigInteger.valueOf(DEFAULT_MINER_TIP)).toGWei(activeSafe.chain.currency.decimals)
-                        maxFeePerGas = Wei(gasPrice).toGWei(activeSafe.chain.currency.decimals).plus(maxPriorityFeePerGas!!)
+                        maxPriorityFeePerGas =
+                            Wei(BigInteger.valueOf(DEFAULT_MINER_TIP)).toGWei(activeSafe.chain.currency.decimals)
+                        maxFeePerGas = Wei(gasPrice).toGWei(activeSafe.chain.currency.decimals)
```
DmitryBespalov, line 184: gasPrice is equivalent to the maxFeePerGas. I think this would charge more gas than needed by the estimation. Can you double-check please? https://ethereum.org/en/developers/docs/gas/#what-was-the-london-upgrade-eip-1559

app/src/main/java/io/gnosis/safe/ui/transactions/execution/TxReviewViewModel.kt
```diff
@@ -251,6 +311,14 @@ class TxReviewViewModel
         } ${activeSafe.chain.currency.symbol}"
     }
 
+    fun totalFee(): String? {
```
DmitryBespalov, line 314: is it possible to add unit tests for the important code areas?

elgatovital: of course, I will increase test coverage in a separate pr

app/src/main/java/io/gnosis/safe/ui/transactions/execution/TxReviewViewModel.kt
elgatovital: Here gasPrice is equivalent to base fee and maxFeePerGas is priority fee added to base fee as specified in eip-1559. The calculation should be correct. I will add some comments to make it more clear.

app/src/main/java/io/gnosis/safe/ui/transactions/execution/TxReviewFragment.kt
elgatovital: I have created a task for this [#1962](https://github.com/safe-global/safe-android/issues/1962)

Issue #1987, opened 2023-08-28
[Android] [TxExec] Monitor transactions submitted for execution


elgatovital:
- Check state of the submitted transaction
- Update state of the local transaction from `awaiting execution` to `pending` 
- Remove execute button from transaction details when coming back from execution flow
- Adjust state of of the submitted transaction in the queued tx list

PR #1990, opened 2023-08-28
Monitor transaction submitted for execution

elgatovital:
Handles #1987

- Persist tx data submitted to execution locally
- Get tx receipt and update tx status to pending if it was not yet indexed

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewModel.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewModel.kt
@@ -1,6 +1,7 @@
 package io.gnosis.safe.ui.transactions
 
 import androidx.annotation.StringRes
+import androidx.annotation.VisibleForTesting
 import androidx.lifecycle.viewModelScope
 import androidx.paging.PagingData
 import androidx.paging.cachedIn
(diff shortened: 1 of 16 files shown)


app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewModel.kt
```diff
@@ -66,37 +83,69 @@ class TransactionListViewModel
...
-                                        isConflict = isConflict,
-                                        localOwners = owners
-                                    )
-                                if (isConflict) {
-                                    TransactionView.Conflict(txView, txListEntry.conflictType)
-                                } else txView
+                                // conflict is resolved if there is local tx with same nonce
+                                // that was submitted for execution
+                                if (txLocal?.safeTxNonce == txListEntry.transaction.executionInfo?.nonce) {
+                                    if (txLocal?.safeTxHash == txListEntry.transaction.id.split("_").last() && txListEntry.transaction.txStatus == TransactionStatus.AWAITING_EXECUTION) {
```
DmitryBespalov, line 107: I noticed that the code here depends on the internal structure of the transaction id. This should not be the case, since the id might change in the future, thus breaking this logic.

In addition, this was one of the API agreements.

DmitryBespalov: > there is no way to get safeTxHash from the transaction list response
maybe locally stored data could include the transaction's safeTxHash, backend transaction id, and ethereum tx hash?

this way, you store the data at the point of transaction submission, and then you can lookup the safeTxHash by the backend transaction id

DmitryBespalov: > api agreements

this was not documented but implied; besides, it's a good design practice

elgatovital: > this way, you store the data at the point of transaction submission, and then you can lookup the safeTxHash by the backend transaction id 

having to request tx details just to get safeTxHash would be inefficient. Better to stick with timestamps for now.

DmitryBespalov: oh no, what I mean is you store the safeTxHash from the transaction details when you execute/submit the transaction - you have that data and the transaction id beforehand. So, if you store it in advance, then you can look it up later when searching by the transaction id. No need for additional network requests

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModel.kt
```diff
@@ -255,6 +277,9 @@ class TransactionDetailsViewModel
                 val canExecute = canBeExecutedFromDevice(newExecutionInfo, owners)
                 val safeOwner = isOwner(newExecutionInfo, owners)
 
+////                 reload details
```
rmeissner, line 280: Should this code be removed before this PR can be merged?

rmeissner (changes requested):
Please add tests for the new repository

Edit: Also a couple view models have new logic which should also be tested

elgatovital:
> Please add tests for the new repository
> 
> Edit: Also a couple view models have new logic which should also be tested

yes, already have task for it https://github.com/safe-global/safe-android/issues/1963

rmeissner:
We should not merge new logic without tests. It makes little sense to do the tests in a separate PR :thinking:

biafra23 (approved):
LGTM 🚀

DmitryBespalov (approved):

app/src/test/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModelTest.kt
```diff
@@ -108,6 +121,7 @@ class TransactionDetailsViewModelTest {
                 any()
             )
         } returns transactionDetails
+        coEvery { transactionLocalRepository.updateLocalTx(any(), any<String>()) } returns null
```
rmeissner, line 124: I would expect in the test that we check that this method was called. This is a crucial step and easy to verify via the  mocks

app/src/test/java/io/gnosis/safe/ui/transactions/TransactionListViewModelTest.kt
```diff
@@ -2,20 +2,42 @@ package io.gnosis.safe.ui.transactions
```
rmeissner, line 2: We made quite extensive changes to this view model, why is this not reflected in the tests? (i.e. we call updateLocalTxLatest and added new mappings)

app/src/test/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModelTest.kt
```diff
@@ -137,6 +151,82 @@ class TransactionDetailsViewModelTest {
...
+        coEvery { transactionLocalRepository.updateLocalTx(any(), any<String>()) } returns TransactionLocal(
+            CHAIN_ID,
+            Solidity.Address(BigInteger.ZERO),
+            BigInteger.ZERO,
+            (transactionDetails.detailedExecutionInfo as DetailedExecutionInfo.MultisigExecutionDetails).safeTxHash,
+            "",
+            TransactionStatus.PENDING,
+            0
+        )
+        coEvery { transactionLocalRepository.delete(any()) } just Runs
```
rmeissner, line 199: Same here, `delete` being called is critical we should verify this.

rmeissner (changes requested):

app/src/test/java/io/gnosis/safe/ui/transactions/TransactionListViewModelTest.kt
elgatovital: added tests

app/src/test/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModelTest.kt
```diff
@@ -157,6 +172,161 @@ class TransactionDetailsViewModelTest {
                 CHAIN_ID,
                 "tx_details_id"
             )
+            transactionLocalRepository.updateLocalTx(any(), any<String>())
```
rmeissner, line 175: To be honest this is not a good verification. We should know exactly what values it is being called with. This is not something blocking the PR anymore, but this is not the test quality I would expect for critical flows.

app/src/test/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsViewModelTest.kt
```diff
@@ -157,6 +172,161 @@ class TransactionDetailsViewModelTest {
...
+                UpdateDetails(txDetails = expectedTransactionInfoViewData),
+                this[0].viewAction
+            )
+        }
+        coVerify(exactly = 1) {
+            transactionRepository.getTransactionDetails(
+                CHAIN_ID,
+                "tx_details_id"
+            )
+            transactionLocalRepository.delete(any())
```
rmeissner, line 329: Same comment here as above. We should verify that exactly the delete is called that we expect. The most complete test would be to check that the method with any parameter was called exactly once and that the parameter was the expected transaction.

PR #1993, opened 2023-08-30
Release 3.9.0

biafra23:
Handles merging the changes made for the release  back to master

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/safe/send_funds/SendAssetReviewViewModel.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/safe/send_funds/SendAssetReviewViewModel.kt
@@ -251,10 +251,10 @@ class SendAssetReviewViewModel
                     SendAssetReviewState(
                         ViewAction.NavigateTo(
                             SendAssetReviewFragmentDirections.actionSendAssetReviewFragmentToSendAssetSuccessFragment(
-                                activeSafe.chain,
-                                safeTxHash,
-                                amountString,
-                                selectedAsset.symbol
+                                chain = activeSafe.chain,
+                                txId = safeTxHash, // to get non cached updates later we need the txId not the hash
+                                amount = amountString,
+                                token = selectedAsset.symbol
                             )
                         )
                     )
--- a/app/src/main/java/io/gnosis/safe/ui/safe/send_funds/SendAssetReviewViewModel.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/safe/send_funds/SendAssetReviewViewModel.kt
@@ -252,7 +252,7 @@ class SendAssetReviewViewModel
                         ViewAction.NavigateTo(
                             SendAssetReviewFragmentDirections.actionSendAssetReviewFragmentToSendAssetSuccessFragment(
                                 chain = activeSafe.chain,
-                                txId = safeTxHash, // to get non cached updates later we need the txId not the hash
+                                txId = safeTxHash, // TODO: to get non cached updates later we need the txId not the hash
                                 amount = amountString,
                                 token = selectedAsset.symbol
                             )
--- a/app/src/main/java/io/gnosis/safe/ui/settings/owner/list/OwnerListViewModel.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/settings/owner/list/OwnerListViewModel.kt
@@ -70,7 +70,7 @@ class OwnerListViewModel
                         .sortedBy { it.name }
                 val acceptedOwners = owners.filter { localOwner ->
                     safe.signingOwners.any {
-                        localOwner.address == it
+                        localOwner.address == it && localOwner.type != Owner.Type.LEDGER_NANO_X
                     }
                 }
                 val balances = rpcClient.getBalances(acceptedOwners.map { it.address })
--- a/app/src/main/java/io/gnosis/safe/ui/settings/owner/list/OwnerListViewModel.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/settings/owner/list/OwnerListViewModel.kt
@@ -70,6 +70,7 @@ class OwnerListViewModel
                         .sortedBy { it.name }
                 val acceptedOwners = owners.filter { localOwner ->
                     safe.signingOwners.any {
+                        //TODO: Modify this check when we have tx execution on Ledger Nano X
                         localOwner.address == it && localOwner.type != Owner.Type.LEDGER_NANO_X
                     }
                 }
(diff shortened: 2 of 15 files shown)


app/src/main/java/io/gnosis/safe/ui/safe/send_funds/SendAssetReviewViewModel.kt
```diff
@@ -251,10 +251,10 @@ class SendAssetReviewViewModel
                     SendAssetReviewState(
                         ViewAction.NavigateTo(
                             SendAssetReviewFragmentDirections.actionSendAssetReviewFragmentToSendAssetSuccessFragment(
-                                activeSafe.chain,
-                                safeTxHash,
-                                amountString,
-                                selectedAsset.symbol
+                                chain = activeSafe.chain,
+                                txId = safeTxHash, // to get non cached updates later we need the txId not the hash
```
elgatovital, line 255: this comment should be a TODO

app/src/main/java/io/gnosis/safe/ui/settings/owner/list/OwnerListViewModel.kt
```diff
@@ -70,7 +70,7 @@ class OwnerListViewModel
                         .sortedBy { it.name }
                 val acceptedOwners = owners.filter { localOwner ->
                     safe.signingOwners.any {
-                        localOwner.address == it
+                        localOwner.address == it && localOwner.type != Owner.Type.LEDGER_NANO_X
```
elgatovital, line 73: here adding FIXME or TODO comment would make sense. Something like: Allow ledger keys to be preselected after ledger execution flow is adjusted.

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewModel.kt
```diff
@@ -133,7 +132,14 @@ class TransactionListViewModel
                             localOwners = owners
                         )
                     } else {
-                        TransactionView.Unknown
+                        getTransactionView(
```
elgatovital, line 135: Under what condition can this branch happen? There is a local transaction with the same nonce, but a different timestamp, thus one of the conflicting transactions that were not selected for execution. Why is the view item created then?

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
```diff
@@ -70,7 +70,7 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
                 Navigation.findNavController(root).navigateUp()
             }
             refresh.setOnRefreshListener {
-                viewModel.loadDetails(txId)
+                viewModel.loadDetails(viewModel.txDetails?.txId ?: txId)
```
elgatovital, line 73: Please add comment explaining the reasoning behind loading details with id. Better would be to introduce additional nav arg called txHash and clearly distinguish btw 2 cases.

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
```diff
@@ -130,7 +130,7 @@ class TransactionDetailsFragment : BaseViewBindingFragment<FragmentTransactionDe
             viewModel.resumeFlow(ownerSelected()!!, ownerSigned())
             resetOwnerData()
         } else {
-            viewModel.loadDetails(txId)
+            viewModel.loadDetails(viewModel.txDetails?.txId ?: txId)
```
elgatovital, line 133: same as above

app/src/main/java/io/gnosis/safe/ui/transactions/execution/TxReviewViewModel.kt
```diff
@@ -123,6 +123,9 @@ class TxReviewViewModel
                     rpcClient.getBalances(acceptedOwners.map { it.address })
                 }.onSuccess {
                     executionKey = acceptedOwners
+                        .filter {
```
elgatovital, line 126: here TODO comment about when to  enable ledger keys would be helpful

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
biafra23: Aas far as I could tell, we do not have the txHash at that point. Because as I understand we only get it after the tx has been proposed. But maybe I am wrong.

elgatovital: you mean txId? It would be helpful to have this explanation in a comment.

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewModel.kt
```diff
@@ -121,9 +121,8 @@ class TransactionListViewModel
                 // that was submitted for execution
                 if (txLocal?.safeTxNonce == txListEntry.transaction.executionInfo?.nonce) {
                     // use submittedAt timestamp to distinguish between conflicting transactions
-                    if (txLocal?.submittedAt == txListEntry.transaction.timestamp.time && txListEntry.transaction.txStatus == TransactionStatus.AWAITING_EXECUTION) {
+                    if ((txLocal?.submittedAt == null || txLocal?.submittedAt == txListEntry.transaction.timestamp.time) && txListEntry.transaction.txStatus == TransactionStatus.AWAITING_EXECUTION) {
```
elgatovital, line 124: I would add a comment explaining that submittedAt is null for incoming transfers

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewModel.kt
biafra23: This branch is executed for any incoming  history transfer (and safe creation, and maybe even module txs). As they are not AWAITING_EXECUTION an do have `txLocal?.submittedAt == null`

app/src/main/java/io/gnosis/safe/ui/transactions/details/TransactionDetailsFragment.kt
biafra23: Yes, I meant to write tzxId

app/src/main/java/io/gnosis/safe/ui/transactions/TransactionListViewModel.kt
elgatovital: txLocal would be null then

biafra23: IIRC txLocal was not null

elgatovital: look at line 94:
```kotlin
 if (type == TransactionPagingSource.Type.QUEUE ) {
            txLocal = transactionLocalRepository.updateLocalTxLatest(safe)
        }
```
for history transactions txLocal should be null. Please add explicit else branch 
```kotlin
 if (type == TransactionPagingSource.Type.QUEUE ) {
            txLocal = transactionLocalRepository.updateLocalTxLatest(safe)
        } else {
txLocal = null
}
```
and then the getTransactionView can be replaced with TransactionView.Unknown. Because this branch should not be reachable for historic transactions.

Issue #1997, opened 2023-08-30
[Android] Ledger key should be hidden for the tx execution


liliya-soroka:
**Describe the bug**
Ledger key should be hidden for the tx execution

**To Reproduce**
1. Connect Ledger
2. try to execute tx using Ledger key
Current result: Nothing happens

**Expected behavior**
The Ledger key should not be available for execution

**Screenshots**

**Smartphone (please complete the following information):**
 - Device: Pixel 5
 - Android Version: Android 13
 - App Version 3.9.0-4039rc

**Environment**
prod
**Additional context**

liliya-soroka:
Verified
app version 3.9.0 -4056rc

PR #1998, opened 2023-08-31
Disable ledger keys for execution

biafra23:
Handles #1997
- Disable selction of ledger key as default
- remove ledger keys from eligible key list 
- disable execution button if no other owner keys are available

@gnosis/mobile-devs


biafra23:
Wrong base branch

Issue #1994, opened 2023-08-30
problem with statuses and stepper on the tx details page


liliya-soroka:
**Describe the bug**
Problem with statuses and stepper on 2nd signing and execution of the tx

Precondition:
safe 2 of N

**To Reproduce**
**1st case**
1. Create tx in the mobile app
2. Open tx details to sign it as a second owner
3. Pulldown to update the tx details
Current result: The second signature disappears from the stepper and "execution is required" is changed to the "confirmation is required" 
5. sign again and try to execute without pulling the updates
6. click "View tx details" after execution
7. "your confirmation is required is displayed as a state
8. pull down to update the tx details
Current result: Cancelled state is displayed and pulling down will not help in the status updating 
Expected result: The correct state should be displayed when the pull to refresh is used

2nd:
1. there is a created tx 
2. sign the tx as a second owner
Current result: the second signature doesn't appear on the tx details without pull down to refresh

3. Execute tx after pull to refresh for signature updates
Current result: Pending status has wrong view ( check the design for the ios), execution step disappears from the stepper
**Expected behavior**
Make sure that pending status and stepper looks simular to the ios ( Execute status should not disappear from the stepper) - ios app view 
<img width="367" alt="image" src="https://github.com/safe-global/safe-android/assets/338622/dbad3a37-7361-4410-9cea-b319d356057f">

**Screenshots**
https://drive.google.com/file/d/16W4kWllQIJCrfEuik76EO-X4OxFlruoR/view?usp=sharing
https://drive.google.com/file/d/16XCBJ6FTR74xhk5JXV0MeQy3EG9qG07t/view?usp=sharing

**Smartphone (please complete the following information):**
 - Device: pixel 5
 - Android Version: Android 13
 - App Version 3.9.0 -4039rc

**Environment**
prod
**Additional context**

biafra23:
```3. Pulldown to update the tx details
Current result: The second signature disappears from the stepper and "execution is required" is changed to the "confirmation is required"```
This might be a caching issue on the backend Wehn I look at response from the server it only contains one confirmation. Does iOS add the second confirmation locally?

liliya-soroka:
Verified
pull to refresh is required to see  confirmation and executed state, but all states are displayed correctly without the issues 

app version 3.9.0-4056rc

PR #2001, opened 2023-08-31
Gh 1994 - Show pending step, show up to date values

biafra23:
Handles #1994 

- Show pending step
- Use different id to retrieve uncached tx details 

@gnosis/mobile-devs

### DIFF
--- a/app/src/main/java/io/gnosis/safe/ui/safe/send_funds/SendAssetReviewViewModel.kt
+++ b/app/src/main/java/io/gnosis/safe/ui/safe/send_funds/SendAssetReviewViewModel.kt
@@ -251,10 +251,10 @@ class SendAssetReviewViewModel
                     SendAssetReviewState(
                         ViewAction.NavigateTo(
                             SendAssetReviewFragmentDirections.actionSendAssetReviewFragmentToSendAssetSuccessFragment(
-                                activeSafe.chain,
-                                safeTxHash,
-                                amountString,
-                                selectedAsset.symbol
+                                chain = activeSafe.chain,
+                                txId = safeTxHash, // to get non cached updates later we need the txId not the hash
+                                amount = amountString,
+                                token = selectedAsset.symbol
                             )
                         )
                     )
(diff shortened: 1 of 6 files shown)


app/src/main/java/io/gnosis/safe/ui/safe/send_funds/SendAssetReviewViewModel.kt
```diff
@@ -251,10 +251,10 @@ class SendAssetReviewViewModel
                     SendAssetReviewState(
                         ViewAction.NavigateTo(
                             SendAssetReviewFragmentDirections.actionSendAssetReviewFragmentToSendAssetSuccessFragment(
-                                activeSafe.chain,
-                                safeTxHash,
-                                amountString,
-                                selectedAsset.symbol
+                                chain = activeSafe.chain,
+                                txId = safeTxHash, // to get non cached updates later we need the txId not the hash
```
MouazAlzahabi, line 255: I got confused by the naming here, txId and safeTxHash

MouazAlzahabi: Are we sure that it's caching problem? why we don't have such on iOS ?

biafra23: because ios uses the txId: 

txIds look likes this: `multisig_0xfcFab1Eb61d3b1E8ff8F1eF1CD8e46F8be30b77f_0xc958935d43d24337ff7a425445ad05ca579231edf98eab9dd008db8447751f7d`

txHashes look like this: `0xc958935d43d24337ff7a425445ad05ca579231edf98eab9dd008db8447751f7d`

biafra23: But i didn't look into ios regarding the first time txId should be passed. In Android we do not know the txId at the time. It is returned in the txDetails.

biafra23: Here is an example request: https://safe-client.staging.5afe.dev/v1/chains/5/transactions/9f157207875105636c594bcce3ebea3eb5e0704f00cf64f59e3d286bc4258189

MouazAlzahabi: I'm actually not sure about this change if it's correct or not,

MouazAlzahabi (approved):

Issue #1995, opened 2023-08-30
[android]wrong time is displayed for the executed tx in the history list


liliya-soroka:
**Describe the bug**
[android]wrong time is displayed for the executed tx in the history list

**To Reproduce**
1. Execute tx and check the history
Current result: execute time+2 hours is displayed as execution time for the txs in the History list, but if you open tx details you will see the correct execution time
**Expected behavior**
The correct execution time according to the user time zone should be displayed in the history

**Smartphone (please complete the following information):**
 - Device: Pixel 5
 - Android Version:  Android 13
 - App Version 3.9.0-4039rc

**Environment**
prod
**Additional context**

liliya-soroka:
Verified
app version 3.9.0-4056rc

PR #2004, opened 2023-08-31
GH-1995 - Removed offset as we do the timezone handling in the client. 

biafra23:
Handles #1995

- The txdetail calls do not have a timezone offset

@gnosis/mobile-devs

### DIFF
--- a/data/src/main/java/io/gnosis/data/backend/GatewayApi.kt
+++ b/data/src/main/java/io/gnosis/data/backend/GatewayApi.kt
@@ -62,14 +62,14 @@ interface GatewayApi {
     suspend fun loadTransactionsHistory(
         @Path("chainId") chainId: BigInteger,
         @Path("address") address: String,
-        @Query("timezone_offset") timezoneOffset: Int = TimeZone.getDefault().getOffset(Date().time)
+        @Query("timezone_offset") timezoneOffset: Int = 0
     ): Page<TxListEntry>
 
     @GET("/v1/chains/{chainId}/safes/{address}/transactions/queued")
     suspend fun loadTransactionsQueue(
         @Path("chainId") chainId: BigInteger,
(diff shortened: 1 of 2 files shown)


data/src/main/java/io/gnosis/data/backend/GatewayApi.kt
```diff
@@ -62,14 +62,14 @@ interface GatewayApi {
...
-        @Query("timezone_offset") timezoneOffset: Int = TimeZone.getDefault().getOffset(Date().time)
+        @Query("timezone_offset") timezoneOffset: Int = 0
     ): Page<TxListEntry>
 
     @GET("/v1/chains/{chainId}/safes/{address}/transactions/queued")
     suspend fun loadTransactionsQueue(
         @Path("chainId") chainId: BigInteger,
         @Path("address") address: String,
-        @Query("timezone_offset") timezoneOffset: Int = TimeZone.getDefault().getOffset(Date().time)
+        @Query("timezone_offset") timezoneOffset: Int = 0
```
MouazAlzahabi, line 72: Do we need this parameter at all then ?

biafra23: I wanted to leave it because it is in the api. And maybe someone wants to use it in the future. I just set the default to 0.

the-real-chrizzo (approved):

data/src/main/java/io/gnosis/data/backend/GatewayApi.kt
DmitryBespalov: Got referenced here from #3327.  This change should bring inaccurate grouping of transactions by the date. 

Since the client doesn't report the offset to the server, the transaction list can have some transactions appearing in the wrong day, if those transactions made closer to midnight.

